PR #9, opened 2014-12-23
Microduino + MacOS additions

IanHarvey:
Hi Roger,
As discussed on the Arduino forum, I've made some mods for support of (a) Microduino Core STM32 and (b) using a Mac OS host to upload. It 'basically works' for me - digital I/O pins can be wiggled up and down - so you may want to take a first look at it at this point.

Thanks,
Ian

### DIFF
--- a/STM32F1XX/boards.txt
+++ b/STM32F1XX/boards.txt
@@ -324,7 +324,7 @@ microduino32_flash.upload.file_type=bin
 microduino32_flash.upload.ram.maximum_size=17000
 microduino32_flash.upload.flash.maximum_size=108000
 microduino32_flash.upload.usbID=1EAF:0003
-microduino32_flash.upload.altID=0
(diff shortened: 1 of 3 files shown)


rogerclarkmelbourne:
Ian,

Sorry to hijack this issue, but just to let you know SPI doesn't compile. I will raise this as an issue in my repo.

IanHarvey:
Roger,

Sorry about that, it looks like it’s not happy with the reduced number of SPI ports available on the Microduino. I’ll send a pull request when I’ve got a fix.

Thanks,
Ian

rogerclarkmelbourne:
Ian,

No worries.

One other thing that will effect you in the latest version, is that the path to DFU Utils on OSX has been changed as 2 other users found their DFU utils were in a different location.

I've updated the wiki to ask users to install via HomeBrew which then uses the path to DFU utils that is now in the Repo

PR #19, opened 2015-01-04
Unify libmaple source

bobc:
I believe I have now fixed the hangup problem we had before. 

We should definitely do plenty of testing before merging master!


rogerclarkmelbourne:
I've done a local pull and I will test to see if its got the same issue.

Thanks

Roger

rogerclarkmelbourne:
Bob

I've found a pre-existing issue where the RAM options don't work any more  :-(

section `.rodata' will not fit in region `ram'
region `ram' overflowed by 1272 bytes
collect2.exe: error: ld returned 1 exit status

This doesn't even work on the "bare minimum" sketch 

Basically, I did a local pull on your changes and found the issue, and then reverted all of your changes back out again and the bug still seems to be there.

I guess we're going to need to go back to a much older version and see when this issue occurred :-(

I presume it's either that the code is now much larger by default, or its a linker settings issue.

bobc:
Yes, the code size grew a bit when I added those linker options.

Personally, I would never use a RAM build, and I can't really think any reason why anyone would use them either. If it was me, I would drop support for RAM builds altogether!

Unfortunately I think it is inevitable that just fixing bugs will tend to increase the code size even if we weren't adding new features. I suppose we should aim to be no worse than the Due, as they are both Cortex M3 processors.

rogerclarkmelbourne:
Bob

No worries
I agree. I'm not sure who would ever use the RAM options.

I'll comment them out of boards.txt, so that if anyone really wants to mess around with them they can, but they won't generally be available

Thanks

Roger

PR #21, opened 2015-01-10
SPI build fixes

IanHarvey:
Hi Roger,

Here are a couple of patches - the first is a build fix for the SPI code on Microduino. I've checked it builds for Microduino and the Maple platforms, and that SPI.begin() runs without aborting, but I've not verified it with an actual SPI device.

The second fix is a tweak to the maple_upload script to allow dfu_util in either /opt/local/bin or /usr/local/bin.

Thanks,
Ian

### DIFF
--- a/STM32F1XX/libraries/SPI/src/SPI.cpp
+++ b/STM32F1XX/libraries/SPI/src/SPI.cpp
@@ -60,16 +60,24 @@ static void enable_device(spi_dev *dev,
                           spi_cfg_flag endianness,
                           spi_mode mode);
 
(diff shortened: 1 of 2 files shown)


rogerclarkmelbourne:
Ian,

Thanks

I will take a look at the changes and merge if there are no issues

Cheers

Roger

rogerclarkmelbourne:
Ian,

Thanks.

I've tested your changes to SPI and they seem OK, 

I've also looked at the change to the OSX upload file, and it seems reasonable to me...

So I've actioned the pull request.

PS. I'd obviously mistaken why it wasn't working on Microduino, the NSS thing must have been a red herring.

PR #33, opened 2015-02-27
OneWire examples does not compile

apla:


rogerclarkmelbourne:
I've done an alternative fix.

I'm not sure why the examples were not compiling for you.

Its a long story.

Basically, I did the port of the library to call the low level gpio_ calls, however after discussion with Bill Perry, it was decided to just use the high level API calls, e.g. digitalRead etc

As in that case I could just add an else clause as a generic fallback for all undefined architectures, and then submit the library back to Paul Stoffregen as a pull request from 

https://github.com/rogerclarkmelbourne/OneWire

I contacted Paul and he said in an email he would accept the update to the master copy of the library, but he wanted a warning to be put into the code.

I added the warning and then sent him a pull request to incorporate the changes into the master copy of this library (OneWire)

I'd hoped that Paul would be proactive and at least acknowledge my pull request, but as you can see here
https://github.com/PaulStoffregen/OneWire/pulls

Paul has not even taken the time to respond to say he will look at this, let alone action the pull request and then update the master download copy of the onewire lib, which he keeps on his own website . http://www.pjrc.com/teensy/td_libs_OneWire.html

If Paul ever updates the library, we will not need to have a copy in the libraries folder, and I was hoping that this would have happened by now

Unfortunately, I'm sensing a lot of resistance by Paul to do anything which benefits the STM32 world, as he makes his living from selling a competitor product, the Teensy, and this pull request would make OneWire work with lots of other competitor hardware as long as its fast enough e.g. TILaunchpad etc

apla:
Ok, as you wish. Did you make any performance tests, low level vs high level?

rogerclarkmelbourne:
I did tests using my oscilloscope and using the API calls is still well within spec. I also spoke to an authority on this (Bill Perry, I can send him your details if you are concerned about the timings etc)

However, the main reason that I had to abandon the use of libmaple specific calls, is that Paul Stoffregen would not accept changes to the master repo as they specific to hardware that he considers to be an unofficial / unsupported by a company.

Hence the only option to get these changes accepted into the master was to use the generic API calls, for political not technical reasons.


If you are interested, and have an AVR Arduino and a OneWire device, I think you will find that if you edit the AVR  ifdef to disable it, so that AVR uses the API calls, you'll find it actually works OK.

I have done those tests as well, and on AVR using API calls is actually slightly outside the OneWire spec for a binary 0 digit, as it holds the bus low for possibly 2uS too long.
However in practice this doesn't seem to make any different, and I suspect that the OneWire devices actually allow a bigger tolerance than is written into the spec, e.g. to allow for noise and long cable runs etc. 
But for most people the direct access stuff is probably unnecessary

And it wouldn't be required for either Teensy or PIC32 or Due as they are plenty fast enough to not need it. And as the OneWIre bus protocol is actually very slow, with a bit period of over 60uS from 
see
http://www.maximintegrated.com/en/products/1-wire/flash/overview/index.cfm

So the data rate is only around 2000 bytes per second (8 bit bytes), less overhead.

I don't think that using API calls will slow the throughput enough to be worth worrying about.

BTW. Onewire has an "Overdrive" mode which is 10 times as fast, but this is not implemented by the OneWire lib, and if it was, I think we would probably need the direct hardware access.

PR #34, opened 2015-03-03
RTClock compilation fixed

apla:
Fixed some paths in header files;
Added #ifdef to check for chip oscillator type

### DIFF
--- a/STM32F1/libraries/Untested/RTClock/utility/rtc_util.c
+++ b/STM32F1/libraries/Untested/RTClock/utility/rtc_util.c
@@ -60,17 +60,23 @@ void rtc_init(rtc_clk_src src) {
 			RCC_BASE->BDCR = RCC_BDCR_RTCSEL_NONE;
 			break;
 
-		case RTCSEL_LSE:
-			rcc_start_lse();
-			RCC_BASE->BDCR |= RCC_BDCR_RTCSEL_LSE;
-
-			break;
-
+		#ifdef RCC_BDCR_RTCSEL_LSI
 		case RTCSEL_LSI:
 		case RTCSEL_DEFAULT:
 			rcc_start_lsi();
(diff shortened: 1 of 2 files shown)


apla:
Hi!

Those patches have a story. 

I tried to make it works really hard on my mac. Got kernel panic after resetting chinese maple mini clone and running `dfu-util`. Created poor man ST-Link on breadboard. Found docs about maple mini external 32.768KHz oscillator pins (FYI: pins 12 and 13). Still cannot upload, but I will figure out why.

rogerclarkmelbourne:
Thanks

I know that "ahull" on the forum was also going to look at RTC, I took a quick look but the updated rcc files in the rtc library were totally different from the ones now in cores

Anyway, when I get chance I'll branch and pull and see if I can upload as I have a STlink (clone) adaptor as well as serial and a maple mini etc (I also have Jtag on order)

rogerclarkmelbourne:
Just curious, but are you saying this fix is unstable?

PS. I have just fixed the uart config stuff, so you may want to do a pull to get those changes (though most people won't have a need to change serial parity or stop bits etc

Also there is a fix for dtostrf not being included by default

I will look at the RTC stuff later, but I have to get on with the day job now ;-)

rogerclarkmelbourne:
Hi apla

Did you try to compile the example that comes with the rtc library, it seems to use functions that are not implemented by your change
e.g.
rcc_start_hse()

I'm not sure if your changes mean that these function calls in the utilities/rtc_util.c are not needed, or whether we still need to take these functions from the replacement rcc.c /h and put them into the core rcc.c/h ??

rogerclarkmelbourne:
I think we now have a separate fix for this

I will update the repo today or tomorrow so that RTC works

rogerclarkmelbourne:
RTC has now been fixed another way. So I'll close this request as its been superseded

PR #35, opened 2015-03-13
Master

victorpv:
These are the gfx_as and ili9342_as libraries that add several additional fonts. Adapted and tested on Maple Mini


rogerclarkmelbourne:
Thanks

If we are going for a specific  STM32 version of then Adafruit lib, there are possibly other optimizations that can be made to the LCD libraries (but I've not had time to look at your changes)

Most notably the calls to the line drawing routines are very inefficient in the version I did, because the way that the LCD libs use the graphics library base class, mean that there is a massive overhead of beginning and ending SPI transactions for every pixel in a line 

Looking at what Paul did for the Teensy, he moved a lot of the Graphics library code into the LCD library, so that he could control when the SPI begin and end were called

I had considered doing the same thing in the LCD lib that I ported, (based on Matthias's work)

But as the speed was usable in the existing code and because I wanted to get other boards STM32F104R V and Z series into the repo . and because I'm working on camera support (OV767) for STM32,  I've not had time to go back and look at integration of the graphics lib inside the LCD lib

Anyway, I'll take a look at this later, but I don't have a ILI9342   :-(  at least I don't think I have ;-)

I have a lot of displays, but I can't remember all their model numbers 

BTW. Was it you that posted a link to the Due code. It is very good and given time could be ported to STM32.

victorpv:
The ili9341 driver is exactly yours just with the name of the functions changed to adapt it to that gfx_as files.
In the gfx itself I didn't change much except for variable types, some char to int, and some int16 to int32 to take 32 bit addresses. Besides that I didn't look at any optimization on either file.
I was already happy enough to get it working, hehe. Seriously I have no idea of C++ and limited of C. I know about electronics, addresses, etc, so although I do not know the language well I could figure what may be going on when the fonts were showing just corrupt graphs.


From what I saw in the file, is not heavily modified from the Adafruit GFX original, but mostly a bunch of added functions at the end to support the extra fonts, and can write right justified in the screen, as well as centered, but those are extra funtions in the file.

The display was really an ILI9341, sorry for the typo, I hope you didn't scratch your head much on that ;)

victorpv:
Thinking on it, I believe the original author mentioned that changed the name in purpose so both the original adafruit libraries and his modified ones could coexist in the system. If we go optimizing, we should probably rename everything back to the standard names. I think that may more practical for everyone, so could use unmodified sketches with the modified libraries without an issue.
I can look at loops doing stupid stuff wasting time, but I don't know the stm32 registers and the SPI peripheral good enough to do any optimization on that.
I do remember seeing a page somewhere of someone that used the libmaple dma functions to speed up SPI. When I have time I will look on it and see I can understand enough of how DMA works in the stm and how paul did in the teensy and see if I can suggest something. It is much more fun learning C by trying to write or modify code and having to learn how it works than by reading book with no purpose in mind, so I may not be able to get it all running, but may be able to get it close enough for you or someone else to finish it without so much trouble.

victorpv:
I was having a look at their modified library. So they have a fifo that they fill with different commands and data, and they have different functions to just load one more byte in the fifo, and to load the last one and wait till the end of transmission.
I would think we should be able to do pretty much the same with buffer space and DMA, only instead of a hardware fifo we would use memory and DMA, but we can have one function loading data to the buffer, and one more that loads the last byte and trigger the end of transmission and wait to clear the dc and cs lines.
I will check the DMA libraries from libmaple see if we can reuse a good chunk of it. The current SPI implementation do not include any DMA capability right?

rogerclarkmelbourne:
Victor.

The current SPI doesn't use DMA, or have any functions for DMA.

There is an SPI DMA example here

http://polplaiconesa.com/tutorials/SPI+DMA.html

I'm pretty sure it compiles, but I've not had chance to see whether it actually works (I presume so)

Re: Differences in function names

I think its better if we use the normal function names for our version of these libraries, so that if people find code that uses these libraries, they don't need to modify any code for it to work.

Thanks

Roger

victorpv:
I will change the function names in the GFX library so it works with normal function names in the ILI9341 one. That way no matter what changes you make in that one, the GFX will still work with it.
That example of DMA SPI is the same one I had seen. But I understand it uses the functions from libmaple, we probably would need to just use those functions in the ILI library for transfers of blocks. In any case I think transferring 480 bytes at once should be a huge improvement over setting CS and DC lines, transferring one byte, clearing those lines, returning from the function... It seems so much overhead when most of the actions in the screen involve multiple pixels.

I will test the DMA example from that page just to confirm it still works fine, I still even have the display connected to the maple and all, so should take a minute to see if it works or not.

rogerclarkmelbourne:
Victor,

I've not had any luck with the DMA stuff

The SPI example seems to cause some sort of receiving SPI DMA to kick off, as I see lots of SPI clocks, but when I tried removing the RX DMA lines nothing happens :-(

I've notice in the github libmaple there is this example

https://github.com/leaflabs/libmaple/blob/master/examples/test-usart-dma.cpp

However to run it you have to remove the main() etc

And it still doesn't seem to work

It looks like leaflabs changed the way that the dma stuff is setup and use something called pipes in their newer code.
However the possibly old functions still exist.

So I'm not sure which one is supposed to work, possibly both.

Matri at leaflabs seems to sometimes respond to issues on libmaple, so you could try posting an issue there if none of the code seems to work, as I've not touched anything to do with DMA

victorpv:
Me neither, I also found this sdcard library with DMA from the same person:
https://github.com/polpla/Maple-SDFat-DMA/blob/master/Sd2Card.cpp
I was trying to DMA it to fill blocks in my display that's the only SPI device I have working around. I may try to wire an SD card reader and try my luck.
I have set a bunch of write serials in my code, and it never triggers an interrupt that is supposed to trigger at the end of the transfer, so I think the transfer is never going on.
From the sdcard library it seems like a TX without DMA RX should work, that library does it a few times, but in any case I believe the modifications they made to support those tubes broke the older functions. All around in the dma code they show those routines are deprecated.
I have the maple IDE installed and has the old libraries with no tubes, so I may try copying those over and removing the new ones, and see what happens.

rogerclarkmelbourne:
Victor

You may have seen this but there is a short thread on the maple forum about the DMA tubes 
http://forums.leaflabs.com/topic.php?id=1882

And as you say it looks like the other stuff is deprecated, though usually deprecated doesn't mean it no longer works. But in this case I suspect perhaps it does mean you can't use it any more.

Looking at the USART DMA demo, I think I misunderstood what it does
I think it waits until it has a buffer full of incoming data before triggering a transfer, but I'm not absolutely sure.

I will need to re run it and take a closer look

A better DMA example wold just be to use DMA to print some text via serial, I.e DMA a C string buffer to the serial port.

I also read the wiki on DMA on leaflabs site, and the DMA channel inside the DMA device needs to be chosen to match the peripheral, which is why the SPI stuff uses channel 2 and channel 3 for SPI and it makes a difference between send and receive which channel you can use

Anyway, I will take another look today, as I need to DMA image lines to my ILI9431 from my ov7670 camera

victorpv:
Saw it a bit earlier too, but haven't tried.
I am sure I was setting up the DMA with the old commands correctly. Reviewed the list from this page a few times:
http://wiki.leaflabs.com/index.php?title=DMA
plus double checked with sdcard dma library, which uses SPI1 too. 

I was planning to try the tubes later.
The transfer I was trying to do was just filling colors in the screen, and as long as something arrive I should see the screen lines changing, even if the DMA ends up reading past the buffer or anything, still the display would treat that just as color data and be ok. But I think the most clear indication of failure was that the DMA was set to trigger an interrupt at the end, and that was never happening. The sketch was not crashing, but it was just never initiating the DMA.
I'll give a try too to the tubes later, with a bit of luck that will work :)

rogerclarkmelbourne:
Thanks Victor

I looked at the Serial DMA example again, but it doesnt seem to work for me

This could be because it was designed to use SerialUSB to debug the data and Serial1 (hardware) to test, but I'm debugging on Serial1 (aka Serial serial for me, as I'm using a generic board and not using USB Serial)

I've posted a question on the old leaflabs maple forum  in case any of the few people who use that site can help e,g I think there is a still a moderator on that foruim even though it only gets about 1 post per week

Actually there seem to be new users of Maple mini who still don't know that we are all on the Arduino forum and that they can use the 1.6.1 IDE etc

victorpv:
I used the serial example as a base for an SPI transfer.
I am getting an error code -255 when setting up the tube.
I am still investigating it, but seems like this part of dma.c would return that error code:
103     if (!cfg_req_ok(channel, cfg)) { 
104         return -DMA_TUBE_CFG_EREQ; 

So I suspect I didn't set all the required parameters right. I will be checking that function to see what is it checking for.

victorpv:
Well I resolved that, one of the flags was not correct, now I get status 0 but still not working.
It seems like I get to the ISR routine, but the interrupt flags are all 0 and I should have least one set, to indicate either an error or success.

rogerclarkmelbourne:
Victor,

Thanks for keeping me updated
Unfortunately I've not had time to investigate the DMA stuff today.

Hopefully, I'll get time during one evening in the week

victorpv:
I was checking how the SPI to the ILI transfers occur in the normal way. Every time a byte is written one is read, even if discarded. I suspect the ILI is sending an acknowledge back. As I didn't setup DMA RX, that would fill the RX buffer and trip the interrupt as soon as the first byte goes out.
I am checking a couple of things more and will test setting up RX first so the RX buffer (1 byte) never fills up. If that doesn't work, I'll let it there until Marti or someone else answers in the forum.

victorpv:
Ok, so I set both TX and RX SPI DMA, and tried again. It is definitely sending something. The screen is filled black before the DMA transfer, and I see some white and colors dots in the corner once the DMA transfer starts.
I don't have a logic analyzer or an oscilloscope to see what exactly is wrong, but at least the fact that white dots appear in the line I am trying to write to, indicate that at least the DMA transfer started.
I also noticed that if I put my finger in the sclk signal, it paints more dots and different colors. 
I tried changing the SPI speed from DIV2 to DIV16, but the result is the same, a couple of white dots at the beginning of the line that should be filled, and if I touch sclk I get more random dots along that line.
Also after a transfer is complete, DMA should be disabled, then address the next line, then DMA enabled again, but does not seem to happen. It seems like the DMA transfer keeps going. Not sure if it is because some the registers do not get set correctly, or because the SPI peripheral is waiting for something else to happen.

At the moment I am leaving it there as I can't see the SPI signals to get any further. I have been considering getting a cheap analyzer off ebay, may be the right time...

Anyway my conclusions are:
-Seems to actually transmit something when I set both TX and RX DMA, I guess the ILI was indeed sending something back, and causing an Overrun condition. There may be other ways to address that.
-DMA does not stop after the number of bytes set to transfer like it should.

rogerclarkmelbourne:
Thanks Victor

Can you post your test code, I have a logic analyser, so could look to see if the data bytes are really being sent.

However I suspect that what us happening is that the SPI is doing multiple read requests, which the display is interpreting as writes

SPI doesn't have a direction control line, so of the display was expecting data, and the SPI clock runs, for a read, the display will read the data on the MOSI line, but the line will not be driven, hence will most likely be either noise, or all zeros or all ones .

Putting your finger on the board, is most likely putting electrical noise on MOSI which results in the speckled pixels you are seeng.


I need to take another look at DMA for data input from the OV7670 camera, so I will probably continue to investigate DMA when I get time.
Mainly because I'm finding it almost impossible to reliably read a 1mhz clocked 8 bit wide data stream from the camera.
But I think it is possible, somehow, to setup a DMA so that when I get a rising edge on the external clock, the DMA reads from the GPIO and puts it straight into ram.

I found this, about DMA ing a buffer to GPIO

https://vjordan.info/log/fpga/stm32-bare-metal-start-up-and-real-bit-banging-speed.html

And also this.  See code a long way down the page

http://forums.leaflabs.com/topic.php?id=774

rogerclarkmelbourne:
Hi Victor,

I've done a post about this on the forum, but I have managed to get DMA to Hardware Serial 1 working

See the code below
I don't have time to work out what to change to get it to work for SPI, but it should just be the device the registers and the DMA channel and Tube !

```
/*

// DMA setup
RCC->AHBENR |= RCC_AHBENR_DMA1EN;

// channel 1: mem:8bit -> peri:8bit 
DMA1_Channel1->CNDTR = 8;
DMA1_Channel1->CMAR = (uint32_t)buffer;
DMA1_Channel1->CPAR = (uint32_t)&(GPIOD->ODR);
DMA1_Channel1->CCR = 0;
DMA1_Channel1->CCR = DMA_CCR1_MEM2MEM | DMA_CCR1_PL | DMA_CCR1_MINC | DMA_CCR1_CIRC | DMA_CCR1_DIR | DMA_CCR1_EN;

while(DMA1_Channel1->CNDTR);
*/


#include <libmaple/dma.h>
#include <libmaple/usart.h>
#include <libmaple/gpio.h>

#define USART_DMA_DEV DMA1
#define USART_DMA_DST_SRC DMA_REQ_DST_USART1_TX

dma_tube_config tube_config;

char *buffer = "Hello World\n";


void setup_tube_config(void) 
{
    // We're sending to the USART data register. Serial.c_dev()
    // returns a pointer to the libmaple usart_dev for that serial
    // port, so this is a pointer to its data register.
    tube_config.tube_src = buffer;
    // We're only interested in the bottom 8 bits of that data register.
    tube_config.tube_src_size = DMA_SIZE_8BITS;
    // We're storing to rx_buf.
    tube_config.tube_dst = &Serial.c_dev()->regs->DR;
    // rx_buf is a char array, and a "char" takes up 8 bits on STM32.
    tube_config.tube_dst_size = DMA_SIZE_8BITS;
    // Only fill BUF_SIZE - 1 characters, to leave a null byte at the end.
    tube_config.tube_nr_xfers = strlen(buffer);
    // Flags:
    // - DMA_CFG_DST_INC so we start at the beginning of rx_buf and
    //   fill towards the end.
    // - DMA_CFG_CIRC so we go back to the beginning and start over when
    //   rx_buf fills up.
    // - DMA_CFG_CMPLT_IE to turn on interrupts on transfer completion.
    tube_config.tube_flags = DMA_CFG_SRC_INC | DMA_CFG_CIRC;// | DMA_CFG_CMPLT_IE;
    // Target data: none. It's important to set this to NULL if you
    // don't have any special (microcontroller-specific) configuration
    // in mind, which we don't.
    tube_config.target_data = NULL;
    // DMA request source.

    tube_config.tube_req_src = DMA_REQ_SRC_USART1_TX;
   // Serial.println(tube_config.tube_req_src  & 0x7);
}

#define USART_TX_DMA_TUBE DMA_CH4

void setup_dma_xfer(void)

{
    // First, turn it on.
    dma_init(USART_DMA_DEV);
    // Next, configure it by calling dma_tube_cfg(), and check to make
    // sure it succeeded. DMA tubes have many restrictions on their
    // configuration, and there are configurations which work on some
    // types of STM32 but not others. libmaple tries hard to make
    // things just work, but checking the return status is important!
    int status = dma_tube_cfg(USART_DMA_DEV, USART_TX_DMA_TUBE, &tube_config);
    
   // Serial.println(status);
    ASSERT(status == DMA_TUBE_CFG_SUCCESS);
    // Now we'll perform any other configuration we want. For this
    // example, we attach an interrupt handler.
    
//    dma_attach_interrupt(USART_DMA_DEV, USART_TX_DMA_TUBE, tx_dma_irq);
    
    // Turn on the DMA tube. It will now begin serving requests.
    dma_enable(USART_DMA_DEV, USART_TX_DMA_TUBE);
}

void setup() 
{


  Serial.begin(115200);
  usart_dev *serial_dev = Serial.c_dev();
  serial_dev->regs->CR3 = USART_CR3_DMAT;
  while (!Serial.available());

setup_tube_config();
setup_dma_xfer();

}

void loop() {
Serial.print(".");
delay(1000);
}

```

victorpv:
My code is a bit of a mess as I was testing the tubes and the old way, so there are big parts commented out. What I did was taken one of the GFX_AS library examples, that fill up lines in different colors, and copied the fastVline from the library to my sketch, to call that one instead of the original, but let the original calls to setup the display and to set the x, y parameters of the fill area . Then tried to do the actual fill up with the DMA call rather than SPI.write.
About touching with my fingers, I only needed to touch between the sclk and the NSS pin, BUT, the first couple of dots in the line get painted before I even touch anything, but consistenly only a couple of dots. If every read clock period caused the display to read random data, it should fill the whole line with random stuff.
Instead it paints a dot or two, and stays there. After that when I touch I get more random which fit in your theory that MOSI. I think the TX is not going on as it should, but I can't explain why RX would work and TX not, and it seems that those random dots are coming from the RX clocking, which is set with DMA too.
Have a look at my code, you can remove the commented lines out to clean the mess a bit:
#ifdef SPI_HAS_TRANSACTION
static inline void spi_begin(void) __attribute__((always_inline));
static inline void spi_begin(void) {
  SPI.beginTransaction(SPISettings(8000000, MSBFIRST, SPI_MODE0));
}
static inline void spi_end(void) __attribute__((always_inline));
static inline void spi_end(void) {
  SPI.endTransaction();
}
#else
#define spi_begin()
#define spi_end()
#endif

/*
 An example showing rainbow colours on a 2.2" TFT LCD screen
 and to show a basic example of font use.

 The existing Adafruit font is still in the library
 Only new font sizes 2,4,6 and 7 are implemented in the Adafruit_GFX_AS library.

 This examples uses the hardware SPI only. Non-hardware SPI
 is just too slow (~8 times slower!)

 Alan Senior 18/1/2015

 */

// 1. Import "dma.h"
//#include <../libmaple/include/libmaple/dma.h>
#include <libmaple/dma.h> //"./dma.h"

#define SPI_DMA_DEV DMA1
#define SPI_TX_DMA_TUBE DMA_CH3
#define SPI_DMA_REQ_SRC DMA_REQ_SRC_SPI1_TX
volatile uint32 irq_fired = 0;
volatile uint32 isr = 0;
dma_tube_config tube_config;

// DMA stuff ************************************


// These are the connections for the UNO
#define sclk 13  // Don't change
#define mosi 11  // Don't change
#define cs   8
#define dc   10
#define rst  9// you can also connect this to the Arduino reset
//#include "../../libraries/Adafruit_GFX_AS/Adafruit_GFX_AS.h"    // Core graphics library
//#include "../../libraries/Adafruit_ILI9341_AS/Adafruit_ILI9341_AS.h" // Hardware-specific library
#include ".\Adafruit_GFX_AS.h"    // Core graphics library
#include ".\Adafruit_ILI9341_AS.h" // Hardware-specific library
#include <SPI.h>

Adafruit_ILI9341_AS tft = Adafruit_ILI9341_AS(cs, dc, rst);       // Invoke custom library

unsigned long targetTime = 0;
byte red = 31;
byte green = 0;
byte blue = 0;
byte state = 0;
unsigned int colour = red << 11;

//DMA buffer
uint8_t bytesToSend[512];

volatile uint32 *mosiport, *clkport, *dcport, *rsport, *csport;
uint32_t  _cs, _dc, _rst, _mosi, _miso, _sclk;
uint32_t  mosipinmask, clkpinmask, cspinmask, dcpinmask;
bool dmaActive;

void DMAEvent() {

  //we get the DMA event
//  dma_irq_cause event = dma_get_irq_cause(DMA1, DMA_CH3);
//  Serial.println (event, BIN);
  delay (250);
/*  switch (event) {
    //the event indicates that the transfer was successfully completed
    case DMA_TRANSFER_COMPLETE:
      //11. Disable DMA when we are done
      dma_detach_interrupt(DMA1, DMA_CH3);
      dmaActive = false;
      dma_disable(DMA1, DMA_CH3);
      Serial.println("Done transfering");
      break;
    //the event indicates that there was an error transmitting
    case DMA_TRANSFER_ERROR:
  */    //11. Disable DMA when we are done
      dmaActive = false;
      dma_disable(DMA1, DMA_CH3);
            dma_disable(DMA1, DMA_CH2);
      Serial.println("ISR");
//      break;
//  }
}
void setup_tube_config(int16_t _size) {
  tube_config.tube_dst = &SPI1->regs->DR;
  tube_config.tube_src_size = DMA_SIZE_8BITS;
  tube_config.tube_src = bytesToSend;
  tube_config.tube_dst_size = DMA_SIZE_8BITS;
  tube_config.tube_nr_xfers = _size;
  tube_config.tube_flags = DMA_MINC_MODE |  DMA_FROM_MEM | DMA_TRNS_CMPLT | DMA_TRNS_ERR;
  tube_config.target_data = NULL;
  tube_config.tube_req_src = DMA_REQ_SRC_SPI1_TX;
//  Serial.print ("DMA_REQ_SRC_SPI1_TX: ");
//  Serial.println (DMA_REQ_SRC_SPI1_TX, BIN);
}

void tx_dma_irq(void) {
  irq_fired = 1;
  isr = dma_get_isr_bits(DMA1, DMA_CH3);
}

void setup_dma_xfer(void) {
  dma_init(DMA1);
//  dma_init(SPI_DMA_DEV);
  int status = dma_tube_cfg(DMA1, DMA_CH3, &tube_config);
//  Serial.print ("Tube status: ");
//  Serial.println (status, DEC);
  //delay (5000);
  ASSERT(status == DMA_TUBE_CFG_SUCCESS);
  dma_attach_interrupt(DMA1, DMA_CH3, tx_dma_irq);
  //  dma_attach_interrupt(SPI_DMA_DEV, SPI_TX_DMA_TUBE, DMAEvent);
  dma_enable(DMA1, DMA_CH3);
}




void drawFastVLine3(int16_t x, int16_t y, int16_t h, uint16_t color) {
  int _width = tft.width();
  int _height = tft.height();
  Serial.println ("entered the new routine");
  //	 delay (5000);
  // Rudimentary clipping
  if ((x >= _width) || (y >= _height)) return;

  if ((y + h - 1) >= _height)
    h = _height - y;

  spi_begin();
  tft.setAddrWindow(x, y, x, y + h - 1);
  spi_end();

//  Serial.println ("Sent the address Window");
  uint8_t hi = color >> 8, lo = color;
  digitalWrite(dc, HIGH);
  digitalWrite(cs, LOW);
//    while (h--) {
//      SPI.transfer(hi);
//      SPI.transfer(lo);
//    }
//    while (h--) {
//      SPI.write(hi);
//      SPI.write(lo);
//    }
  
  SPI.setClockDivider(SPI_CLOCK_DIV2);
  SPI.setBitOrder(MSBFIRST);
  SPI.setDataMode(SPI_MODE0);
  spi_init (SPI1);

  dma_init(DMA1);
  spi_tx_dma_enable(SPI1);
    spi_rx_dma_enable(SPI1);

  uint8_t bytesToSend[h];
  byte bytesReceived[512];
 for (int i = 0; i < h - 1; i = i + 2)
  {
    bytesToSend[i] = hi;
    bytesToSend[i + 1] = lo;
  }

//  Serial.print("Prepared to send: ");
//  Serial.println(h, DEC);
//  Serial.println (bytesToSend[h-1],HEX);
/*
  setup_tube_config(h);
//  Serial.print("Setup tube config ");
//  delay (2000);
  setup_dma_xfer();
//  Serial.print("Setup tube transfer ");
//  delay (2000);

  while (irq_fired = 0) delayMicroseconds(1);
  dma_disable(DMA1, DMA_CH3);
  irq_fired = 0;
//  Serial.print("IRQ Fired. ISR: ");
  Serial.println(isr, HEX);

 */
    dma_setup_transfer(DMA1, DMA_CH3, &SPI1->regs->DR, DMA_SIZE_8BITS,
                       bytesToSend, DMA_SIZE_8BITS, (DMA_MINC_MODE |  DMA_FROM_MEM | DMA_TRNS_CMPLT /*| DMA_TRNS_ERR*/));
    dma_setup_transfer(DMA1, DMA_CH2, &SPI1->regs->DR, DMA_SIZE_8BITS,
                       bytesReceived, DMA_SIZE_8BITS, (DMA_MINC_MODE | DMA_CIRC_MODE /*| DMA_FROM_MEM*/));
            Serial.println ("passed dma setup");
     	 delay (100);
//    dma_attach_interrupt(DMA1, DMA_CH3, DMAEvent);
//          Serial.println ("passed attach");
//     	 delay (2000);
    dma_set_priority(DMA1, DMA_CH2, DMA_PRIORITY_VERY_HIGH);
    dma_set_priority(DMA1, DMA_CH3, DMA_PRIORITY_VERY_HIGH);
           Serial.println ("passed priority");
      	 delay (100);
    dma_set_num_transfers(DMA1, DMA_CH3, h);
    dma_set_num_transfers(DMA1, DMA_CH2, h);
               Serial.println ("passed num transfer");
      	 delay (100);

                   Serial.println ("passed dma active");
      	 delay (100);
//    SPI.read (bytesReceived, 1);
//    Serial.print ("This was waiting in the TX buffer");
//    Serial.println (bytesReceived[0], HEX);
// for (int i=0; i<h; i++) {
    dmaActive = true;
   dma_enable(DMA1, DMA_CH2);
    dma_enable(DMA1, DMA_CH3);
    Serial.println ("passed dma enable");
    delay (100);
//    while (dmaActive) delayMicroseconds(1);
//    Serial.println ("passed dma active");
     	 delay (200);
 //   dma_disable(DMA1, DMA_CH3);
    Serial.println ("dma disabled");
    delay (100);
// }
  digitalWrite(cs, HIGH);
}


void setup(void) {
  Serial.begin (115200);
  delay (5000);
  tft.begin();
  tft.setRotation(2);
  tft.fillScreen(ILI9341_BLACK);

  targetTime = millis() + 1000;
  //

  pinMode(dc, OUTPUT);
  pinMode(cs, OUTPUT);
  csport    = portOutputRegister(digitalPinToPort(cs));
  cspinmask = digitalPinToBitMask(cs);
  dcport    = portOutputRegister(digitalPinToPort(dc));
  dcpinmask = digitalPinToBitMask(dc);

  // 4. Initialize DMA.


  //5. Enable DMA to use SPI communication; both TX (output) and RX (input).
  //   spi_tx_dma_enable(SPI1);
  //   spi_rx_dma_enable(SPI1);
  Serial.println ("passed dma_init");
  /*
    byte bytesToSend[512];
    //fill the array with data (put your own)
    for (int i = 0; i < 512; i++) {
      bytesToSend[i] = 0x0F;
    }



       // 6. Setup a DMA transfer (for both TX and RX). If we only want
       // to read (RX) or write (TX) it is fine to just setup one. In this
       // case we want to transfer (write and get a response) so we do both.
       // Parameters:
       //  - DMA
       //  - DMA channel
       //  - Memory register for SPI
       //  - The size of the SPI memory register
       //  - The buffer we want to copy things to or transmit things from
       //  - The unit size of that buffer
       //  - Flags (see the Maple DMA Wiki page for more info in flags)
       dma_setup_transfer(DMA1, DMA_CH2, &SPI1->regs->DR, DMA_SIZE_8BITS,
                          bytesToSend, DMA_SIZE_8BITS, (DMA_MINC_MODE | DMA_TRNS_CMPLT | DMA_TRNS_ERR));

       // 7. Attach an interrupt to the transfer. Note that we need to add
       // the interrupt flag in step 6 (DMA_TRNS_CMPLT and DMA_TRNS_ERR).
       // Also, we only attach it for one of the transfers since they are
       // going to finish at the same time because they are in sync.
       dma_attach_interrupt(DMA1, DMA_CH2, DMAEvent);

       //8. Setup the priority for the DMA transfer.
       dma_set_priority(DMA1, DMA_CH2, DMA_PRIORITY_VERY_HIGH);


       // 9. Setup the number of bytes that we are going to transfer.
       dma_set_num_transfers(DMA1, DMA_CH2, 512);

       // 10. Enable DMA to start transmitting. When the transmission
       // finishes the event will be triggered and we will jump to
       // function DMAEvent.
       dma_enable(DMA1, DMA_CH2);
    */
}

void loop() {

  if (targetTime < millis()) {
    targetTime = millis() + 10000;
    for (int i = 0; i < 240; i++) {
 //     Serial.println ("About to start");
      drawFastVLine3(i, 0, tft.height(), colour);
      switch (state) {
        case 0:
          green += 2;
          if (green == 64) {
            green = 63;
            state = 1;
          }
          break;
        case 1:
          red--;
          if (red == 255) {
            red = 0;
            state = 2;
          }
          break;
        case 2:
          blue ++;
          if (blue == 32) {
            blue = 31;
            state = 3;
          }
          break;
        case 3:
          green -= 2;
          if (green == 255) {
            green = 0;
            state = 4;
          }
          break;
        case 4:
          red ++;
          if (red == 32) {
            red = 31;
            state = 5;
          }
          break;
        case 5:
          blue --;
          if (blue == 255) {
            blue = 0;
            state = 0;
          }
          break;
      }
      colour = red << 11 | green << 5 | blue;
    }

    /*   // The standard ADAFruit font still works as berfore
       tft.setTextColor(ILI9341_BLACK, ILI9341_BLACK); // Note these fonts do not plot the background colour
       tft.setCursor (68, 5);
       tft.print("Original ADAfruit font!");

       // The new larger fonts do not use the .setCursor call, coords are embedded
       tft.setTextColor(ILI9341_BLACK); // Do not plot the background colour
       // Overlay the black text on top of the rainbow plot (the advantage of not drawing the backgorund colour!)
       tft.drawCentreString("Font size 2",120,14,2); // Draw text centre at position 120, 14 using font 2
       tft.drawCentreString("Font size 4",120,30,4); // Draw text centre at position 120, 30 using font 4
       tft.drawCentreString("12.34",120,54,6); // Draw text centre at position 120, 54 using font 6
       tft.drawCentreString("12.34 is in font size 6",120,92,2); // Draw text centre at position 120, 92 using font 2
       // Note the x position is the top of the font!

       // draw a floating point number
       float pi = 3.14159; // Value to print
       int precision = 3;  // Number of digits after decimal point
       int xpos = 90;      // x position
       int ypos = 110;     // y position
       int font = 2;       // font number only 2,4,6,7 valid. Font 6 only contains characters [space] 0 1 2 3 4 5 6 7 8 9 0 : . a p m
       xpos+=tft.drawFloat(pi,precision,xpos,ypos,font); // Draw rounded number and return new xpos delta for next print position
       tft.drawString(" is pi",xpos,ypos,font); // Continue printing from new x position
    */
  }
}
//experimetal dma

victorpv:
I see you commented out the interrupt part, so there isn't any ISR, and the tube flags also comment out to generate an interrupt on complete.
I never tried without any flash, could give it a try. Besides that, the part where I used the tubes is pretty similar to yours, just with SPI registers instead. I wonder is something else needs to be configured in the SPI peripheral, but from the examples I read only, I could not find any extra step.

Perhaps I am not setting the flash right? or we should avoid interrupts?
My code gets to generate an interrupt and I get to the ISR routine, but that should print the content of the isr bit for DMA1 CH3, that should indicate what triggered the interrupt, and it was always 0. I wonder if my nvic table gets messed up something way and that ISR routine gets called on something different than DMA finishing the transfer.

victorpv:
About DMA from gpio, have a look at stm32 AN2548 GPIO_SRAM, I am reading some of stm examples on dma, and saw that one about reading GPIO.

victorpv:
Hey Roger,

Made a little bit of progress. Now I got half a line filled with the same color!! I think that is a big victory, DMA transfer is happening but I have to check if I set the number of bytes correctly etc.
I found another USART TX example here http://forums.leaflabs.com/topic.php?id=1719, and this one at least posted the final code that worked for him. I think it is pretty much the same you were trying. Now my SPI transfer is ending halfway down the line, rather than completing the line and going to the next one, so I'll recheck everything and let you know.
When I get it working I'll try to clean up the code and paste it here.

rogerclarkmelbourne:
Thanks victor

I did look at SPI dam again yesterday, but could not transfer at all using a modified version of my USART DMA

I think perhaps I need to read data via dma as well as write 

Re: interrupt not being called

What buffer mode do you use? Perhaps circular is no good?

I read the manual about SPI dma but it was not very helpful :-(

thanks for the information on GPIO DMA I will need to find the source code for that Application note

victorpv:
Ok, I am giving up at the moment on detecting when the transfer is finished.
The minimum amount of time I can wait between activating the DMA transfer and deactivating it, is 150uS. Currently that is wasted time, we could do something better with it if we could detect the end of the transfer properly.
Still it goes down from 212mS filling a screen in the old way, to 60mS filling it with DMA, and if those 150uS were not wasted, but rather used in filling up a second buffer for the next line, I'm sure the speed up would be huge.
Have a look at the code, perhaps you can figure out why the flags are not working for me. I hope it also help you with your camera project, you should be able to setup multiple buffers and just transfer one after another while doing other stuff.

This is the code in it's current working incarnation, using the _AS libraries for the normal part, and a function in the sketch for the DMA. If you have an ILI9341 it should compile and work fine, and shows in the screen alternatively how long it takes to complete a pass with DMA, and a pass with SPI.transfer.

#ifdef SPI_HAS_TRANSACTION
static inline void spi_begin(void) __attribute__((always_inline));
static inline void spi_begin(void) {
  SPI.beginTransaction(SPISettings(8000000, MSBFIRST, SPI_MODE0));
}
static inline void spi_end(void) __attribute__((always_inline));
static inline void spi_end(void) {
  SPI.endTransaction();
}
#else
#define spi_begin()
#define spi_end()
#endif

/*
 An example showing rainbow colours on a 2.2" TFT LCD screen
 and to show a basic example of font use.

 The existing Adafruit font is still in the library
 Only new font sizes 2,4,6 and 7 are implemented in the Adafruit_GFX_AS library.

 This examples uses the hardware SPI only. Non-hardware SPI
 is just too slow (~8 times slower!)

 Alan Senior 18/1/2015

 */

// 1. Import "dma.h"
//#include <../libmaple/include/libmaple/dma.h>
#include <libmaple/dma.h> //"./dma.h"

#define SPI_DMA_DEV DMA1
#define SPI_TX_DMA_TUBE DMA_CH3
#define SPI_DMA_REQ_SRC DMA_REQ_SRC_SPI1_TX
volatile uint32 irq_fired = 0;
volatile uint32 isr = 0;
dma_tube_config tube_config;

// DMA stuff ************************************


// These are the connections for the UNO
#define sclk 13  // Don't change
#define mosi 11  // Don't change
#define cs   8
#define dc   10
#define rst  9// you can also connect this to the Arduino reset

#include ".\Adafruit_GFX_AS.h"    // Core graphics library
#include ".\Adafruit_ILI9341_AS.h" // Hardware-specific library

#include <SPI.h>

Adafruit_ILI9341_AS tft = Adafruit_ILI9341_AS(cs, dc, rst);       // Invoke custom library

unsigned long targetTime = 0;
byte red = 31;
byte green = 0;
byte blue = 0;
byte state = 0;
unsigned int colour = red << 11;

bool dmaActive;

// ISR routine not used at all, doesn't work for some reason.
void DMAEvent() {

  dmaActive = 0;
  dma_disable(DMA1, DMA_CH3);
}

void drawFastVLine3(int16_t x, int16_t y, int16_t h, uint16_t color) {
  int _width = tft.width();
  int _height = tft.height();

  // Rudimentary clipping
  if ((x >= _width) || (y >= _height)) return;

  if ((y + h - 1) >= _height)
    h = _height - y;

  spi_begin();
  tft.setAddrWindow(x, y, x, y + h - 1);
  spi_end();
  uint8_t hi = color >> 8, lo = color;
  digitalWrite(dc, HIGH);
  digitalWrite(cs, LOW);
  byte bytesToSend[h * 2];
  for (int i = 0; i < (h * 2) - 1 ; i = i + 2)
  {
    bytesToSend[i] = hi;
    bytesToSend[i + 1] = lo;
  }
  spi_tx_dma_enable(SPI1);
  dma_init(DMA1);
  dma_setup_transfer(DMA1, DMA_CH3, &SPI1->regs->DR, DMA_SIZE_8BITS,
                     bytesToSend, DMA_SIZE_8BITS, (DMA_MINC_MODE |  DMA_FROM_MEM | DMA_TRNS_CMPLT));
  dma_set_num_transfers(DMA1, DMA_CH3, h * 2); // 2 bytes per pixel
  dma_attach_interrupt(DMA1, DMA_CH3, DMAEvent);
  dmaActive = true;
  dma_enable(DMA1, DMA_CH3);
  /***** this section should check that the transfer is complete and continue ********/
  //    while (dmaActive) {
  //      uint8 status_bits = dma_get_isr_bits(DMA1, DMA_CH3);
  //      if (status_bits & 0x2) {
  //        dmaActive = false;
  //      }
  //    }
  //    dma_disable(DMA1, DMA_CH3);
  //    dma_clear_isr_bits(DMA1, DMA_CH3);


  delayMicroseconds(150);
  //  delay (1);
  dma_disable(DMA1, DMA_CH3);

  digitalWrite(cs, HIGH);
  spi_tx_dma_disable(SPI1);
}


void setup(void) {
  Serial.begin (115200);
  delay (5000);
  tft.begin();
  tft.setRotation(2);
  tft.fillScreen(ILI9341_BLACK);

  targetTime = millis() + 1000;
  //

  pinMode(dc, OUTPUT);
  pinMode(cs, OUTPUT);
}

void loop() {

  if (targetTime < millis()) {
    //   targetTime = millis() + 10000;
    targetTime = millis();
    for (int i = 0; i < 240; i++) {
      //     Serial.println ("About to start");
      drawFastVLine3(i, 0, tft.height(), colour);
      switch (state) {
        case 0:
          green += 2;
          if (green == 64) {
            green = 63;
            state = 1;
          }
          break;
        case 1:
          red--;
          if (red == 255) {
            red = 0;
            state = 2;
          }
          break;
        case 2:
          blue ++;
          if (blue == 32) {
            blue = 31;
            state = 3;
          }
          break;
        case 3:
          green -= 2;
          if (green == 255) {
            green = 0;
            state = 4;
          }
          break;
        case 4:
          red ++;
          if (red == 32) {
            red = 31;
            state = 5;
          }
          break;
        case 5:
          blue --;
          if (blue == 255) {
            blue = 0;
            state = 0;
          }
          break;
      }
      colour = red << 11 | green << 5 | blue;
    }
    tft.setTextColor(ILI9341_BLACK, ILI9341_BLACK); // Note these fonts do not plot the background colour
    tft.setTextSize(2);
    tft.setCursor (24, 10);
    tft.print("DMA time: ");
    tft.print(String(millis() - targetTime));
    tft.print("ms");
    targetTime = millis() + 10000;

    while (millis() < targetTime);

    targetTime = millis();
    for (int i = 0; i < 240; i++) {
      //     Serial.println ("About to start");
      tft.drawFastVLine(i, 0, tft.height(), colour);
      switch (state) {
        case 0:
          green += 2;
          if (green == 64) {
            green = 63;
            state = 1;
          }
          break;
        case 1:
          red--;
          if (red == 255) {
            red = 0;
            state = 2;
          }
          break;
        case 2:
          blue ++;
          if (blue == 32) {
            blue = 31;
            state = 3;
          }
          break;
        case 3:
          green -= 2;
          if (green == 255) {
            green = 0;
            state = 4;
          }
          break;
        case 4:
          red ++;
          if (red == 32) {
            red = 31;
            state = 5;
          }
          break;
        case 5:
          blue --;
          if (blue == 255) {
            blue = 0;
            state = 0;
          }
          break;
      }
      colour = red << 11 | green << 5 | blue;
    }
    tft.setTextColor(ILI9341_BLACK, ILI9341_BLACK); // Note these fonts do not plot the background colour
    tft.setTextSize(2);
    tft.setCursor (24, 10);
    tft.print("Old style: ");
    tft.print(String(millis() - targetTime));
    tft.print("ms");
    targetTime = millis() + 10000;
  }
}

rogerclarkmelbourne:
Thanks Victor

I will take a look.

victorpv:
About the buffer mode, I was not setting it to circular. It sends correctly the 640 bytes that make up a line and stops sending there, but I can not detect when it has finished with either an interrupt or looking at the ISR flags, that according to Marti USAR example, should get set even if the interrupt is not enabled. 
So at this point I don't think the problem was the interrurpt not getting enabled, but rather the flag not setting, or getting cleared, for some reason.
It still works pretty good like that.
If I have time I may try to modify the _AS library to take advantage of DMA when filling lines and blocks, and just make a delay such as delayMicrosecond(_numBytes/4);
In my current code, it takes 150uS for 640bytes, so without calculating other overheads, I think /4 would be a safe bet (160uS for 640). Anyway detecting the flags correctly would be the ideal solution, but I have read all I could find, and still didn't figure out what is wrong with that.

victorpv:
Hey Roger, I got the ISR working. 

I declared dmaActive as volatile, and changed the ISR routine to set dmaActive = false, rather than 0. 
I did a test and dmaActive = 0 works fine too, but false I guess looks more correct.
Declaring that value volatile was what made the trick.

I am editing the code in the post above with the corrections. Now we can properly do other stuff while the DMA is going on if we wanted, and just wait for the previous one to finish.
I will add basic DMA support to the ILI9341 library. I am by no means a programmer, so when I have it modified I will send you a pull request to see if you can polish it.

I am not sure why that value required to be declared volatile, none of the examples I saw declared it that way, but I thought it was worth a shot... and bam, it works :)

rogerclarkmelbourne:
Hi Victor,

Thats fantastic news 

Well done.

Can you post the basic source code that you used for testing, or is it just the same ??

I thought the ISR wasnt being called at all, but it sounds like it was being called, but didnt stop the transfer ??

Thanks

Roger

victorpv:
Roger, when trying to add DMA support to the library, it fails to compile because DMA registers, functions, etc are not declared in that scope. 
What would be the best/most compatible way of including the DMA stuff?
Thanks

rogerclarkmelbourne:
Hi Victor

yes, Just include the dma stuff

The only issue may be the SPI

I added another function to my SPI class called dev() that returns the SPI hardware device
So that the device can be accessed by anything using that SPI device

However I have not uploaded that change, as I was still messing around with the DMA stuff yesterday and had not got it working

I'd accept a change like that to the SPI library if required

victorpv:
The DMA ISR was being called, but it modified a variable to inform the rest of the code that the DMA was finished and could continue. The problem was with that variable, it was not getting modified until I changed it to volatile. I am not totally certain volatile was needed or just the compiler was giving me a hard time, but from now on I will use volatile for any variable I modify in ISRs. 
My guess is that the compiler thought that function was not being called because there isn't a normal call for it, so it was never taking care of what it could do with a variable. That is just a guess.
I have not touched the SPI library, I would think that any change you make in that regard should not affect DMA.
So far I have included DMA functions in the ILI library with a pretty good speed up in screenfills, lines, rectangles. I am trying to track a bug with rounded rectangles, and then will send you a pull request.

The code itself is the last one I posted, just brought that to the library rather than having an extra function in the sketch, and applied it to fastVline, fastVline, and Fills. 

The only pending things I belive are:
1.-Track that bug with rounded rects, the code hangs, not sure if there is a problem with DMA itself or I inadvertely changed something else.
2.-Try using the tubes, as that would give us a common API for F2/F4 devices.

I am thinking if we should kept this library with the same name, or rename it, as it is getting heavily modified from the original. What do you think?

victorpv:
Corrected the bug, there were some functions in one of the test sketches calling to draw lines with 0 height or 0 width. Such a line would not draw a single pixel in the screen, so they were causing the DMA to try to transfer 0 bytes. So the DMA would never finish because it would never send a single byte.
I added a couple of checks in the fastHline, fastVline and rectangle routinges (the only three that use DMA) so whenever height or width is 0 it just return, as there is nothing to draw.
I am sending a pull request with the modified libraries, have a look at them.
I think if you used a good amount of ram for buffering, and double buffering so you can fill one from the camera while the other is going to the screen, you should get a fairly good performance, mostly limited by the SPI bus speed.

victorpv:
Well just checked and github is including all my changes in this commit. I renamed some files, and removed the old ILI9341 library, left those files in a zip in case I needed them (they are in my commit by mistake).
The ILI library is the one you gave me before in the arduino forum, with the DMA functions added in conditional lines for the STM32, so unless support for other micros was already broken, it should still work fine for them.
DMA seems to work fairly good, I feel confident we can accelerate anything that moves blocks with SPI. Sadly the fonts have a lot of pixel drawing, and any line in an angle is made with pixels too, that I don't think we can accelerate with DMA unless we made buffers for big chunks of the screen. Still gains a fair amount of speed. You can compile your stm32_graphictest sketch with the new library and the old one to see the difference in some parts.

victorpv:
Ok, I just changed a few things. All the examples compile and work using the new library just fine.
I named the new library _STM to continue with what was done with the Lcd7920 library. 
At least in my case, I like to know that I am compiling with the optimized library, rather than one that I may have in other folders for the AVR. The only change needed is to include the correct file, and the cs, rst and dc pins, and any sketch should work.
I think that if months from now I download another version of the Adafruit ILI library for whatever reason, it will help having different names, otherwise I could find sketches compiling with the wrong one.

victorpv:
The ILI9341_due modified for STM is included in this pull request so you can test it.
I posted in the original thread to that library asking if the author wants to incorporate the changes, if not, then we just keep our optimized version.
It does a few of things you had commented before, like only toggling CS when really needed and not between 2 data transmissions, so to do that it adds a bunch of other functions to break things down. I did not compare the compiled size of a sketch with this one and with our previous version, but without extra fonts I would think this one is a bit bigger. If speed is needed, it is worth every bit, is super fast.

rogerclarkmelbourne:
Pulled. Thankyou

PR #37, opened 2015-03-22
Added setRotation (uint8_t) for compatibility

victorpv:
2 small changes:
The author of the due library had turned setRotation to only use a
certain class as input, breaking compatibility with any sketch that used
an integer to set the rotation, so I added that function that in return call his.
Please check the change if you have time to see if there is a more elegant way.

In the arcs example I corrected an include line.


victorpv:
Trying to sync properly with your branch first, I am pulling this request

rogerclarkmelbourne:
Hi Victor

I branched to to do initial testing, then merged the branch into the master

So your latest stuff from the pull should be in the master

Thanks

Roger

PR #55, opened 2015-05-07
Issue47

RickKimball:
Empty sketch shrinks nicely:

    Sketch uses 3,472 bytes (5%) of program storage space. Maximum is 64,000 bytes.
    Global variables use 1,928 bytes of dynamic memory.

Blink:

    Sketch uses 4,092 bytes (6%) of program storage space. Maximum is 64,000 bytes.
    Global variables use 1,928 bytes of dynamic memory.

ASCIITable:

    Sketch uses 6,720 bytes (10%) of program storage space. Maximum is 64,000 bytes.
    Global variables use 2,288 bytes of dynamic memory.


rogerclarkmelbourne:
See 

https://github.com/rogerclarkmelbourne/Arduino_STM32/issues/56

There are already some fixes for some of those items, which got put into leaflabs/libmaple after to diverged 

Its probably worth looking at their solutions  ;-)

Thanks

Roger

rogerclarkmelbourne:
umm
I'm loosing track of which branch this is in

I was pretty sure I did the update for the size calculation in one of the branches, I'll need to double check which branch its in. as I know its a safe change / low risk, I don't seem any problem with using it

rogerclarkmelbourne:
Rick

BTW. I don't commit unrelated changes at the same time. It makes it difficult/ impossible to only revert one of the changes.

So I'll need to do them as 2 separate commits

rogerclarkmelbourne:
I've now done these as separate manual updates and commits

victorpv:
Did the changes proposed by Rick for the sketch size work?
I see you have applied some of those changes, but not sure if everything need to reduce the sketch size.

rogerclarkmelbourne:
Victor

It doesnt change the size of the sketch, the fix was for the value that was displayed.

Apparently it was incorrect.

I applied his fix and checked and it looks like the size of the bin file is now what the IDE says the code size is

PR #58, opened 2015-05-22
SPI DMA functions.

victorpv:
Adds 5 new functions: DMA Transfer, DMA 8bit send, DMA 16bit send, write
16 bit int. and mode change to change between 8bit and 16 bit transfer.
On the existing ones I did small changes to correct some problems when receiving data, then added the 16bit ones that can send a single byte, a DMA buffer, and send a word in DMA not incrementing memory (to fill a device such as a display).
There is an extra function that changes the peripheral between 8 and 16 mode. Can be run on the fly, does not require to setup the SPI port again. You change the mode, do a 16 bit send, change the mode again, do an 8 bit send. It helps a lot with displays that use 16bit for colors, as it goes from 2 SPI transactions to only 1.

### DIFF
--- a/STM32F1/libraries/SPI/src/SPI.cpp
+++ b/STM32F1/libraries/SPI/src/SPI.cpp
@@ -183,6 +183,19 @@ void SPIClass::setBitOrder(BitOrder bitOrder)
 	this->begin();
 }
 
+/*	Victor Perez. Added to test changing datasize from 8 to 16 bit modes on the fly.
+*	Input parameter should be SPI_CR1_DFF set to 0 or 1 on a 32bit word.
+*	
+*/
+void SPIClass::setDataSize(uint32 datasize)
+{
+		uint32 cr1 = this->spi_d->regs->CR1; 
+		datasize &= SPI_CR1_DFF;
+		cr1 &= ~(SPI_CR1_DFF); 
+		cr1 |= datasize;  
+		this->spi_d->regs->CR1 = cr1; 
+}
+
 void SPIClass::setDataMode(uint8_t dataMode)
 {
 /* Notes.  As far as I can tell, the AVR numbers for dataMode appear to match the numbers required by the STM32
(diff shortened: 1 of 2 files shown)


rogerclarkmelbourne:
Hi Victor

It compiles for me, so I've merged the pull request.

I'm sure people will let me know if there are any problems ;-)

RickKimball:
I like new functionality. However, is there some way we can make it apparent to people who use these functions that they are extensions to the SPIClass not part of what you would find on other cores? People might use these, then try to compile the code for a different arduino core and it will fail.

I would prefer to see a SPI2Class or something?

victorpv:
I don't see how making it a different SPIClass would help with errors if they try to compile in other cores though.

Most of libmaple is hardware specific, as are most of the libraries in the project, so I think anyone trying to use them should already imagine they are specific, but we could add a wiki page about the new functions and then put a warning on it that they are specific. 

As far as I am concerned, it doesn't matter to me whether they are in the same class or a different one, but the main ones have been in the SPI library for a couple of months now, and several drivers use them, and I haven't seen anyone posting an question or issue about them not compiling for AVR or DUE or anything, so I would think it is safe to keep them there.

RickKimball:
I guess I'd like the see the SPI API consistent with all other Arduino SPI implementations. That way it is easier to move scripts across machines without the user having to change things.

If nothing else, you should spew a warning when you compile that the user is using a platform specific API method.

RickKimball:
As far as the libmaple hardware specific stuff, that is hidden behind a public API. The user doesn't see that.  If the user writes to libmaple they are risking cross platform compatibility.

This is why it is hard to use some of the libraries from the arduino/avr world.  They have hard coded avr specific code.  If they had stuck to the published API their library would just work with the stm32duino platform.

RickKimball:
I guess in reality none of this matters. They are no official published Arduino APIs. They just do what they want and change when they want. .. so nm

victorpv:
Writing to libmaple is often just required to use STM specific hardware, such as the peripherals. Even for GPIO, pinMode has more options than any arduino because the STM allows several different modes not supported in Arduino.
As we are not taking away any functionality from the SPI "standard" functions, or changing them, but adding completely new functions, I think it doesn't matter if we put them in another class or not. If anyone takes code that uses those functions and try to compile it on another CPU, it will fail.
To me is like using the extra pin modes in GPIO. If you use them, and try to compile on AVR, it will fail.

victorpv:
About the arduino API, I believe we tried to see if they had implement SPI DMA in the DUE to use the same function names, but sadly they had not so far, they examples we see if using the DUE is all doing low level calls like we were doing with libmaple, that's why we decided to "invent" new function names for those.
If you have found somewhere they are creating new functions for SPI DMA access, definitely would be good to know and we could use the same names and parameters.

rogerclarkmelbourne:
I will post to the IDE developers, but I suspect they won't have time to even think about a common API for DMA SPI, as they are so busy firefighting 

I the only comparable hardware e.g. due and teensy and PIC32 have not bothered to define a nice new API, so IMHO we may as well lead the way if no one else is willing to do it.

Rome wasn't built in a day and no matter what we do, it won't be perfect and won't please everyone, but at last for most people it will be an advancement

PR #62, opened 2015-05-24
Ili9341 update

victorpv:
Updated it to use the new SPI DMA functions, and also added some other speed improvements.
Now it changes the SPI port to 16 bit mode when sending coordinates or color data, and back to 8bit after the transfer. That speed up normal SPI transfers (not much difference when doing DMA transfers).
It switches the port back to 8bit right away, so if the port is used for some other peripheral at the same time it does not cause any issue.
I have tested several examples and the Pig-Scope sketch.


victorpv:
Cancelling, it is including more things than I intended.

PR #63, opened 2015-05-24
Ili9341

victorpv:
Updated driver to work with new SPI functions.
Also has some other speed improvements. Tested it with several examples and the PigoScope sketch.


rogerclarkmelbourne:
Victor,.
Is this the same fix you PM'ed me about ?

I'm not sure why, but I don't seem to get all the pull request emails from GitHub, I'll check my junk mail filters, it could be that some of them are getting marked as junk

I only noticed your 2 pull requests because of another request from Vasillis

victorpv:
This is a different one. This one you named _STM because it was highly optimized for the STMF1, with DMA transfers and all.
This one needs updating because otherwise it conflicts with the new SPI dma functions.

The other ILI library is the one that you and Ray ported before, that did not have DMA. But at one point you started modifying it and adding some DMA code to it, although not completely working. That one does not have _STM in the name, and I think we should revert to the version that was working fine, not using any DMA code. I still need to test that one.

This one in this pull, I have tested it with a bunch of different examples, and seems to work perfectly happy and is really fast.

rogerclarkmelbourne:
OK. I'll merge this one.  Thanks

PR #64, opened 2015-05-24
Re script

victorpv:
RAM size corrected in Rx series. RC in supposed to have 48KB, but I am using 64 without an issue. RD and RE are supposed to have 64.


rogerclarkmelbourne:
Done

PR #65, opened 2015-05-26
Unused included libraries in hello_STM.ino file have been removed.

Serasidis:


rogerclarkmelbourne:
Pull looks fine, I'll action it

PR #67, opened 2015-05-28
Ili9341 without dma

victorpv:


victorpv:
This one reverts the ILI9341 to the state it was working fine in March. I just tested with with the pig-o-scope.

rogerclarkmelbourne:
done

PR #68, opened 2015-05-28
Spi changes

victorpv:
This eliminates compiler warnings when the compiler doesn't know which one to use.
The same function works for both 8bit and 16bit writes (in fact the function is the same, but the 16bit version accepts both 8bit and 16bit input). Tested with both ILI libraries, the one with DMA that does 16bit writes, and the one without that only does 8bit writes.


rogerclarkmelbourne:
Done

victorpv:
Personally, I leave it because it is easier for me to go an uncomment a section of code than having to go looking for which commit modified it to then copy that back etc. Commits often affect multiple files and has is more work to dig around them.
I see loads of commented code on different libraries that is left there for reference on how it was before, and helps me seeing the differences.
But I am not a professional programmer, so I am pretty sure there are better ways than mine.
If you guys think is better to remove the commented out sections, I can do that for future pulls.

RickKimball:
I can see that while you are developing you would use that approach. However, once you have decided to commit that change. If it isn't important enough to completely remove all traces, then you probably shouldn't be taking it out.

RickKimball:
In github, there is a history of code changes for every file. If you want to see what happened with a file just click on that link .. for example:

https://github.com/victorpv/Arduino_STM32/commits/d165e4d01343b67e9bab5ac2099ddcd23e2677db/STM32F1/libraries/SPI/src/SPI.cpp

That shows that there has been much mucking around with the write function.

victorpv:
I knew about the history in github, but still have to go and check the commits to find our which one removed what. I guess it's fine for files that aren't modified too often, or not likely to need to be reverted, but what if it is a file that gets a lot of commits, and the code in needed 2 months from now and no one remembers the exact time a function was taken away?
Perhaps adding a comment on what date it was commented out and why right above the function that is commented, and then a clean up in a later date when it is apparent that didn't break anything?

victorpv:
Something like:

/*
* This function commented out on 5/28, and replaced with one with same name that takes 16bit input.
* Remove on next update if SPI library works fine with new function.
*/
//void SPIClass::write(uint8 byte) {
....

RickKimball:
Files shouldn't be changing unless they are broken, refactored, or redesigned.  Especially this one. This is a core file. It shouldn't be changing unless some glaring function is broken missing or exceedingly slow.

RickKimball:
To be honest I'm not sure why you removed that function. I guess I'd have to see the warning.

RickKimball:
Why doesn't our SPI.h look like this ?

https://github.com/arduino/Arduino/blob/master/hardware/arduino/sam/libraries/SPI/SPI.h

The only public functions we have should be the same as the Arduino one.  The details of reading and writing to the platform / variant peripheral shouldn't be exposed in the SPI class.

victorpv:
I changed that function to be able to use the 16bit mode in the SPI peripheral. 
The previous write function took only byte as input. Initially I added a new write function that would only take 16bit int. as input. So the compiled could decide which one to use based on the size of the parameter, but happens that many libraries take a 16bit int, shift it 8 bits, and then write it, with the intention of writting the top 8bit after being shifted to the bottom 8bits, so the compiler is expected to do conversion discarding the top 8bits, which should be 0 after the shift.
When the compiler was finding a function doing that, it did not know if it should use the write function taking 16 bit as input, or the one taking 8bit as input, and was giving warnings.
Given the fact that the peripheral doesn't care what you write in the top 8 bit of that register if is set to 8bit mode, as it is set most of the time, then the write function that can take 16bit input works for both cases. The compiler doesn't need to decide what version to use, instead it uses that one, if the peripheral is set for 8bits, it ignores anything over the bottom 8bits. If it is set for 16bit, then it sends 16 bits.
I tested it with a library that never uses 16bit transfers, and with one that does it heavily, and works worked as expected.
There is more details on that in the datasheet about the 16bit mode in the SPI peripheral.

About the content of the library in general, Roger could answer that question, but I think most of our code has been ported as is from libmaple, that developed it like that.

RickKimball:
The problem with making functions public that aren't part of the Arduino SPI API list of functions is that other developers will see them, use them (not knowing they aren't part of the normal SPI) and then their library will only work with that version of the SPI.  This makes it harder to come up with libraries that work properly cross platform.

victorpv:
I don't really care one way or the other, as there are many other things that are not standard between platforms, but it was Roger's idea to integrate the dma transfer functions in the library so they can be reused, rather than having to write them over again in the different libraries.

RickKimball:
Are you using those calls in your Library? the dma ones?

victorpv:
Just to try to understand what you think would be the right way, are you suggesting that we add those functions to a different library (let's just call it "advanced SPI" library)?
If so, people that use that "advanced SPI" library in their device libraries, will still have problems using that cross platform, as the "advanced SPI" library will not be available for DUE or Teensy, or AVR, and if they copy over the "other" library, it will also not work in that platform anyway. Is pretty much the same as trying to use the DUE dma in an STM, will not work, or trying to use the Teensy SPI FIFO, will not work.
Having something like this teensy optimized library instead? https://github.com/xxxajk/spi4teensy3

victorpv:
Those calls are being used in the libraries I adapted for the STM, in the STMF1 part of the code.

victorpv:
Before those functions were added to the SPI library, I was just setting up DMA within the library, same way as the ILI9341_due library does for the due. I believe the _due version of the library that I adapted still has the dma setup in it rather than calling those functions, I don't remember right if that one has been updated yet.

RickKimball:
Yes, I think the right way would be for the SPI API to track the current DUE SPI library as we are closer in architecture to that than the avr chips. If you look at the public API for  the DUE SPI, there aren't any DMA things in there
https://github.com/arduino/Arduino/blob/master/hardware/arduino/sam/libraries/SPI/SPI.h

victorpv:
Sounds like a good idea.
Roger, what do you think? as you are the owner of the repo you have the last word on how the code should be organized.
Rick, regarding bringing the code on par with the due, do you think you could add the transaction stuff in which we are lagging way behind them? I don't think anyone in our forums has been able to tackle that down yet.

RickKimball:
Yeah I've been looking at doing that and then I gacked when i started looking that SPI.h

victorpv:
You have to remember this is all based on leaflabs' work from years ago, not everything looks exactly like arduino. That starting with their SPI which was called HardwareSPI. No library using SPI would work out of the box with that without adding at least a few lines of code, same with the USBSerial stuff...

victorpv:
And then there is no easy way to find why they did this or that, as they have officially stopped supported libmaple a couple of months ago... so their decision process for many things is just a mistery.

RickKimball:
My impressions is they really didn't like the Arduino API and were hoping people would use libmaple from the c level.

rogerclarkmelbourne:
As far as I can tell, the Arduino IDE team have no interest in supporting advanced features in the Due etc.

However I don't see why this should make us not use DMA if its available, or the backup registers etc, or CAN bus, or I2S etc.

We could propose interface additions to the IDE developers mailing list or possibly to the IDE issue in github, but if you look at the number of unactioned pull requests to the IDE, there are 110 currently pending enhancements that have not been actioned,
So I think we have a next to zero chance of them bothering to add new functionality even if we make it compatible with the Due for them to use


These changes are additions to SPI rather than affecting core functionality and see the best compromise in a difficult situation.

PR #70, opened 2015-05-28
fix SetupClock168() for Netduino2Plus + add STM32Stamp/Netduino2Plus …

martinayotte:
…boards + add dfuse_addr in upload


rogerclarkmelbourne:
Martin, 
The tools folder is shared with the F1 and F3, so platform.txt will need to be modified in both of these, to pass the extra arg even if its not used.

Also you only appear to have updated the linux script.

The OSX script and Windows bat file would also need to be updated, even for F4

martinayotte:
Hi Roger,
If you look to the tool/linux/maple_upload, I added a "if" statement for the fifth argument, so it is backward compatible, so no hurry to add it for F1/F3 yet, except if they becomes requires.
For OSX tool/linux/maple_upload, I've made the additionnal change, do you need a new PR ?
But for Windows, it is completely different, it is using maple_loader.jar, so I don't know what to do, especially that it seems to pass 9 arguments.

rogerclarkmelbourne:
Martin,

OK.

I'll pull manually to local branch and see what happens on windows etc

No need to do another PR

rogerclarkmelbourne:
Martin,

I've tested adding the {upload.dfuse_addr} to platform.txt and it doesnt seem to cause any issues on Windows, i.e it seems to be ignored by the Jar file that processes the DFU upload.

I'm not convinced that putting a processor specific #ifdef in the code is the best method to switch this

I'd rather put in in a #define for the value of that clock, and probably make another couple of variants, or at least put that clock frequency as an argument in the boards.txt, but although it seems a bit wasteful, I favor making more variants

After all the board that needs this is a 405 not a 407.

martinayotte:
Hi Roger,
I understand your relunctance against #ifdef, specially that I faced that myself in the SPI case.
But, adding this in another way can be more boring for other boards than having only one #ifdef for this exception. Note that it is not an exception for F405 against other F4*, my STM32Stamp has also a F405, but it still has the default 8MHz. The exception comes from the fact that Netduino2Plus designers didn't which to have 2 oscillators, one 8Mhz for CPU and one 25MHz for the EthPHY. So there are maybe other F4* boards that are sharing oscillator, maybe a F429. We will see when news boards will be added ...

rogerclarkmelbourne:
I'd rather we added something like build.PLL_M or build.PLL_M to all boards and change platform.txt to add -DPLL_M={build.PLL_M}

a bit like we do with the CPU freq

PR #71, opened 2015-05-29
Added a fix for booting the VS1053B boards into mp3 decoding instead of booting into MID (modeSwitch function).

Serasidis:


RickKimball:
Why are you redefining the PROGMEM defines?

Serasidis:
Could you be more specific ?

RickKimball:
There is no reason to use PROGMEM on the stm32 platform. If you want to store something in flash just use:

static const char SOME_STRING[] = { "ABCDE"}; 

Same goes for that huge chunk of unsigned int for the HelloMP3:
Use this: and it ends up in flash.

static const unsigned char HelloMP3[] = {
  0xFF,0xF2,0x40,0xC0,0x19,0xB7,0x00,0x14,0x02,0xE6,0x5C, /* ..@.......\ */

The way you have that one now it ends up in both .data and in the ram. 

You should not use PROGMEM for code that is meant to run on the STM32 platform.

Serasidis:
You are right! Thanks Rick!
I tested it and it worked. I will modify the **hello_STM** example.

rogerclarkmelbourne:
I did this fix manually, as merging didn't seem to work

PR #74, opened 2015-05-29
Added a fix for booting the VS1053B boards into mp3 decoding instead …

Serasidis:
…of booting into MIDI (modeSwitch function).


rogerclarkmelbourne:
Vasillis

Are you sure this compiles?

You seem to be calling modeSwitch() but there isnt a function called modeSwitch() in the library

Do you need to push some other changes to the repo ?

rogerclarkmelbourne:
I did this fix manually, as merging didn't seem to work

PR #78, opened 2015-05-31
Weakrefs

victorpv:
isrs.S edited to remove all the weak and use weakref instead.


rogerclarkmelbourne:
Victor,

This doesn't compile for me. I get a load of errors about duplicate definitions for the timer irq's

PR #81, opened 2015-06-06
timer merge

victorpv:
From my tests merging all the functions from timer.c and timer_f1.c in a single file solves the issues with code missing when whole_archive is used.


rogerclarkmelbourne:
Thanks

I will give it a try

rogerclarkmelbourne:
Victor

The pull request doesnt compile for me

I think it also requires the pinmap in flash changes, or at least the change that has pointers to objects rather than the objects themselves

As I get errors like


C:\Users\rclark\Documents\Arduino\hardware\arduino_stm32\STM32F1\cores\maple\libmaple\timer.c:52:18: error: conflicting type qualifiers for 'TIMER1'
 timer_dev *const TIMER1 = &timer1;


I'm happy to change the code to use pointers to objects, but I'm not sure putting PIN MAP in flash has been generally accepted as a good idea

(Well I'm reasonably happy with PIN MAP in flash, but I know Ray objects as he thinks it slows things down a bit)

victorpv:
Sorry my mistake, cause I had modified those files for pinmap in flash before.

The change just consist on taking all the content from timer_f1 and moving it to timer.c, no other changes needed, so you can just edit your files. The include lines on top are the same for both, so just move all the functions over and leave timer_f1 empty. That should work fine with the original isrs.S without other changes.

--- Original Message ---

rogerclarkmelbourne:
Victor

No worries

I'll give that a try

Thanks

rogerclarkmelbourne:
I went back to the previous version as I think we all agreed it was the safer thing to do

PR #83, opened 2015-06-10
Example of using the SPI_1 and SPI_2 ports on STM32F103.

Serasidis:


rogerclarkmelbourne:
I think people will object to using the direct gpio calls in this example

The example should just be about SPI and use normal digitalWrite etc

You could add a separate example about libmaple gpio calls if you want, but we may move away from libmaple at some point, so including libmaple specific calls in sketches is probably not "best practice"

Serasidis:
OK, I will remove the direct gpio calls.

PR #84, opened 2015-06-11
Example of using the SPI_1 and SPI_2 ports on STM32F103.

Serasidis:


rogerclarkmelbourne:
Vasillis

I've modified your example, so that it actually uses both SPI channels, rather than having SPI 2 commented out.

This will work on all boards, even the C8, so everyone should be able to give it a try

Serasidis:
Yes, it is better that way.
Thanks!

rogerclarkmelbourne:
No worries

I was also thinking of using pointers to the 2 classes but thought this would be too advanced for some novice users.

PR #86, opened 2015-07-01
FLAC decoder patch

tjstyle:
By uncommenting "#define USEFLAC" in VS1003_STM.h the FLAC patch will be
loaded at begin time.


rogerclarkmelbourne:
Thanks

PR #90, opened 2015-07-09
Update PIN_MAP to allow PWM on PB9

cdrose:


rogerclarkmelbourne:
Looks generally OK, but if you look at the files changed, it seems to have picked up some other editing to do with changes in tabbing

Ideally there would just be one line changed, but I suspect the other lines are just formatting

rogerclarkmelbourne:
PS. Its best to put a comment when you commit into your local repo copy, as this gets pulled though as part of the PR

I use git commit -am "My commit message"

cdrose:
Ah right, sorry. Yeah I fixed the tabbing out of habit. So I need to make sure to explain all changes made so you can easily see if they are appropriate or not.

rogerclarkmelbourne:
OK. I think it depends on what editor you use. It looked OK in notepad++ to me, but it could well have had tab issues

PR #96, opened 2015-08-01
HardWire bug fixes

fergul:
Hi, 
Whilst playing around with the hardware based I2C on the Maple mini I found a couple of bugs within the existing HardWire implementation causing it to not work correctly with the i2c_scanner sketch (also discussed here -> http://stm32duino.com/viewtopic.php?f=27&t=71).

The first bug is related to not accounting for only sending an address, as used in the scanner. The second is that the HardWire process function always returns a 0 status, incorrectly identifying errors as a success. They have both been fixed and endTransmission now correctly reports the result of the I2C transmission. I've also added the i2c_scanner sketch to the Wire examples folder to help with testing both Wire and HWire.

Lastly, dfu-util won't compile correctly on Linux due to the dos line endings in the autogen script which I have now fixed.

Let me know if you would like me to change or clarify anything. 
(p.s. I'm in the process of signing up to the forum to push a few other libraries I've ported as well)


rogerclarkmelbourne:
Ok. I will take a look.

Re:dfu-util not compiling.
Thats odd, I compiled it in Linux Mint (32 bit). I will get one of the other users on STM32duino.com who uses linux to see if they have the same issue

Re:I2C issues

The scanner was working Ok on software I2C, but I've never used HardWire.
When i created the repo, last year, using code from leaflabs/libmaple, there was a bug in Software I2C which prevented the scanner from working, but from what I recall, it was not related to the return value, it was a problem with reading or resetting the data pins.

Anyway, i will investigate both.

rogerclarkmelbourne:
Thanks.

I tested both Wire and Hardwire scanners and also tested the changes to Hardware using a BMP180 barometer module and a version of the Adafruit bmp085 library, which I changed to use Hardwire, and everything seemed to work fine.

The only change I made was to add Serial.Begin(115200) to the examples, so that they work on boards where the user is using a USB to Serial converter

fergul:
Great, glad to hear it worked.

rogerclarkmelbourne:
No worries ;-)

PR #98, opened 2015-08-11
convert USBSerial and HardwareSerial to derived from Stream instead of Print

martinayotte:
convert USBSerial and HardwareSerial to derived from Stream instead of Print


rogerclarkmelbourne:
Thanks.

PR #99, opened 2015-08-21
fix bug about DhcpClass instantiation to save 40K by avoiding link with GCC 'new' memory allocation code

martinayotte:
fix bug about DhcpClass instantiation to save 40K by avoiding link with GCC 'new' memory allocation code


rogerclarkmelbourne:
Looks OK. So I've actioned the PR. 

Thanks

PR #101, opened 2015-08-25
Pin map in flash

victorpv:
This should bring pin_map to flash.
I haven't tested this commit extensively, but I did was first resync back to you, and then copy over the changes from a folder that had pin_map in flash and I have used it for several months with timers, adc, etc all working, so should work right away.
If you get any error let me know and I go comparing files again to make sure I didn't leave anything behind.

Now, little trick. If in boards.cpp, you define pin map with this attribute:
extern const stm32_pin_info __attribute__ ((section (".data"))) PIN_MAP[BOARD_NR_GPIO_PINS] = {

you are forcing pin_map back to RAM, but still having it completely initialized before any code is run. So you get the speed improvement of RAM, but with the advantage of a table fully populated from start, which should help with any constructor trying to use it.


rogerclarkmelbourne:
Thanks victor

I'll do a local branch and pull, and see if it work for me

Cheers

Roger

rogerclarkmelbourne:
Victor

I've run some tests, and the PIN_MAP array is in flash, but its full of pointers to RAM. 

e.g. If I do
'''
  Serial.println((int) PIN_MAP,HEX);
'''

I get 

800B608

i.e PIN_MAP is in RAM

However if I look at the contents of the GPIOA I get

20000C78

which I can also see by doing

  Serial.println((int) PIN_MAP[0].gpio_device,HEX);

So some of this is still in RAM

I can't really see how the compiler can do this really, I guess it knows the location of these the gpio structs e.g. GPIOA in Ram and can write it into the the Flash

But I wonder when GPIOA gets initialised, 

I suspect that its not necessarily initialised until after the global constructors.

Actually I wrote a quick test sketch to save the location of GPIOA in a class then made a global version of it

'''
class test
{
  public:
  
  int gpioLocation;
  test()
  {
    gpioLocation= (int)GPIOA;
  }
};
test myTest;
void setup() {

}
void loop() {
  Serial.println(myTest.gpioLocation,HEX);
  delay(1000);
}
'''

And in this test GPIO is available in the global constructor, but it doesn't seem to be valid test, as I tested it in the master branch of the repo and GPIO was in ram at the same location.


Anyway. 
I did some other tests running SPI etc and everything seems to work fine. Also I'm using the ADC in my tests

So...

I think its probably safe to release to the master version

But I'd also like to see if I can move GPIOA etc into Flash as well at a later date

PS. I did try changing

'''
gpio_dev gpioa = {
    .regs      = GPIOA_BASE,
    .clk_id    = RCC_GPIOA,
    .exti_port = EXTI_PA,
};
'''

to
'''
gpio_dev const gpioa = {
    .regs      = GPIOA_BASE,
    .clk_id    = RCC_GPIOA,
    .exti_port = EXTI_PA,
};
'''

i.e added const to it, but the core no longer compiles, and would need a lot of changes everywhere that a gpio_dev is passed (which is lots of places)

victorpv:
I remember we discussed that before at some point. The core would need a lot of changes to make it all const, which would then bring all that to flash. But I found that even though those structures are in RAM, they are defined at compile time, so the compiler puts them in .bss area in flash, which is copied to RAM as soon as the MCU starts running, with some assembler code, before C code is called.

To bring all that to flash, I would rather declare a new pointer that is not const but using the original name, and points to the struct which gets to be const, but with a new name. I think that would waste a few bytes of memory but save us from extensive changes to the core. Something such as:

gpio_dev * gpioa = gpioa_struct
gpio_dev const gpioa_struct = {
.regs = GPIOA_BASE,
.clk_id = RCC_GPIOA,
.exti_port = EXTI_PA,
};

It's too early so probably I didn't think something right, but you can get the idea.

PR #105, opened 2015-08-29
Fixes for Nucleo STM32F103RB

LieBtrau:
1. Mapping USART2 to Serial instead of USART1
2. WireBase write functions return type set to boolean, so that calling functions can check that there's no buffer overflow.

### DIFF
--- a/STM32F1/libraries/SPI/src/SPI.cpp
+++ b/STM32F1/libraries/SPI/src/SPI.cpp
@@ -65,24 +65,24 @@ static spi_baud_rate determine_baud_rate(spi_dev *dev, uint32_t freq);
 #endif
 
 static const spi_pins board_spi_pins[] __FLASH__ = {
-#if BOARD_NR_SPI >= 1
+    #if BOARD_NR_SPI >= 1
     {BOARD_SPI1_NSS_PIN,
      BOARD_SPI1_SCK_PIN,
      BOARD_SPI1_MISO_PIN,
      BOARD_SPI1_MOSI_PIN},
-#endif
-#if BOARD_NR_SPI >= 2
+    #endif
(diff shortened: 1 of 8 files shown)


rogerclarkmelbourne:
The Nucleo changes are OK (I've checked with madias on the stm32duino forum

However the changes to I2C don't seem to be compatible with the Arduino API.

I've checked the Arduino API and it just seems to return size_t from all wirte functions, and this is just the size / quantity value that is passed in by the calling code (as far as I can see  by looking at wire.h for both AVR and SAM based Arduino boards)

See

https://github.com/arduino/Arduino/blob/master/hardware/arduino/sam/libraries/Wire/Wire.h and

https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/libraries/Wire/Wire.h

I agree that the return types are not correct, ie. should not be void, but I don't think they should be bool either.

In situations like this we normally follow what the Arduino SAM architecture does, as its also a 32 Bit ARM device

Hence I think that these functions probably need to return size_t

Or have I missed something ??

Note, if what you are trying to do is enhance the Arduino API, its best to request the change to core libraries to on https://github.com/arduino/Arduino and ask them for an enhancement.
But the general problem is that any changes to core libraries may brake lots of existing code that relies on the code operating as it does at the moment, (even if the way it works now could be improved)

Thanks

LieBtrau:
Hi Richard,
About the I2C, you're probably right about the fact that it should be size_t instead of boolean.  The function could then return the number of bytes that have actually been written to the TX-FIFO.  That's better than using a boolean.  To be honest, I didn't check the Arduino API.

Anyway, thanks for investing your time in supporting Arduino on the STM32.  Your work saved me a lot of time already.

PR #107, opened 2015-09-06
change IPAddress::toCharArray() into IPAddress::toString()

martinayotte:
change IPAddress::toCharArray() into IPAddress::toString() 
This is more obvious according to some discussion in ESP forums, so I made the same change here.


rogerclarkmelbourne:
Thanks

PR #109, opened 2015-09-08
add GPIOF/GPIOG pins to allow F429 using them

martinayotte:
Those GPIOF/GPIOG pins were missing ...


rogerclarkmelbourne:
Thanks

PR #113, opened 2015-09-11
Changed install.sh script to check for uucp group which is typical of Arch Linux

simonantonio:
Improved the linux install script to check if the [uucp](https://en.wikipedia.org/wiki/UUCP) group exists as is typical on [Arch](https://www.archlinux.org/) 

If the group is found it will copy the modified udev rules (Subsystem USB, Group uucp, Mode 0666)

See [Users and Groups](https://wiki.archlinux.org/index.php/Users_and_groups#User_groups) in the Arch wiki for groups.

plugdev does not exist on Arch


simonantonio:
dialout and plugdev do not exist on Arch,

rogerclarkmelbourne:
Thanks. I will ask @ahull to check this as I'm not a Linux person

PR #118, opened 2015-09-25
Fix file open to actually turn on inbound parity checking.

mwm:
The INPCK bit is an i_flag bit, not a c_flag bit as originally done
here. I have no idea what setting INPCK in c_flag does on Linux, but
on FreeBSD it caused the test of the c_values after they were set to
fail. This fixes that, and presumably the serial interface will now
actually check the parity of incoming data.


rogerclarkmelbourne:
Thanks.

I will need to get this double checked on some other flavors of Linux

PR #123, opened 2015-10-27
add generic STM32F103T[8B] variants

jcw:
As you suggested on the [forum](http://www.stm32duino.com/viewtopic.php?f=28&t=680)

Created the 103T from a copy of the 103C files.
No changes, other than renames and removing pins PB8 and above from `boards.{h,cpp}`.

I've also defined PA0 as the DM+ enabling pin, to match the board I have.


rogerclarkmelbourne:
Hi @jcw 

As the PA0 DM+ thing is specific to that HY-TinySTM103T board, I think we'd be better off calling the board by that name.

As @Markm on the forum has his own STM103T board and I think someone else also is building one, and neither of them use PA0, they use the direct GPIO pullup method to force USB bus enumeration.

(BTW, using an external transistor like that board does is more technically correct for the USB spec, from what I understand, but most generic boards only seem to use a pullup on USB D+, hence we have to have code to support this as the "generic" version.)

I'll do a pull and rename the folder and change the name in boards.txt if this is OK with you?

jcw:
Good points - absolutely!

rogerclarkmelbourne:
OK. I'll pull into a new branch, change the names and merge back in.

rogerclarkmelbourne:
Hi @jcw 
FYI

I'm just trying to work out how best to use your new folder for both the HY-TinySTM103T and also the home made boards that @markm and one other person are making.

It may be better just to call the folder stm32f103t  or stm32f103t_boards and switch the form of USB reset system from boards.txt

Otherwise we have 2 virtually identical folders.

I think its also possibly highlighted a potential small bug in the other generic boards, with the use of 

#define BOARD_USB_DISC_DEV        GPIOA
#define BOARD_USB_DISC_BIT        0

As I have a feeling that the other generic boards may be toggling

#define BOARD_USB_DISC_DEV        GPIOB
#define BOARD_USB_DISC_BIT        10

When this pin should be available as normal GPIO

So I'll hold off pushing the your changes back until I've double checked this

jcw:
There is a lot of duplication in the current setup, if you look at how much I had to copy over to turn an F103C variant into an F103T variant vs the few minimal changes to the boards.* files.

Could you perhaps check this in as F103T variant for now, minus the PA0 mod?

rogerclarkmelbourne:
Yes.

I can check it in. I don't think its going to do any harm ;-)

jcw:
That would be nice - I can then add a 10 kΩ pull-down on PA0 as temporary fix. Thx.

rogerclarkmelbourne:
There seems to be some sort of fault on their git server at the moment. 

I just asked me to login when I pushed, which is not normal, and when I give the correct login details, my authentication failed.

I have seen it do this before.

I'll leave it half and hour and try again

rogerclarkmelbourne:
Ah.  I remembered, I can do it directly inside github and I'll just delete my local repo (I have a copy of the local repo as well, which I can restore)

rogerclarkmelbourne:
OK
It should be merged now.

jcw:
Oops, small typo in that F103T board definition - in [boards.txt](https://github.com/rogerclarkmelbourne/Arduino_STM32/blob/master/STM32F1/boards.txt#L298), "jtag_c8" should be changed to "jtag_t8".

rogerclarkmelbourne:
Ok.

I will update my local copy and see if it will let me push the changes.

BTW. the names of these linker files is really a legacy from the original Leaflabs IDE, and in some places I have changed them to more sensible names ;-)

PR #126, opened 2015-10-29
run at 48 MHz off the internal RC oscillator

jcw:
This change only applies to the Nucleo F103RB board:

Make the system clock run from the internal RC clock i.s.o. a crystal (which is not present by default on a Nucleo). Set up the PLL to run at 48 MHz (72 is not supported, 64 is max).

(feel free to drop this pull request if you don't want to merge it)

### DIFF
--- a/STM32F1/boards.txt
+++ b/STM32F1/boards.txt
@@ -130,7 +130,7 @@ nucleo_f103rb.upload.altID=1
 nucleo_f103rb.upload.auto_reset=true
 
 nucleo_f103rb.build.mcu=cortex-m3
-nucleo_f103rb.build.f_cpu=72000000L
+nucleo_f103rb.build.f_cpu=48000000L
 nucleo_f103rb.build.board=STM_NUCLEU_F103RB
 nucleo_f103rb.build.core=maple
 nucleo_f103rb.build.extra_flags=-DMCU_STM32F103RB -mthumb    -march=armv7-m  -D__STM32F1__ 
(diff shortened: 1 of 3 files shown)


rogerclarkmelbourne:
I think if we have a define e.g. NUCLEO_USE_INTERNAL_CLOCK and then #ifdef the code changes so that by default NUCLEO_USE_INTERNAL_CLOCK is not defined,  we can then  use

the build.extra_flags, which are currently

nucleo_f103rb.build.extra_flags=-DMCU_STM32F103RB -mthumb    -march=armv7-m  -D__STM32F1__ 

to add the define for this, when using the new board variant.

Or add a new menu, which defaults to the external osc (in the first menu item)

jcw:
Ok, I'll see what I can do. But _please_ consider making the non-crystal version the default.

jcw:
I think this should do it. Two boards, a new one and the existing "w/ crystal" one.

rogerclarkmelbourne:
"I think this should do it. Two boards, a new one and the existing "w/ crystal" one.

OK

Can you update the PR so that there are 2 different boards, and use a #define to switch the clock osc speed and PLL settings etc

jcw:
I _think_ that's what the [last commit](https://github.com/rogerclarkmelbourne/Arduino_STM32/commit/0d0f910cbbb60b1d84ddce5e5b530e0346155a75) above, from does - two boards, first one called "STM Nucleo F103RB (STLink)", second one "STM Nucleo F103RB w/ crystal (STLink)".

Or did you mean one board, two variants? I'll need to figure out how to set that up.

jcw:
Please wait. I'll clean up the PR and fix the merge conflict. Then you can have a look again.

jcw:
On second thought - I'd rather make the non-crystal version run at 64 MHz, the max supported with HSI. There's no reason to support USB, as this requires an external crystal for stabiity anyway.

Will wait for your response, then resubmit a new PR for the development branch.

rogerclarkmelbourne:
OK.

As the Nucleo doesn't have a USB port connected to the MCU you may as well run it at 64Mhz.

BTW what do STM recommend, for its frequency of operation?

Does higher frequency mean less accurate timings ? (I don't suppose it does)

jcw:
I don't think I've seen any recommendations about this from STM. The datasheet mentions up to 64 MHz when running off the HSI oscillator, in the section of current consumption, for example.

Frequency accuracy is in %, i.e. relative - this stays the same regardless of PLL / divider use.

jcw:
Replaced by PR #138.

Issue #125, opened 2015-10-29
SPI default pin not available on F103Tx


jcw:
On the STM32F103Tx µC, pin PC13 does not exist (nor does the 2nd SPI device, in fact).
This is set as the default pin for slave select in [SPI.h](https://github.com/rogerclarkmelbourne/Arduino_STM32/blob/master/STM32F1/libraries/SPI/src/SPI.h#L91-L92).
Would it be possible to change that default to PA4?

The relevant change is this:
```
diff --git a/STM32F1/libraries/SPI/src/SPI.h b/STM32F1/libraries/SPI/src/SPI.h
index 9bfe4f4..b807a8a 100644
--- a/STM32F1/libraries/SPI/src/SPI.h
+++ b/STM32F1/libraries/SPI/src/SPI.h
@@ -88,8 +88,8 @@
 #endif
 
 // PC13 or PA4
-//#define BOARD_SPI_DEFAULT_SS PA4
-#define BOARD_SPI_DEFAULT_SS PC13
+#define BOARD_SPI_DEFAULT_SS PA4
+//#define BOARD_SPI_DEFAULT_SS PC13
 
 #define SPI_MODE0 SPI_MODE_0
 #define SPI_MODE1 SPI_MODE_1
```

PR #127, opened 2015-10-29
see https://github.com/rogerclarkmelbourne/Arduino_STM32/issues/125

jcw:
see #125

(feel free to drop this pull request if you don't want to merge it)


rogerclarkmelbourne:
Seems logical, 

But I'd like to test it first.

I think generally that this definition is not used

I there are a few lines in the code that reference this define, but they have been commented out.

We generally rely on the libs or code that use SPI to set and reset the CS pin that they need.

It also looks like it was changed to PC13 from PA4 at some time in the past (I may have changed it, but I can't remember, as there is so much code and so many changes)

Initially we had an issue with using PA4 as SPI CS because its the hardware CS pin and it clashed with GPIO use of the same pin, however several months ago we did a fix for this and make the pin GPIO only (and not hardware CS / SS aka   NSS)

jcw:
Yes, for master SPI use the pin probably has to be a GPIO, not a SS (slave select).

jcw:
Made one more change to avoid defining an "SPI2" device on the F103T.

This #ifdef should actually be defined across all variants - it's an example of where the different variants are starting to diverge due to copying i.s.o. sharing identical code. No big deal here, but the task of streamlining / reintegrating all the variants at some future date is only going to get harder.

rogerclarkmelbourne:
Does that example still compile

It looks like it would fail on this line if SPI_2 is not defined

 data = SPI_2.transfer(0x55); //Send the HEX data 0x55 over SPI-2

jcw:
Ah, oops. You're right - shall I change the code, or do you want to leave it as is and perhaps note that it won't compile on F103T's?

rogerclarkmelbourne:
Strange.

It pull fails with an error message 

'fatal: Couldn't find remote ref spi-fix-for-F103Tx
'unexpected end of stream

Strange.

If I look here

https://github.com/rogerclarkmelbourne/Arduino_STM32/pull/127/files

It says pr from unknown repository

Perhaps Github is currently having issues.

I'll need to remember to try it again tomorrow

jcw:
Oops - my fault. I've deleted and recreated my fork because of a fight I had with Git (and lost).  
Suggest closing this PR, then I'll retrieve my changes and submit as a new PR?

rogerclarkmelbourne:
The changes are simple.

I'll change them manually on my local copy and recommit / push to the development branch

jcw:
Ok, I'm closing this to avoid completely getting lost in branches and PR's. Insufficient Git fu, grrrr...

rogerclarkmelbourne:
I agree, there are a lot of PR's going on at the moment, But I think need to come back to this one and make sure it gets implemented

PR #128, opened 2015-10-29
Added missing Ethernet library filesMaster

Serasidis:


rogerclarkmelbourne:
I'm just checking with Martin, about whether the files he added to the F4 are better than the default Arduino ones.
Martin seemed to say he had improved one of them I'm waiting for Martin to reply to the forum

rogerclarkmelbourne:
Hi Vassilis

As I've not really had a definite response from Martin, I'll action your PR as its low risk

Serasidis:
Roger, if we have any future problem with those files I will fix it and send you a new PR.

PR #129, opened 2015-10-30
reorder variants so smallest mem size is first

jcw:
This changes the order of variants in the menu.

The reasoning is that the IDE selects these by default when switching board types. Code built for a smaller variant is most likely to run on one with more memory, but not the other way around.

### DIFF
--- a/STM32F1/boards.txt
+++ b/STM32F1/boards.txt
@@ -1,6 +1,5 @@
 # STM32 Boards
 
-
 menu.device_variant=Variant
 menu.bootloader_version=Bootloader version
 menu.upload_method=Upload method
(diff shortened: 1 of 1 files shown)


rogerclarkmelbourne:
Did you intend to change only the new generic T8 board, or some other boards ?

It looks like a lot of small things got changed?

jcw:
I swapped the T and C variants, and I cleaned up the whitespace. My apologies for making a mess of the first commit. The first + second commit together only change whitespace w.r.t. original version.

rogerclarkmelbourne:
OK.

It happens a lot.
Git seems to track the slightest change, even if the result appears he same visually ;-)

jcw:
Thx. I'm still fighting with git, after all these years of (naïve) use...

rogerclarkmelbourne:
I know... It takes a lot of getting used to.

Sometimes I screw it up so badly, I just delete my local copy and re-clone back from github

PR #131, opened 2015-10-31
wait for time to elapse instead of a cycle-counting loop

jcw:
see discusssion on the forum at http://www.stm32duino.com/viewtopic.php?f=3&t=695&p=7601#p7601

This change makes delay() wait for millis() to report the requested passing of time. The original code uses delay_us() in system/libmaple/include/libmaple/delay.h, which counts processor cycles. The delay will be too long if there is background interrupt activity.

### DIFF
--- a/STM32F1/cores/maple/wirish_time.cpp
+++ b/STM32F1/cores/maple/wirish_time.cpp
@@ -34,10 +34,9 @@
 #include <libmaple/delay.h>
 
 void delay(unsigned long ms) {
-    uint32 i;
-    for (i = 0; i < ms; i++) {
-        delayMicroseconds(1000);
-    }
+    uint32 start = millis();
+    while (millis() - start < ms)
+        ;
 }
 
 void delayMicroseconds(uint32 us) {


RickKimball:
I like this change. However, I'd caution you to look through the code and see where delay is used. In any of those found locations, it should be verified that the SysTick timer is running when that code runs.  

Also, any user code that is timing sensitive may break if the timing is really sensitive.  Not that that is a bad thing. Just a word to the wise that if you change this, it may create a wave of broken code complaints.

jcw:
The AVR version in the Arduino runtime uses the same approach, although considerably more sophisticated (on AVR, micros() also works from the hardware timer counter).

Careful! Your suggested code will fail when milliseconds + delay overflows (≈ 49 days).
As submitted in the PR, millis() will be wrong, but delay() will continue to work.

As for startup order and dependencies - yes, doing things in statically constructed objects is always a tricky business. I think the AVR setup for the timer is extra early to allow this, but IMO that's a race no one will ever win. Best advice I've seen is: put very little in the constructor, and do most initialisation during a "begin()" or "init()" call, placed in "setup()".

For reference: the STM's runtime init order can be found [here](https://github.com/rogerclarkmelbourne/Arduino_STM32/blob/master/STM32F1/variants/generic_stm32f103t/wirish/boards.cpp#L65-L76).

RickKimball:
Yes you are correct about the overflow... nm :)

rogerclarkmelbourne:
I presumed there was a reason that millis() were implemented (by leaflabs) in a time wasting loop rather than via interrupts.

It may be worth PM'ing @gbulmer on the forum and asking him, as he is the moderator on the official leaflabs forum, so has been involved with libmaple for much longer than any of us.

I recall posting to the leaflabs forum about this, and getting a response form @gbulmer which I think was words to the effect of , This is never worked that well....   
But perhaps thats just my understanding of what his reply was.

Either way, as the guys at Leaflabs, have declared that libmaple is no longer supported by them, its no use posting an issue on their original libmaple repo, as they will almost certainly not reply.

I've only had a few github conversations with Marti from LeafLabs and that was over 6 months ago, and even then they were not long conversations, as I gather than leaflabs have moved on to newer and better things and do not have time for even simple questions on libmaple

rogerclarkmelbourne:
Guys. I created a new development branch, so I guess I can pull this into there for people to test

Can you confirm that you have not found any issues with this since you issued the PR ?

jcw:
No issues so far, but I've not accumulated my different PR's in one branch to use together, so I'm not really testing everything optimally yet. This timer change seems safe to me. Could be a nice way to introduce the dev / bleeding-edge branch if other people want to test and review changes like these.

rogerclarkmelbourne:
OK.

Its been merged and pushed into the development branch

jcw:
Thanks!

stevstrong:
Just a small question:
instead
`
    uint32 start = millis();
    while (millis() - start < ms);`
wouldn't it more effective this way:
´´´    uint32 stop = millis()+ms;
    while ( millis() < stop );´´´
The condition processing would take less time, since it contains only one comparison instead of a subtraction and a comparison?

jcw:
There is a very subtle issue related to 32-bit int overflows (which happens after 49 days). Your suggestion would lead to a very long loop if `stop` has overflowed.

Correct code always trumps efficiency :)

The extra calculation is probably one µC clock cycle, but more importantly: in this case the whole point of delay() is to consume time. We'd just be twiddling our thumbs faster.

stevstrong:
> There is a very subtle issue related to 32-bit int overflows (which happens after 49 days). Your suggestion would lead to a very long loop if stop has overflowed.

I' think your version would also fail by this check...
>  in this case the whole point of delay() is to consume time. We'd just be twiddling our thumbs faster.

I agree :smirk:

RickKimball:
Set the millisecond clock to something close to the max number of a uint32_t and run it stevstrong, you will it it overflows properly.

stevstrong:
Of course, you'r right ... :thumbsup:
Never try to optimize code right after having lunch...

PR #132, opened 2015-11-01
corrected typo - ADC_CR2_TSVREFE

stevstrong:
seems to be not critical

### DIFF
--- a/STM32F1/system/libmaple/stm32f1/include/series/adc.h
+++ b/STM32F1/system/libmaple/stm32f1/include/series/adc.h
@@ -78,7 +78,7 @@ extern const struct adc_dev *ADC3;
 #define ADC_CR2_EXTTRIG_BIT             20
 #define ADC_CR2_JSWSTART_BIT            21
 #define ADC_CR2_SWSTART_BIT             22
-#define ADC_CR2_TSEREFE_BIT             23
+#define ADC_CR2_TSVREFE_BIT             23
 
 #define ADC_CR2_ADON                    (1U << ADC_CR2_ADON_BIT)
 #define ADC_CR2_CONT                    (1U << ADC_CR2_CONT_BIT)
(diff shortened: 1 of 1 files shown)


rogerclarkmelbourne:
We'd also need to change the example that uses it


https://github.com/rogerclarkmelbourne/Arduino_STM32/blob/2e9b4f3955aa648bcc78301529b74ea652a9f4d8/examples/MrBrunetteExamples/InternalTempSensor/InternalTempSensor.ino

line 22

Preferably we also update the GD32 and STM32F4 as well, but I can do those two

stevstrong:
OK, thanks.

rogerclarkmelbourne:
Steve,

I'm made a new branch called development and have effectively pulled your PR's into that branch

https://github.com/rogerclarkmelbourne/Arduino_STM32/tree/development

I'll post to the forum and perhaps some of the others will have time to pull and switch to that branch and test.

I did a quick test with a board that has an SPI LCD attached and it still works fine, however getting by SD test rig (breadboard) into a state where I can run it will take longer as its been in a cupboard for a while and also I took out the SD card to use for something else

stevstrong:
Hi Roger,
```
I'm made a new branch called development and have effectively pulled your PR's into that branch
https://github.com/rogerclarkmelbourne/Arduino_STM32/tree/development
I'll post to the forum and perhaps some of the others will have time to pull and switch to that branch and test.
```
It's a good idea, hopefully there will be some volunteers to test this branch.

rogerclarkmelbourne:
Hi Steve,

I think the problem is that hardly anyone uses Git to update their local copy :-(

And even less people have time to update and switch to the development branch.

So the only way I get extensive testing is to put stuff on the Master branch and wait weeks and weeks for people to update.
Or I just get issues when a new users downloads for the first time, which is not an ideal situation.

Anyway, I think this issue is basically resolved, so I'm going to close this PR so I can keep track of how many PR's still need to be actioned

PR #133, opened 2015-11-01
enable USB device when uploader is BMP

jcw:
Add -DSERIAL_USB -DGENERIC_BOOTLOADER to all variants when uploading with BMP - this unifies the sketch behaviour regardless of which upload mechanism is being selected for a sketch. This way, "Serial" always refers to USB for all the generic variants.

### DIFF
--- a/STM32F1/boards.txt
+++ b/STM32F1/boards.txt
@@ -185,12 +185,12 @@ genericSTM32F103C.menu.upload_method.serialMethod.upload.tool=serial_upload
 genericSTM32F103C.menu.upload_method.STLinkMethod=STLink
 genericSTM32F103C.menu.upload_method.STLinkMethod.upload.protocol=STLink
 genericSTM32F103C.menu.upload_method.STLinkMethod.upload.tool=stlink_upload
-genericSTM32F103C.menu.upload_method.STLinkMethod.build.upload_flags=-DCONFIG_MAPLE_MINI_NO_DISABLE_DEBUG=1 -DSERIAL_USB -DGENERIC_BOOTLOADER
+genericSTM32F103C.menu.upload_method.STLinkMethod.build.upload_flags=-DCONFIG_MAPLE_MINI_NO_DISABLE_DEBUG -DSERIAL_USB -DGENERIC_BOOTLOADER
 
 genericSTM32F103C.menu.upload_method.BMPMethod=BMP (Black Magic Probe)
 genericSTM32F103C.menu.upload_method.BMPMethod.upload.protocol=gdb_bmp
 genericSTM32F103C.menu.upload_method.BMPMethod.upload.tool=bmp_upload
-genericSTM32F103C.menu.upload_method.BMPMethod.build.upload_flags=-DCONFIG_MAPLE_MINI_NO_DISABLE_DEBUG
+genericSTM32F103C.menu.upload_method.BMPMethod.build.upload_flags=-DCONFIG_MAPLE_MINI_NO_DISABLE_DEBUG -DSERIAL_USB -DGENERIC_BOOTLOADER
 
 ########################### Generic STM32F103R ###########################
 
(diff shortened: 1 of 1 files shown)


rogerclarkmelbourne:
Sorry.
I can't action this because this is not what Rick Kimball (the original author of the BMP stuff) uses his BMP for.

He wants it defaulted off (as per his response to the forum)

I looked at how both of you could be accomodated and the only way appears to be by having 2 upload methods for BMP. One without serial (which needs to be the default) and one with Serial.

I don't think I can add this as a separate menu, as it would effect all the upload methods, and would end up defaulting USB Serial to Off for even the Bootloader upload ;-(

The main problem is the IDE doesn't support cascading menus. i.e We can't have a menu that only appears when BMP is selected

This has been a problem for some time, hence the proliferation of menus, ( and also some hacks like setting the linker file flash size to the max of all types in that series of processors)

We'll need to discuss options with Rick on the forum before I can go ahead with this.

jcw:
Oh? I thought he agreed to it? See [this post](http://www.stm32duino.com/viewtopic.php?f=28&t=117&start=40#p7675) on the forum. We can continue discussion there, if needed.

jcw:
To continue to to use BMP for serial debugging, make your debug code use an adjustable reference:
````
#if DEBUGGING
#define Log Serial1
#else
#define Log Serial
#endif
...
Log.println("a debug print");
```

rogerclarkmelbourne:
Hi Jean-Claude

This will require existing users of the BMP to modify all their code.

I can't make changes like this which affect existing users, as I usually get a lot of complains if I do things like this.

Sorry, if the pace of change seem glacial, but I have to consider the existing users.

jcw:
Ok. I consider this a design mistake, but I do understand the tyranny of backwards-compability.

For me, it's unfortunately a show stopper, as I'm going to use and document the BMP as uploader of choice on the jeelabs.org weblog, and until I have a satisfactory solution for uploads through USB, I will probably need to maintain a fork / branch with this change in it.

One way out would be to add a second BMP variant for all the boards. Nasty, but not impossible. For me personally, it would be sufficient to do this for just the F103T variant, but that's even more of a kludge. I could create a separate PR for this, if you like.

rogerclarkmelbourne:
I think a second variant of BMP with serial support is probably the best approach to maintain backwards compatibility.

I guess I can put it to the vote, but I don't know how many people regularly read the forum, and its a pain to PM everyone that has posted about the BMP.

On a similar backwards compatibility issue, there is a hack where Serial.begin() is called in the init code for USB Serial. This is what LeafLabs did, however for Hardware serial you need to call it yourself.

I'm tempted to change this so that you need to call Serial.begin() for all serial types, and for usb serial, this would cause the re-enumeration.

The advantage of this, is for the people who only want HID, as it can probably be integrated into the core, as if they call HID.begin() , or some function like that, and don't call Serial.begin() the board will just be a HID device (and the same applies to Midi etc)

Its not an ideal way to support different USB devices, but its a good short term solution until someone work out how to build all the stucts needed for the usb composite device (Windows )

jcw:
Calling begin() for all Serial* objects you use sounds like a solid convention to me. It's what I think people are expected to do on Arduino Mega's etc as well. Making the call idempotent (i.e. silently ignoring a repeated call) could be added later in case there is still confusion over who inits each device.

I don't know anything about non-serial USB, so can't comment on that.

RickKimball:
I'm using a fork of this tree that has the BMP setup the way I use it. I don't have any problem with you changing this scheme fits to better accommodate other users.

rogerclarkmelbourne:
Thanks Rick

I'll do a poll about Serial.begin(), but hopefully we should not have too much opposition to making it more compliant with the core Arduino API than it currently is

rogerclarkmelbourne:
I think I uncovered a problem with not enabling by default :-(

Reset for upload won't work.

I can see issues for anyone who doesnt realise this is how the reset works.

I wonder how the Due handles this. I'll need to check

jcw:
Replaced by PR #136.

PR #136, opened 2015-11-05
enable USB device when uploader is BMP

jcw:
re-applied changes, for the development branch, closes #133 PR


jcw:
Looks like my other fix for the 17% cycle-eating bug got inserted into this PR by accident.
I'm sorry for the mess, but the two changes in this PR could be merged now, I think.

Still trying to increase my github-fu, not very successfully so far.

rogerclarkmelbourne:
I've manually added your fix into the USB Serial, into the development branch

I don't want to pull the changes to boards.txt at the moment, as I need to double check what going on.
I thought that STLink did originally have USB Serial enabled, but it looks like part of your change is to add this.  

So I'll need double check whats going on.

jcw:
I haven't changed any ST-Link behaviour. I just removed the redundant "=1" settings from those command line options. This fix only adds the Serial device via USB when using BMP.

This is only for generic F1 boards, I'm not touching any of the others. If you add a label to the GitHub issue tracker called "F1" or "STM32F1", then you could flag these issues/PRs to clarify. Or I can add "[F1]" in front of the title, if you prefer.

RickKimball:
Originally I had added that define with a value. My thinking was that in the future I would use it to be able to disable the SERIAL_USB by setting it to 0. Sometimes it is easier to set a value than try and undefine it.

jcw:
A bit confusingly, "-Dblah" on the gcc command line is equivalent to "#define blah 1" in the source.

RickKimball:
Try compiling this code using the different defines
```
#include <stdio.h>

#define SERIAL_USB
//#undef SERIAL_USB
//#define SERIAL_USB 0
//#define SERIAL_USB 1

int main(int argc, char **argv)
{
#ifdef SERIAL_USB
  printf("SERIAL_USB is defined\n");
#endif
#ifndef SERIAL_USB
  printf("SERIAL_USB is not defined\n");
#endif
#if SERIAL_USB == 1
  printf("SERIAL_USB is 1\n");
#endif

}
```
I get  compiler error with the code above:
```
/tmp/testme.c: In function ‘main’:
/tmp/testme.c:16:16: error: operator '==' has no left operand
 #if SERIAL_USB == 1

jcw:
Yes, in source code, you have to specify the value. On the command line, "=1" is added by default.

RickKimball:
Interesting, you learn something everyday

feurig:
Since this is an lvalue issue shouldnt the code be.
#ifdef SERIAL_USB
#ifdef SERIAL_USB == 1
xxxx
#endif
#endif

RickKimball:
$ gcc -DSERIAL_USB testme.c 
testme.c: In function ‘main’:
testme.c:16:19: warning: extra tokens at end of #ifdef directive [enabled by default]
 #ifdef SERIAL_USB == 1
                   ^

jcw:
As you can see, #ifdef doesn't take an expression, just a token.
You can use either of the following:

    #ifdef USB_SERIAL
    #if USB_SERIAL

The second one would be my choice. It hinges on another crazy "convention" in C & gcc, where "USB_SERIAL" is replaced by "0" in an #if, if it has not been defined anywhere.

The "#ifdef" was the only pragma in very early C implementations, "#if" etc came later. I tend to move away from all #ifdef uses in my own code, and try to avoid the extra "defined(...)" where possible. In many cases, the rule that undefined names are treated as zero in an #if tends to work fairly well.

Yes, it's crazy stuff in this day and age, but hey, I didn't make this up, I'm just the messenger :)

rogerclarkmelbourne:
JeanClaude

I can't action this PR as its referenced against the Master branch of the repo and some of the changes have already been implemented in the development branch

Also. The same PR changes loads of unrelated things. This is not good practice, as it makes it harder for me to revert individual changes.

So. Sorry, if this is a pain, but you'll need to remove this PR and generate some new ones, referenced against the development branch, and also, preferrably have separate PR's to do the separate things.

And I suspect perhaps half the things you want to do have already been done (in the development branch)

Thanks

Roger

jcw:
Fair enough - yes, it's gotten messy, as this was done before the development branch got created, I think. Still fighting mightily with git, or more probably the twisted understanding I have of it. Closing.

RickKimball:
Can we add an option to not use a bootloader? Maybe another menu item?

rogerclarkmelbourne:
We could add this as a menu, I think..

The menu item would need to change the linker setting and also the vector table, but we do this implicit depending on the upload type.

I just wonder whether adding another menu is just going to make things even more complicated, as we already have an upload type option and the sub-type of the series (i.e normally this just changed the flash memory size)

But if you are using USB to Serial and the internal bootloader, it seems strange to also have to set that you don't want to have it configured for the software bootloader. 
For STLink and BMP I can see the benefit for this is you wanted to debug something using either tool, and for some reason still wanted the booloader, but this seems a very rare occasion.

PR #137, opened 2015-11-05
fix SPI reference for F103T

jcw:
The SPI library references PC13 in its header, which is not present on all F103 packages. This PR changes it to PA4, which is the default for SPI1 i.s.o. SPI2. Supersedes #125, which was for the master branch and got a bit messed up.

(I hope I got the details of this PR right this time!)


rogerclarkmelbourne:
I'll take a look (probably at the weekend when I have some free time)

rogerclarkmelbourne:
Looks OK

I'll merge into development

rogerclarkmelbourne:
Done ;-)

PR #138, opened 2015-11-06
support Nucleo without and with crystal

jcw:
This has now been set up as two variants on the same Nucleo F103 menu entry.
The default, i.e. first variant is the official crystall-less version @ 64 MHz.
The other one is what was already in the project - i.e. kept at 72 MHz and supporting USB.

Supersedes PR #126.


rogerclarkmelbourne:
I have no way to test this, as I don't have a Nucleo, but I will pull and merge into the development branch and perhaps someone else can test it.

PR #139, opened 2015-11-15
Relax stm32flash checks

jcw:
This change to the stm32flash utility is needed to be able to open CDC-ACM type serial USB devices on at least Mac OSX (maybe also Linux).

It removes a check that the cflag value ends up precisely as expected by stm32flash. I don't think this is important in this context.

One more step will be needed to fully address this PR: it needs re-built versions of stm32flash for at Lac OSX and Linux (Windows code is not affected by this change).


jcw:
Drat - wrong branch - I'm closing this again, sigh...

PR #140, opened 2015-11-15
Relax stm32flash checks (again)

jcw:
This change to the stm32flash utility is needed to be able to open CDC-ACM type serial USB devices on at least Mac OSX (maybe also Linux).

It removes a check that the cflag value ends up precisely as expected by stm32flash. I don't think this is important in this context.

One more step will be needed to fully address this PR: it needs re-built versions of stm32flash for at Lac OSX and Linux (Windows code is not affected by this change).

Replaces incorrect PR #139.

### DIFF
--- a/tools/src/stm32flash_serial/src/serial_posix.c
+++ b/tools/src/stm32flash_serial/src/serial_posix.c
@@ -182,11 +182,18 @@ static port_err_t serial_setup(serial_t *h, const serial_baud_t baud,
 	if (tcsetattr(h->fd, TCSANOW, &h->newtio) != 0)
 		return PORT_ERR_UNKNOWN;
 
+/* this check fails on CDC-ACM devices, bits 16 and 17 of cflag differ!
+ * it has been disabled below for now -jcw, 2015-11-09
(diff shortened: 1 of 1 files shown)


rogerclarkmelbourne:
Hi JeanClaude

I recall rebuilding for Linux at some time in the past, so I should be able to do that for you.

I think I must have also rebuilt for Windows as there was a bug with COM port numbers greater than single digit (e.g. COM10 didnt work)

So I must have recompiled for WIndows. However I will need to work out if it needs MinGW or CygWin to compile it.

rogerclarkmelbourne:
I've actioned the pull request but I will need to rebuild and test and upload the binaries, for Windows and Linux

jcw:
Thanks. One step at a time. I also should send you (or PR) the Mac OSX binary, I suppose?

rogerclarkmelbourne:
Yes. 

Saves me needing to fire up my Mac and compiling it etc ;-)

jcw:
Here's the build for Mac OSX 10.11.1.

PR #152, opened 2015-12-10
put back original bkp.* files (not symlink)

martinayotte:
I did a mess yesterday : I though doing a remove of old bkp.c.xxx and bkp.h.xxx, and adding bkp.c and bkp.h, but instead, those were symbolic links and GIT were threating then as is, and I didn't notice that.
(But that good to know if one day we need really symbolic links for common stuff)


rogerclarkmelbourne:
Ah.

I did wonder why you'd effectively deleted those files ;-)

PR #157, opened 2016-01-10
Update dfu-utils for OS X 10.11.2

derkostka:
Update binaries for 10.11.2 of avoid the following error:
„dyld: Library not loaded: /usr/local/lib/libusb-1.0.0.dylib“


rogerclarkmelbourne:
Thanks

I will test on my Mac.

rogerclarkmelbourne:
Thanks

Issue #158, opened 2016-02-04
LiquidCrystal_I2C library has the wrong bit order


ddrown:
Forum thread: http://stm32duino.com/viewtopic.php?f=14&t=882

Suggested resolution: removal of the built-in LiquidCrystal_I2C library, as the generic Arduino library works on the stm32

PR #159, opened 2016-02-04
Remove libraries/LiquidCrystal_I2C/

ddrown:
closes rogerclarkmelbourne/Arduino_STM32#158


rogerclarkmelbourne:
Thanks

That saves me a load of effort

PR #163, opened 2016-02-13
Make sure both USB_ISTR_SUSP & USB_ISTR_WKUP are both cleared when US…

geekylou:
…B susppends or resumes.
This is required as when we see a suspend event it looks like we immediately afterwords trigger a resume from the bit being set previously.

### DIFF
--- a/STM32F1/cores/maple/libmaple/usb/stm32f1/usb.c
+++ b/STM32F1/cores/maple/libmaple/usb/stm32f1/usb.c
@@ -210,7 +210,7 @@ void __irq_usb_lp_can_rx0(void) {
 
 #if (USB_ISR_MSK & USB_ISTR_WKUP)
     if (istr & USB_ISTR_WKUP & USBLIB->irq_mask) {
-        USB_BASE->ISTR = ~USB_ISTR_WKUP;
+        USB_BASE->ISTR = ~(USB_ISTR_WKUP | USB_ISTR_SUSP);
         usb_resume(RESUME_EXTERNAL);
     }
 #endif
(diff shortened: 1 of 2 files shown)


rogerclarkmelbourne:
Thanks

How do I test this PR.

Do you have an example of what breaks, so I can test before and after the PR is applied

geekylou:
I created a test here:
https://github.com/geekylou/ArduinoSTM32Tests/blob/master/Blink_powersave_test/Blink_powersave_test.ino

rogerclarkmelbourne:
Thanks

I'll take a look

BTW. I know Steve mentioned that we have another related issue, which I may be able to fix using your PR 

If you use Serial.print and the board is not connected to USB, somewhere in the underlying USB handling there is a timeout when attempting to send the data back to the PC.

So the Serial.print ends up blocking until the USB subsystem times out.

I'm not precisely sure where I'd need to add it, but perhaps just change USBSerial  print to just return if USB is not connected, as I suspect it would fix the blocking  / timeout issue

geekylou:
It looks like the USB serial code already checks if the device is connected but this isn't enough as we should also check if DTR is active as well (DTR is set when a client is actually connected on the other end).
After adding this I'm not seeing my maple mini freeze any more while calling Serial.Print without a serial terminal or if there is no USB device connected.

rogerclarkmelbourne:
Hi

I will try to look at this soon.

I was delaying as I had a load of changes in the "development" branch which I wanted to merge into the master. But I did that 2 days ago, and so far no one has any problems.

And it looks like your PR is compatible still

So I'll see if I can find time to test and merge later today

PR #168, opened 2016-03-07
max flash size is incorrect for the discovery F407

ddrown:
ref: http://stm32duino.com/viewtopic.php?f=39&t=537&start=30


rogerclarkmelbourne:
Thanks

PR #170, opened 2016-04-07
Fix build

rngkll:
I tested by building the blink base example.

### DIFF
--- a/STM32F4/cores/maple/HardwareTimer.h
+++ b/STM32F4/cores/maple/HardwareTimer.h
@@ -33,7 +33,7 @@
 
 // TODO [0.1.0] Remove deprecated pieces, pick a better API
 
-#include "timer.h"
+#include "libmaple/timer.h"
 
 /** Timer mode. */
 typedef timer_mode TimerMode;
--- a/STM32F4/cores/maple/boards.h
+++ b/STM32F4/cores/maple/boards.h
@@ -39,9 +39,9 @@
 #ifndef _BOARDS_H_
 #define _BOARDS_H_
 
-#include "libmaple.h"
-#include "gpio.h"
-#include "timer.h"
+#include "libmaple/libmaple.h"
+#include "libmaple/gpio.h"
+#include "libmaple/timer.h"
 
 #include "wirish_types.h"
 
--- a/STM32F4/cores/maple/ext_interrupts.h
+++ b/STM32F4/cores/maple/ext_interrupts.h
@@ -24,8 +24,8 @@
  * SOFTWARE.
  *****************************************************************************/
 
-#include "libmaple_types.h"
-#include "nvic.h"
+#include "libmaple/libmaple_types.h"
+#include "libmaple/nvic.h"
 
 /**
  *  @file ext_interrupts.h
--- a/STM32F4/cores/maple/io.h
+++ b/STM32F4/cores/maple/io.h
@@ -33,8 +33,8 @@
 #ifndef _IO_H_
 #define _IO_H_
 
-#include "gpio.h"
-#include "adc.h"
+#include "libmaple/gpio.h"
+#include "libmaple/adc.h"
 
 #include "wirish_time.h"
 
--- a/STM32F4/cores/maple/wirish.h
+++ b/STM32F4/cores/maple/wirish.h
@@ -34,7 +34,7 @@
 #define _WIRISH_H_
 
 #include <stdlib.h>
-#include "libmaple.h"
+#include "libmaple/libmaple.h"
 
 #include "wirish_types.h"
 #include "boards.h"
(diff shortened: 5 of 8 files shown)


martinayotte:
Why do you need to add "libmaple" path to all those include when there is already the path passed to compiler using "-I{build.core.path}/libmaple" ?

rngkll:
Thanks, not sure why at build time, is not using the includes in platform.txt
the only 2 includes are the following.
"-I/Users/asegura/Dropbox/Arduino/hardware/Arduino_STM32/STM32F4/cores/maple" "-I/Users/asegura/Dropbox/Arduino/hardware/Arduino_STM32/STM32F4/variants/discovery_f407"

If you can point me out to documentation on how to fix the issue the correct way, I'll do it.

Thanks.

martinayotte:
About the change to plateform.txt, ie : about "{build.path}/{archive_file}", why this is needed ?
Are you using Arduino-1.6.5 ? If not, what is your build environment, because all the above changes are not required with 1.6.5 ...

martinayotte:
For the include paths, can you turn on Verbose flag and copy the whole call to arm-none-eabi-g++ ?
All the includes should be there ... If not, we need to chase this issue instead of patching the sources with plasters.

ddrown:
If you're using a newer build than Arduino 1.6.5, I have fixes for it in my branch.  You can install stm32f4 support, and I've verified that the blink sketch compiles and works.

More info at
http://stm32duino.com/viewtopic.php?f=41&t=844

martinayotte:
Hi Ddrown,

Digging the thread you've mentioned is not easily provide answers...
How does the 1.6.7+ handle the expansion of the "-I{build.core.path}/libmaple" path ?
Is it that "{build.core.path}" macro doesn't exist anymore ? how it is now named ?
(also, what about "{build.path}" ?

rogerclarkmelbourne:
Guys I'm not sure if its possible for the platform.txt to be compatible with 1.6.5 (and earlier) and also compatible with 1.6.6 (and newer)

Hence why its stated in several places that the repo only works with 1.6.5

I know @ddrown  has done a version which works with 1.6.6, which is great, but I thought that if those changes were made to the repo, that everyone would need to update their copy of Arduino to 1.6.7 etc

This will cause problems for people like @mrbrunette etc who use the Arduino.llc IDE and possibly other people using make files


I'm not sure if there is a fix that would make platform.txt work with all versions of the IDE :-(

ddrown:
Ah, the thread is just about installing from the Board Manager.  The fixes for the STM32F4 for a recent Arduino IDE are here (also included are the changes needed to split the tools into something the board manager can use): https://github.com/ddrown/Arduino_STM32/commit/a268ede31f399ac51694191cd3eee81cf1c86a9a#diff-4d3659c6f4424444945a0580c761a94a

The recipe.ar.pattern using build.path has been marked as "legacy" in newer Arduino IDEs

For the includes, I added a new flag:
compiler.cpreprocessor.flags="-I{build.core.path}/libmaple"

martinayotte:
The only way we can support both would be to provide the change but leave them commented, so that endusers with newer IDE would simply have to comment old ones and uncomment new ones.

phenom128:
is there any way that the ide version can be made visible and thus make a
conditional test possible?

martinayotte:
Unfortunately, the only version shown is in the About dialog. 
I don't think the Arduino IDE expose any version number to plateform.txt, since this file is not code execute, but settings parsed by IDE itself, the execution of the commands is then done by the IDE.
That could be a good suggestion to Arduino IDE team, but I won't expect them to listen to the proposal.
So, this has nothing to do with the STM32 forum.

berryelectronics:
how about two branches? One for 1.6.6+ and one for pre-1.6.6 ?

~Straw

martinayotte <notifications@github.com> schrieb am Fr., 8. Apr. 2016 um
19:16 Uhr:

martinayotte:
Maybe too much of hurden for Roger ...
Better having 2 files : plateform.txt and plateform.txt-166, EndUser can simply overwrite the first with the other.

berryelectronics:
Then we should note that somewhere (in the readme maybe?)

martinayotte:
For the includes, I figured out a way to be compatible with both 1.6.5 and 1.6.8 by adding both keywords : 
`compiler.cpreprocessor.flags={compiler.libs.c.flags}`
So, new keyword equal old keyword.
(I'm still chasing some other issue with 1.6.8 within a VM, for example, all {object_file} keywords are translated to "/dev/null", no Objs output ... :cry: )

rngkll:
Hi, Iḿ using version 1.6.8 of the IDE.

I'm testing with ddrown repository using master branch and I can build the blink example.

martinayotte:
In my case, a Blink sketch is also working fine with official STM32 branch.
But I have problem with my big STM32F405Stamp sketch where all {object_file} keywords are translated to "/dev/null". I'm still try to figure out what IDE 1.6.8 has in it's brain :cry: 
(Same sketch works fine under 1.6.5)

rngkll:
Hi, Iḿ using version 1.6.8 of the IDE.

I'm testing with ddrown repository using master branch and I can build the blink example.

rngkll:
Closed because was not a clean fix.

rogerclarkmelbourne:
OK.

Thats a shame.

PR #174, opened 2016-04-17
Install udev-rules for ST-link

WereCatf:
As the title says, install udev-rules for being able to use ST-link to upload sketches without having to be root.


rogerclarkmelbourne:
Thanks

PR #183, opened 2016-06-30
fix LeafLab bug in USBSerial::disableBlockingTx()

martinayotte:
Hi Roger,

@edogaldo found a bug in USBSerial::disableBlockingTx(), it was almost a typo from LeafLab guys. :-)

... and I've found a missing flag I forgot to commit since 1 year for Netduino2Plus.
(We were lucky that no one else than me is using that board, otherwise the issue would have surface earlier)

Ciao !


rogerclarkmelbourne:
Hi Martin

We have the luxury of being one of the few people using this repo, so I am happy to just to the merge straight on github ;-)

PR #184, opened 2016-07-05
F1 Usart improvements

edogaldo:
- added buffered-interrupt based TX
- buffers can be dynamically allocated via HardwareSerial.begin()
       - min RX buffer size: 16
       - default RX buffer size: SERIAL_RX_BUFFER_SIZE
       - min TX buffer size: 1
       - default TX buffer size: SERIAL_TX_BUFFER_SIZE
- added HardwareSerial.enableBlockingTx()/disableBlockingTx() (blocking enabled by default)
- improved ring_buffer to support the full buffer size instead of buffer size - 1

### DIFF
--- a/STM32F1/cores/maple/HardwareSerial.cpp
+++ b/STM32F1/cores/maple/HardwareSerial.cpp
@@ -117,11 +117,6 @@ static void disable_timer_if_necessary(timer_dev *dev, uint8 ch) {
 #warning "Unsupported STM32 series; timer conflicts are possible"
 #endif
 
-void HardwareSerial::begin(uint32 baud) 
-{
-    beginNew(baud);
-}
-
 /*
  * Roger Clark.
  * Note. The config parameter is not currently used. This is a work in progress.  
--- a/STM32F1/cores/maple/HardwareSerial.cpp
+++ b/STM32F1/cores/maple/HardwareSerial.cpp
@@ -117,25 +117,23 @@ static void disable_timer_if_necessary(timer_dev *dev, uint8 ch) {
 #warning "Unsupported STM32 series; timer conflicts are possible"
 #endif
 
-/*
 void HardwareSerial::begin(uint32 baud) 
 {
-    begin(baud, SERIAL_8N1);
+    beginNew(baud);
 }
 
-void HardwareSerial::begin(uint32 baud, uint8_t config) 
-{
-    begin(baud, config, SERIAL_RX_BUFFER_SIZE, SERIAL_TX_BUFFER_SIZE);
-}
-*/
 /*
  * Roger Clark.
  * Note. The config parameter is not currently used. This is a work in progress.  
  * Code needs to be written to set the config of the hardware serial control register in question.
  *
 */
+void HardwareSerial::begin(uint32 baud, uint8_t config) 
+{
+    beginNew(baud, SERIAL_RX_BUFFER_SIZE, SERIAL_TX_BUFFER_SIZE, config);
+}
 
-void HardwareSerial::begin(uint32 baud, uint16 tx_buf_size, uint16 rx_buf_size, uint8_t config)
+void HardwareSerial::beginNew(uint32 baud, uint16 tx_buf_size, uint16 rx_buf_size, uint8_t config)
 {
  //   ASSERT(baud <= this->usart_device->max_baud);// Roger Clark. Assert doesn't do anything useful, we may as well save the space in flash and ram etc
 
--- a/STM32F1/cores/maple/HardwareSerial.cpp
+++ b/STM32F1/cores/maple/HardwareSerial.cpp
@@ -117,10 +117,17 @@ static void disable_timer_if_necessary(timer_dev *dev, uint8 ch) {
 #warning "Unsupported STM32 series; timer conflicts are possible"
 #endif
 
+/*
 void HardwareSerial::begin(uint32 baud) 
 {
-	begin(baud,SERIAL_8N1);
+    begin(baud, SERIAL_8N1);
 }
+
+void HardwareSerial::begin(uint32 baud, uint8_t config) 
+{
+    begin(baud, config, SERIAL_RX_BUFFER_SIZE, SERIAL_TX_BUFFER_SIZE);
+}
+*/
 /*
  * Roger Clark.
  * Note. The config parameter is not currently used. This is a work in progress.  
@@ -128,7 +135,7 @@ void HardwareSerial::begin(uint32 baud)
  *
 */
 
-void HardwareSerial::begin(uint32 baud, uint8_t config) 
+void HardwareSerial::begin(uint32 baud, uint16 tx_buf_size, uint16 rx_buf_size, uint8_t config)
 {
  //   ASSERT(baud <= this->usart_device->max_baud);// Roger Clark. Assert doesn't do anything useful, we may as well save the space in flash and ram etc
 
@@ -141,7 +148,10 @@ void HardwareSerial::begin(uint32 baud, uint8_t config)
 
     disable_timer_if_necessary(txi->timer_device, txi->timer_channel);
 
-    usart_init(this->usart_device);
+    rx_buf_size = rx_buf_size < 16 ? 16 : rx_buf_size; // min RX buf size = 16
+    tx_buf_size = tx_buf_size == 0 ? 1 : tx_buf_size; // min TX buf size = 1
+    
+    usart_init_new(this->usart_device, rx_buf_size, tx_buf_size);
     usart_config_gpios_async(this->usart_device,
                              rxi->gpio_device, rxi->gpio_bit,
                              txi->gpio_device, txi->gpio_bit,
@@ -183,15 +193,23 @@ int HardwareSerial::availableForWrite(void)
  * so just return 1, meaning that 1 char can be written
  * This will be slower than a ring buffer implementation, but it should at least work !
  */
-  return 1;
+  return !(rb_is_full(this->usart_device->wb));
 }
 
 size_t HardwareSerial::write(unsigned char ch) {
 
     usart_putc(this->usart_device, ch);
-	return 1;
+    return 1;
 }
 
 void HardwareSerial::flush(void) {
     usart_reset_rx(this->usart_device);
+}
+
+void HardwareSerial::enableBlockingTx(void) {
+    usart_enabableBlockingTX(this->usart_device);
+}
+
+void HardwareSerial::disableBlockingTx(void) {
+    usart_disabableBlockingTX(this->usart_device);
 }
--- a/STM32F1/cores/maple/HardwareSerial.cpp
+++ b/STM32F1/cores/maple/HardwareSerial.cpp
@@ -125,7 +125,7 @@ static void disable_timer_if_necessary(timer_dev *dev, uint8 ch) {
 */
 void HardwareSerial::begin(uint32 baud, uint8_t config) 
 {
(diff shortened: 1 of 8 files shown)


RickKimball:
Code like this is a waste of time.
```
static ring_buffer usart1_rb = {0};
```
variables like that are going to be initialized by the c runtime at startup. The way you wrote it means it would be in the .data sections instead of the .bss section.  However, the compiler is probably going to see that it is initialized to zero and just ignore you.  If you actually need to initialized it to some none zero value it would make sense but initializing it to zero doesn't.

https://github.com/rogerclarkmelbourne/Arduino_STM32/blob/master/STM32F1/variants/generic_stm32f103c/wirish/start_c.c#L77

RickKimball:
Instead of creating a new function name

edogaldo:
Hello Rick, thank you very much for your comments, very appreciated!
Here my feedbacks.

> Code like this is a waste of time.
`static ring_buffer usart1_rb = {0};`

My concern was to make sure that pointer buf in ring_buffer structure was initialized to NULL before first invocation to rb_init_new(). 
This allows me to distinguish between the first invocation to rb_init_new() and subsequent ones on the same buffer so that:
- in the first invocation (buf == NULL) I just allocate the new buffer
- in subsequent invocations (buf != NULL) I can deallocate the old buffer if needed (if the buffer size changed)
Of course if you know that this is not needed because I can reliably assume that the compiler already initializes 'buf = NULL' or you can give me a better solution, I'm open to change it.

>Instead of creating a new function named beginNew() .. why not just provide default values for the existing one and changes its signature in the header file [...]

On my point of view it's likely that users are interested in setting the buffer size without having to explicitly specify a value for "config", this would not be possible just extending the begin() signature by adding the buffer size arguments after "config".
In my first commit I changed the begin signature adding the buffer size arguments between "baudrate" and "config" but Roger and others correctly observed that this could have created confusion or backward compatibility issues in users code.
This was the reason why I opted out for the decision to leave the begin interface unchanged and instead provide support for the new feature via a new api.

RickKimball:
If you declare like this:
```
static ring_buffer usart1_rb
```

Is initialize to all zeros by this code:
```
    /* Zero .bss. */
    dst = (int*)&__bss_start__;
    while (dst < (int*)&__bss_end__) {
        *dst++ = 0;
    }
```
Which is why I put a link to it in my comment above:

https://github.com/rogerclarkmelbourne/Arduino_STM32/blob/master/STM32F1/variants/generic_stm32f103c/wirish/start_c.c#L77

edogaldo:
Thank you Rick, just to double check I correctly understood: what you mean is that = {0} is useless since the static structures are already cleared by start_c(), correct?

RickKimball:
> On my point of view it's likely that users are interested in setting the buffers size without having to explicitly specify a value for "config" (I think very few people bother about "config"),

 beginNew(uint32 baud, uint16 tx_buf_size=SERIAL_TX_BUFFER_SIZE, uint16 rx_buf_size=SERIAL_RX_BUFFER_SIZE, uint8_t config=SERIAL_8N1);

Well that argument doesn't really work if I just want to change the receive buffer, I would still have to provide the size of the transmit buffer even though I only wanted to change the rx buffer size. Which by the way is probably the one people are going to want to change more often than tx buffer size.

I don't see a problem reorganizing the order of the arguments to make it easier to have one function name with defaults.  However in spite of that you could still use the name begin for any of those combos as long as there unique signature for each

begin(uint32 baud); // default
begin(uint32 baud, uint8_t config=SERIAL_8N1);
begin(uint32 baud, uint16_t rx_buf_size=SERIAL_RX_BUFFER_SIZE, uint16_t tx_buf_size=SERIAL_TX_BUFFER_SIZE. uint8_t config=SERIAL_8N1);

RickKimball:
> Thank you Rick, just to double check I correctly understood: what you mean is that = {0} is useless since the static structures are already cleared by start_c(), correct?
Yes

RickKimball:
begin(uint32 baud, uint8_t config=SERIAL_8N1); ...

config should really be an enum or at least a constant

edogaldo:
> However in spite of that you could still use the name begin for any of those combos as long as there unique signature for each
```
begin(uint32 baud); // default
begin(uint32 baud, uint8_t config=SERIAL_8N1);
begin(uint32 baud, uint16_t rx_buf_size=SERIAL_RX_BUFFER_SIZE, uint16_t tx_buf_size=SERIAL_TX_BUFFER_SIZE. uint8_t config=SERIAL_8N1);
```

Uhm..
How can the compiler distinguish the correct method in this call?
`Serial.begin(9600,64);`

RickKimball:
Yes  true ..

One function then and then 

void begin(uint32_t baud, uint8_t config=0b110000, uint32_t rx_buffer_size=64, uint32_t tx_buffer_size=64)

edogaldo:
So much ugly having also beginNew()?! It would guarantee the user will to set the buffers size..

RickKimball:
so does the one above

stevstrong:
I would also stick to the old function name, just that the parameter sequence should be additive:
```
begin(param A);
begin(param A, param B);
begin(param A, param B; buf_t rx, buf_t tx);
```

edogaldo:
Thank you @stevstrong, I'll take this into account!

@RickKimball: dear Rick, I'd need one more clarification from you about:
`static ring_buffer usart1_rb = {0};`
is it a real error or is it just useless?
For "real error" I mean it can/will create problems.
If it's a real error, can you explain me why?

Thanks in advance and best, E.

RickKimball:
In the case of stm32duion all .bss data is initialized to zero because of the code I showed in start_c.c

if you use 
```
static ring_buffer usart1_rb = {0};
```
It may use 2 times as much space depending on the compiler settings. You are telling it save a copy in the data segment. That means space is used in both flash and ram.

If you use
```
static ring_buffer usart1_rb;
```
Only space in the .bss segment (ram) is used and it is initialized to zero

http://stackoverflow.com/questions/610682/do-bss-section-zero-initialized-variables-occupy-space-in-elf-file/610816#610816

RickKimball:
An example:
````
struct foo_64 {
  volatile unsigned head;
  volatile unsigned tail;
  char buff[64];
};

struct foo_512 {
  volatile unsigned head;
  volatile unsigned tail;
  char buff[512];
};

static foo_512 foo_bss;
static foo_64 foo_data= { 0 };
```
Lets look at the symbols and see where they end up:
$ arm-none-eabi-nm -CS sketch_jul08a.ino.elf  | grep foo
20000594 00000208 b foo_bss
08003498 00000048 r foo_data

foo_bss is only in ram:
The 64 bytes + 4 +4 of foo_bss end up in ram ... at address 0x20000594

foo_data is in .rodata and ram
The 512 bytes + 4 + 4 of foo_data end up in .rodata and in ram the flash is just a bunch of zeros

$ arm-none-eabi-readelf -x .rodata sketch_jul08a.ino.elf | grep -e 80034
  0x08003470 00000000 00000000 df100008 fb100008 ................
  0x08003480 01130008 f7100008 35110008 17110008 ........5.......
  0x08003490 c3120008 00000000 **00000000 00000000 ................
  0x080034a0 00000000 00000000 00000000 00000000 ................
  0x080034b0 00000000 00000000 00000000 00000000 ................
  0x080034c0 00000000 00000000 00000000 00000000 ................
  0x080034d0 00000000 00000000 00000000 00000000 ................**
  0x080034e0 48656c6c 6f20576f 726c640d 0a00636f Hello World...co
  0x080034f0 756e7465 72203d20 0048656c 6c6f2041 unter = .Hello A
$

RickKimball:
Shouldn't HardwareSerial::end() wait for the tx_buffer to empty before disabling the usart? It looks like all it does now is wait for the current character to finish and then it clears the buffers.

RickKimball:
Why are you renaming a well established name in the ring buffer structure from tail to used?  This just confuses people if they know anything about ringbuffers.

https://github.com/edogaldo/Arduino_STM32/blob/d5f9f7bf48fa303557e6529ab731b25bc465f8f3/STM32F1/system/libmaple/include/libmaple/ring_buffer.h#L55

RickKimball:
Do you have a test sketch you used to verify all this stuff works?

edogaldo:
> Shouldn't HardwareSerial::end() wait for the tx_buffer to empty before disabling the usart? It looks like all it does now is wait for the current character to finish and then it clears the buffers. 

I guess you are right (at least for the blocking scenario) I'll fix it.

> Why are you renaming a well established name in the ring buffer structure from tail to used? This just confuses people if they know anything about ringbuffers.

Well if you look deeper I did something more than just changing the attribute from "tail" to "used": I changed the approach so that I can use the full buffer size and not ([buffer size] - 1) like in the current implementation; I anyway paid attention to make it transparent for ring_buffer clients.

edogaldo:
> Do you have a test sketch you used to verify all this stuff works?

Here the sketch I'm using to "stress test" it:
`//#define Serial Serial1  // TX A2

#include <libmaple/scb.h>
byte nodeID = 10;
byte packetNum = 11;
float C = 42.611;
float F = 42.611;
float H = 42.6;
float vBatt = 3.14;
int packetLoss = 10;
int packetTotal = 10000;

#define tx_buf_start 1
#define tx_buf_stop 300
uint16 tx_buf_size=tx_buf_start;


void printData() {
  Serial.print(nodeID);
  Serial.print(",");
  Serial.print(packetNum);
  Serial.print(",");
  Serial.print(C);
  Serial.print(",");
  Serial.print(F);
  Serial.print(",");
  Serial.print(H);
  Serial.print(",");
  Serial.print(vBatt,3);
  Serial.print(",");
  Serial.print(packetLoss);
  Serial.print(",");
  Serial.println(packetTotal);
  //Serial.println(SCB_BASE->VTOR,HEX);
  //Serial.println("abcdefghijklmnopqrstuvwxyz0123456789abcdefghijklmnopqrstuvwxyz0123456789abcdefghijklmnopqrstuvwxyz0123456789abcdefghijklmnopqrstuvwxyz0123456789abcdefghijklmnopqrstuvwxyz0123456789abcdefghijklmnopqrstuvwxyz0123456789");
}

void setup() {
  //Serial.begin(57600);
}

void loop() {
  //Serial.beginNew(9600,tx_buf_size);
  //Serial.beginNew(57600,tx_buf_size);
  Serial.beginNew(115200,tx_buf_size);
  //Serial.disableBlockingTx();
  unsigned long start = micros();
  printData();
  unsigned long end = micros();
  Serial.print("tx_buf_size[");Serial.print(tx_buf_size);Serial.print("]: ");Serial.print(end - start);Serial.println(" us");
  Serial.println();
  tx_buf_size *= 2;
  if(tx_buf_size > tx_buf_stop){
    tx_buf_size = tx_buf_start;
    //Serial.println("=================");
    Serial.println("@@@@@@@@@@@@@@@@@@@");
    Serial.println();
  }
  delay(1000);
}` 

I also tested it with some default Arduino sketches performing comparations with the current implementation and it works as long as the original version works as well (it doesn't work where the original version doesn't work)

edogaldo:
> I think it is wrong to change the meaning of a ring buffer that is pretty much standard everywhere else. I really don't have a problem with 1 byte being wasted. Most people expect that.

I'm not changing the meaning or the interface, I'm just changing the implementation
There are function to handle the structure, it's to be expected that users access the structure using the functions..

RickKimball:
to make a chunk code use ```

```
This is code
```

edogaldo:
I won, thanks! :)

edogaldo:
Too much complex and did not get success..

PR #186, opened 2016-07-08
Changed serial speed on serial_upload, added gnu11 compiler param

bigjohnson:
I've changed the speed to 57600 because with STMF103C8 the upload don't work.
But the old speed was for new micro, we need to parametrize it?
Added -std=gnu++11 -std=gnu11 parameters in platform.txt

### DIFF
--- a/STM32F1/platform.txt
+++ b/STM32F1/platform.txt
@@ -16,13 +16,13 @@ compiler.warning_flags.all=-Wall -Wextra -DDEBUG_LEVEL=DEBUG_ALL
 # ----------------------
 compiler.path={runtime.tools.arm-none-eabi-gcc.path}/bin/
 compiler.c.cmd=arm-none-eabi-gcc
-compiler.c.flags=-c -g -Os {compiler.warning_flags} -MMD -ffunction-sections -fdata-sections -nostdlib --param max-inline-insns-single=500 -DBOARD_{build.variant} -D{build.vect} -DERROR_LED_PORT={build.error_led_port} -DERROR_LED_PIN={build.error_led_pin} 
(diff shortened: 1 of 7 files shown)


rogerclarkmelbourne:
I don't understand this changes.

1. Serial upload works for many people with the existing settings. 
How do you know that it will continue to work at this lower speed for everyone ??
Also, even if that speed works for other people it will be considerably slower and I'm likely to get complaints about downgrading this feature

2. I have no idea what differences the change to the gcc std flag will have on the compiled output
I will need to find out from experts on the forum, what impact this has

3. Why did you change the BMP upload? which exactly is this change. What does it fix ?

bigjohnson:
1) Was a my usb to serial issue... discard it.
3) No the code is the same, I don't touched it.
2) In the Arduino's 1.6.x they added these flag for new features compatibility.
They are copied from Arduino DUE compile flag.
I insert them to compile LMIC Arduino's official library and tested.
Thanks

Alberto

rogerclarkmelbourne:
Albert

OK.

I'll discuss the compiler flags with the guys on the forum who have much more knowledge about this than I do

RickKimball:
@bigjohnson what usb dongle are you using?

bigjohnson:
CP2102 on Osx 10.9.5

rogerclarkmelbourne:
The jury is out on whether the change to cflags is needed See
http://www.stm32duino.com/viewtopic.php?f=3&t=1241&hilit=std%3Dgnu11

The Serial upload speed seems to be a driver fault on your machine

And you didnt mean to change the BMP upload, but the github diff shows that there was a change

So I will close this PR as its not required / superseded

Issue #102, opened 2015-08-26
Wire library has SCL and SDA definitions but then doesn't use them. 


rewolff:
I looked in "Wire.h" to find what pins the I2C bus would show up on. Found definitions for SCL and SDA but... nothing! Turns out there is a 

```
//TwoWire Wire(SCL, SDA, SOFT_STANDARD);
TwoWire Wire(PB6, PB7, SOFT_STANDARD);
```

at the bottom of Wire.cpp which ignores the definitions of SCL and SDA.

rogerclarkmelbourne:
Sorry

Can you explain in more detail.

rewolff:
It's been a while since I looked at your project. IIRC,.... 

At I Needed to change the SDA and SCL pins of the Wire (I2C) module. So I changed the SDA and SCL definitions near the top of the file. Then near the bottom the "Wire" is instantiated but then doesn't use the SDA and SCL definitions.

rewolff:
Here is the offending code: 
https://github.com/rogerclarkmelbourne/Arduino_STM32/blob/master/STM32F1/libraries/Wire/Wire.cpp#L192
(Look at wire.h to deterimine which pins are going to be used and then.... no reaction. Turns out the "Wire" object is instantiated with PB6 and PB7.

rogerclarkmelbourne:
OK

What really needs to happen is that in Wire.h those definitions are changed to PB6 and PB7 as the pin numbers only apply to Maple mini boards, and they are not the only board that is now supported.

Then yes, I could change it back at the bottom of Wire.cpp

Also as it looks like the constructor defaults it to the same settings as in the bottom of wire.cpp, I suspect it can be written as 

TwoWire Wire()

BTW. Normally if someone wants Wire on nonstandard pins they declare their own instance of 

TwoWire myWire(MY_SCL_PIN, MY_SDA_PIN, MY_SPEED);

rewolff:
Well the inconsistency caught me out. I know nothing of "customary" pins, so I went to the source, and found a neat definition of where the signals were going to be. Then I took my SCL and SDA wires and plugged them onto my board at those positions. Nothing!

[...]

PR #197, opened 2016-08-07
Wire uses wrong SDA/SCL definitions

SamuelBrucksch:
Related to issue #102


rogerclarkmelbourne:
Thanks

It Looks like a sensible change, but I need to setup a test to confirm before I can merge it.

rogerclarkmelbourne:
Thanks.

I tested the change and it worked OK for me, so I have merged straight into master

Cheers

Roger

PR #204, opened 2016-09-06
update STM32F4 environment for macosx

kuwatay:
Add definitions of tools for macosx (platform.txt)
Update stlink binary to support STM32F407 on macosx


rogerclarkmelbourne:
I think its looks OK. 

But, I don't use the F4 stuff that much, so I'll get some of the other users (@martinayottte and @sheepdoll) to test before I action the pull request

rogerclarkmelbourne:
I don't think any of the F4 users have time to test this, so I may as well just merge it, and see if anyone finds any problems

PR #205, opened 2016-09-07
Update for recipe.ar.pattern and library 'Category not valid' IDE warnings

ozayturay:
Update for recipe.ar.pattern and library 'Category not valid' IDE warnings


rogerclarkmelbourne:
The problem with changing the recipe in platform.txt is that everyone has to upgrade to IDE 1.6.9 (or newer)

Community members were reluctant to do this.

rogerclarkmelbourne:
I think there has been plenty of time for people to upgrade to at least 1.6.9 so I'm going to action this PR now

PR #211, opened 2016-09-23
Master

bubulindo:
This modification adds support for a STM32ADC library. There are a
couple of examples on how to use it. The library allows usage of the
ADC with DMA in circular mode or scan mode, it allows the ADC to be
used in scan mode. And also to use the EOC interrupt.
More functionality needs to be added and will be in time.

Eventually the util files will be moved into the adc files in the core.

### DIFF
--- a/STM32F1/cores/maple/HardwareTimer.cpp
+++ b/STM32F1/cores/maple/HardwareTimer.cpp
@@ -142,6 +142,28 @@ void HardwareTimer::refresh(void) {
     timer_generate_update(this->dev);
 }
 
+/*  CARLOS Changes to add encoder mode.*/
+
+//direction of movement. (to be better described). 
+    uint8 HardwareTimer::getDirection(){
+        return get_direction(this->dev);
+    }
+    
+//set if the encoder will count edges on one, which or both channels. 
+    void HardwareTimer::setEdgeCounting(uint32 counting) {
+        (dev->regs).gen->SMCR = counting;//TIMER_SMCR_SMS_ENCODER3; //choose encoder 3, counting on 
+    
+    }
+    
+    uint8 HardwareTimer::getEdgeCounting() {
+        return (dev->regs).gen->SMCR;
+    }
+    
+    
+
+//set the polarity of counting... not sure how interesting this is.. 
+    void HardwareTimer::setPolarity(){} 
+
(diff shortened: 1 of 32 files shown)


RickKimball:
How does this change the memory footprint from the original version? From glancing at the code, it appears you have changed the adc_dev structure from a constant compile time structure to one that is dynamic and ends up in ram now.

bubulindo:
Hello,

The analogInput example with the current core gives me the following:

14.108 bytes of program
2800 bytes of dynamic memory.

The analog Input example with my modifications gives me the following:

14.204 bytes of program
2816 bytes of dynamic memory.

I was actually expecting a bit more... but on the other hand, the const
keyword modification is only required to be able to deal with interrupts on
the ADC as per the libmaple standard way.
Considering the possibility of running DMA to keep the ADC running, the
interrupt becomes interesting for the analog watchdog as it's not something
that could be automated without code.

I believe that the STM32ADC pretty much has everything needed to run in
DualMode, although I need to play with it some more.

rogerclarkmelbourne:
So it only appears to have added 16 bytes to RAM usage

BTW. DMA from ADC already works, @ahull on the stm32duino uses it for his Pig-O-Scope and I know other people have taken his code and used it for other purposes

I presume this change is to enable a callback on DMA completion ?

bubulindo:
No, the change enables a call back on ADC completion. DMA completion is
separate interrupt.

bubulindo:
Hello gents,

I had some time to play with my ZET6 board today and noticed a couple of
things to correct.

Concerning the previously asked question, the interrupt that removing the
const keyword will address is the NVIC_ADC1_2 (or NVIC_ADC3) which is only
active on End of conversion, Injected End of Conversion (which I have no
clue what means) and the Analog Watchdog.

Considering the fact that if DMA is available, it will be far more
efficient than using an interrupt the only purpose would be for the Analog
Watchdog as it does not have DMA on it. Although at a cost of 16 bytes, I
think it is acceptable.

Moving forward, I'm going to test the dual ADC modes (and if it is as
straightforward as the Pig-O-Scope made it seem) and also the different
triggers for conversion as this can be interesting.

If you have any feedback or ideas about this, let me know.

rogerclarkmelbourne:
I've not had time to fully test this, but as far as I can tell, everything works OK when running the new code.

Thanks

Roger

PR #212, opened 2016-09-25
Fix return value for Print::write(const void *buffer, uint32 size)

victzh:
Print::write(const void *buffer, uint32 size) always returns 0, which is against specification. This patch fixes it.


rogerclarkmelbourne:
Thanks

PR #213, opened 2016-09-27
In TFT_Show_Font_ILI9341 example change init to begin

mcauser:


rogerclarkmelbourne:
Thanks 

I'll test the change

PR #217, opened 2016-10-15
USB core and CDC fixes

jaretburkett:
Various fixes for the USB core on the STM32F103. 

1. Fixed issue caused by commit [637174e63a277d3b3daa280adaa44c85bc2a5c07](https://github.com/rogerclarkmelbourne/Arduino_STM32/commit/637174e63a277d3b3daa280adaa44c85bc2a5c07) that caused issues with sending large amounts of data to the microcontroller. FIX: Revert back 2 lines of code. 

2. Added endian fix that was added to the bootloader core, but was never added to the Arduino Core. [Link here](https://github.com/leaflabs/maple-bootloader/commit/b7c8ce89331653c6d92fd7a52a43f444c8bcb739)

Discussion of these fixes can be found on the [stm32duino.com form](http://stm32duino.com/viewtopic.php?f=9&t=1451).

### DIFF
--- a/STM32F1/cores/maple/libmaple/usb/stm32f1/usb_cdcacm.c
+++ b/STM32F1/cores/maple/libmaple/usb/stm32f1/usb_cdcacm.c
@@ -451,7 +451,7 @@ uint32 usb_cdcacm_rx(uint8* buf, uint32 len) {
 
     /* If all bytes have been read, re-enable the RX endpoint, which
      * was set to NAK when the current batch of bytes was received. */
-    if (n_unread_bytes <= (CDC_SERIAL_BUFFER_SIZE - USB_CDCACM_RX_EPSIZE)) {
+    if (n_unread_bytes == 0) {
         usb_set_ep_rx_count(USB_CDCACM_RX_ENDP, USB_CDCACM_RX_EPSIZE);
         usb_set_ep_rx_stat(USB_CDCACM_RX_ENDP, USB_EP_STAT_RX_VALID);
     }
@@ -567,7 +567,7 @@ static void vcomDataRxCb(void) {
 
 	n_unread_bytes += ep_rx_size;
 
-    if (n_unread_bytes <= (CDC_SERIAL_BUFFER_SIZE - USB_CDCACM_RX_EPSIZE)) {
+    if (n_unread_bytes == 0) {
         usb_set_ep_rx_count(USB_CDCACM_RX_ENDP, USB_CDCACM_RX_EPSIZE);
         usb_set_ep_rx_stat(USB_CDCACM_RX_ENDP, USB_EP_STAT_RX_VALID);
     }
--- a/STM32F1/cores/maple/libmaple/usb/usb_lib/usb_core.c
+++ b/STM32F1/cores/maple/libmaple/usb/usb_lib/usb_core.c
@@ -31,9 +31,6 @@
 #define USB_StatusIn() Send0LengthData()
 #define USB_StatusOut() vSetEPRxStatus(EP_RX_VALID)
 
-#define StatusInfo0 StatusInfo.bw.bb1 /* Reverse bb0 & bb1 */
-#define StatusInfo1 StatusInfo.bw.bb0
-
 /* Private macro -------------------------------------------------------------*/
 /* Private variables ---------------------------------------------------------*/
 u16_u8 StatusInfo;
@@ -857,11 +854,11 @@ u8 Setup0_Process(void)
     pInformation->USBbmRequestType = *pBuf.b++; /* bmRequestType */
     pInformation->USBbRequest = *pBuf.b++; /* bRequest */
     pBuf.w++;  /* word not accessed because of 32 bits addressing */
-    pInformation->USBwValue = ByteSwap(*pBuf.w++); /* wValue */
+    pInformation->USBwValue = *pBuf.w++; /* wValue in Little Endian */
     pBuf.w++;  /* word not accessed because of 32 bits addressing */
-    pInformation->USBwIndex  = ByteSwap(*pBuf.w++); /* wIndex */
+    pInformation->USBwIndex  = *pBuf.w++; /* wIndex in Little Endian */
     pBuf.w++;  /* word not accessed because of 32 bits addressing */
-    pInformation->USBwLength = *pBuf.w; /* wLength */
+    pInformation->USBwLength = *pBuf.w; /* wLength in Little Endian */
   }
 
   pInformation->ControlState = SETTING_UP;
--- a/STM32F1/system/libmaple/usb/usb_lib/usb_core.h
+++ b/STM32F1/system/libmaple/usb/usb_lib/usb_core.h
@@ -101,8 +101,9 @@ typedef union
   u16 w;
   struct BW
   {
-    u8 bb1;
+    /* Little Endian */
     u8 bb0;
+	u8 bb1;
   }
   bw;
 } u16_u8;
@@ -211,6 +212,8 @@ USER_STANDARD_REQUESTS;
 #define USBwLength USBwLengths.w
 #define USBwLength0 USBwLengths.bw.bb0
 #define USBwLength1 USBwLengths.bw.bb1
+#define StatusInfo0 StatusInfo.bw.bb0
+#define StatusInfo1 StatusInfo.bw.bb1
 
 /* Exported macro ------------------------------------------------------------*/
 /* Exported functions ------------------------------------------------------- */


rogerclarkmelbourne:
Thanks

@stevstrong also has a pending PR he submitted 2 days ago, which has USB fixes and a whole load of other stuff.

https://github.com/rogerclarkmelbourne/Arduino_STM32/pull/217

Basically its impossible for me to test every PR, so I normally leave things a week to give anyone else on the forum, time to test if they want to.

Then I action the PR is no one has reported any issues with the PR

So what I think I need to do is action Steve's PR and the action your PR.

But as they both change the USB (Steve's changes other stuff as well), I think there is a good that there will be merge conflicts or that the mix of both you change and his will break the USB functionality

Anyway. I'll drag out my ILI9341 display and make sure its working, then make a local branch for Steves
then if its OK, I"ll commit it
Then I'll try doing the same with your PR

And. I also have a PR pending that plays with the vector table to add DMA callbacks for ADC, which is contentious and no one has time to test ;-(

So I guess I'll need to try to test that today as well

jaretburkett:
The only conflict I see is with usb_cdcacm.c. In his commit, he changed the same 2 lines that I did (plus one for tx transmissions, based on my recommendation in the conversation on stm32duino.com, but I later modified the code to have a == comparitor vs the <= that he is still using. It will work wither way, but the n_unread_bytes is unsigned so it technically cannot be less than 0. But other than that, his changes will not affect mine, and you can really go with either one, and it will still fix the bug. Everything else should merge together and will function fine.

rogerclarkmelbourne:
OK
I'll make a new branch and pull steves change and then pull your change, and see if my ILI9341 display and the USB still seem to work afterwards ;-)

rogerclarkmelbourne:
The change didnt seem to do any harm on my test board, so I'll merged it with master

jaretburkett:
Sweet.

rogerclarkmelbourne:
BTW. I can't merge Steve's code at the moment because of an issue with SPI (which I think I've now fixed but need feedback from Steve), but also because an automatic merge won't work on the USB stuff

So I'll wait for Steve to get back to me on both points

stevstrong:
Guys, there something strange happening with the usb cdcacm driver.
Today the whole day I was trying to solve the issue, without success.
The problem is: with this fix, the usb serial works, only when reading bytes one by one.
BUT...
 is not working when reading several bytes at once.
When using
`Serial.read(ubsBuf, 60); `
the program freezes after reading and displaying the first train of 60 bytes.
If I read 64 bytes instead of 60, it suddenly works.
I figured out, it works only when the number of read bytes is a power of 2 (2, 4, 8, 16, 32, 64).
Other values bring the software to freeze after reading the first train of data.
Means, no further bytes are detected to be available. The USB host is also freezing, probably waits for the USB slave to be able to receive more data.

Any thoughts?

jaretburkett:
I bet we have to edit the Serial.read function for a buffer. probably need some type of wait timeout.

stevstrong:
I rather think it's a race condition at copying data from pma buffer (within `vcomDataRxCb()`) and signalizing Rx endpoint validity.

rogerclarkmelbourne:
I pulled @stevstrong 's version into the master.

So let me know if either of you can still see this bug

stevstrong:
Well, the more tries I make, the stronger the impression that it is the serial driver on PC buggy.

I monitor the NACKs signalized by the RX routine and the size of unread bytes and all these seem to be OK.

It seems that if the NACK is activated at Rx side, then byte packets are lost. I think the driver does not handle correspondingly the NACK from USB slave.

rogerclarkmelbourne:
When you say USB hardware, do you mean the STM32's internal USB implementation ?
You could look in the STM32F103 errata doc to see if anything like this is listed
(I recall reading it a few years ago, but I can't find a link to it at the moment)

jaretburkett:
@stevstrong I am having a hard time duplicating this bug. Can you post post your code somewhere? Maybe a repo with both program and firmware?

jaretburkett:
Here is the issue.

`uint32 USBSerial::read(void *buf, uint32 len) {
    if (!buf) {
        return 0;
    }

    uint32 rxed = 0;
    while (rxed < len) {
        rxed += usb_cdcacm_rx((uint8*)buf + rxed, len - rxed);
    }

    return rxed;
}

/* Blocks forever until 1 byte is received */
int USBSerial::read(void) {
    uint8 b;
	/*
	    this->read(&b, 1);
    return b;
	*/
	
	if (usb_cdcacm_rx(&b, 1)==0)
	{
		return -1;
	}
	else
	{
		return b;
	}
}`

For a single byte read, you have to put a blocking while loop until serial is available before each read. When reading a buffer, the same code is called for a normal one byte read, without any waiting for the next byte. This can be solved by adding a blocking while loop before bytes in` uint32 USBSerial::read(void *buf, uint32 len) {` or in the cdc core. usb_cdcacm_rx only calls usb_cdcacm_peek and moves the offset forward regardless if a byte is read of not. So this can be fixed many places, but if fixed in the cdc core, it will cause a blocking even for single bytes, which would not fit the Arduino standard. I think the best thing to do is change 
`while (rxed < len) {
        rxed += usb_cdcacm_rx((uint8*)buf + rxed, len - rxed);
    }`
to 
`while (rxed < len) {
        while(usb_cdcacm_data_available() < 1);
        rxed += usb_cdcacm_rx((uint8*)buf + rxed, len - rxed);
    }`

in Arduino_STM32/STM32F1/cores/maple/usb_serial.cpp function uint32 USBSerial::read(void *buf, uint32 len) {. I cannot duplicate the bug right now, so I cannot test it.

stevstrong:
Check the attached INO file, I added two options, select one with `#if 0` or `#if 1`.

The first one waits for one byte available and then reads it. And repeat once. Uses tft.drawPixel().
This works, but is slow.

The second option is to wait for one or more than one byte, set as you want by `#define GET_SIZE 10`. It uses the function `tft.pushColors()`.
[STM32_ImageLoader.zip](https://github.com/rogerclarkmelbourne/Arduino_STM32/files/533467/STM32_ImageLoader.zip)


This will not display any data, or will drop a timeout after a variable interval.

stevstrong:
Jaret, I first wait till the needed number of bytes is available in the buffer:
`while ( Serial.available()>=%my_bytes% ) ;`
This is equivalent to `(n_unread_bytes>= %my_bytes%)`
 and I read only afterwards. This means, normally no blocking is needed.

stevstrong:
OK, I finally got some workable results.

### Description

There are actually two issues:
1. buffer underflow
2. buffer overflow

#### 1. Buffer Underflow
It happens when the CPU is processing the data more quickly than the USB endpoint can deliver new data (each 1 ms). It looks like some bytes are missing from the expected stream.
##### Root cause:
some variables are used both in the Rx callback function (executed as interrupt) and on the main level. By evaluating at least one of them on the main level leads to a race condition which creates the situation that the interrupt level stores  the bytes, but the main level misses them.
##### Resolution:
Rework of the USB reception part from the system based on an `rx_offset` and `n_unread_bytes` to a system based on `rx_head` and `rx_tail`. This system assures that on the main level and interrupt level only one of the volatile variable is changed, respectively.

jaretburkett:
@stevstrong I also had issues when running `while ( Serial.available()>=%my_bytes% ) ;` I believe it is because of this current fix does  not allow the usb to send more bytes until the full buffer is read. This is what is giving it the speed increase. So you have 512 buffer, read 500 bytes, and then wait for 14 bytes, you will only get 12 more from usb until the entire buffer is read out. Otherwise the endpoint will not be turned back on and your while loop will hang waiting for bytes it will not receive until all the current bytes are read out. That is why I believe a workaround for reading into a buffer is to run the blocking while loop before each byte read. Of course, this is a workaround, and a rx_head and rx_tail implementation in the usb core could solve this issue as well, depending on implementation. But then you go back to reactivating the end point over and over  without taking full advantage of the buffer, which is what was slowing things down in the first place.  I personally do not use the Serial.read(&buff, len); I prefer to implement it on my end. It is the same result in the end, as it is just a function. In my own implementations, when I block for each byte, I have no issues.

stevstrong:
@jaretburkett 
Your situation is real when you don't test the available bytes. But if you do available check before reading them, no blocking should occur, since the bytes are already in the buffer ready to be read out.

jaretburkett:
Well the Serial.available() just returns n_unread_bytes from usb_cdcacm. This number is generated for the CDC buffer, not the pma rx endpoint buffer. Until the cdc buffer is completly drained to 0, the endpoint is not activated again. So if you have a full cdc buffer, and read all but 1 byte from it, the pma buffer remains empty until you reactivate the endpoint by reading the last byte from the cdc buffer. So if you read all but 1 byte and wait for 2, you will never get 2 until you read the 1 byte sitting in the cdc buffer, because the pma buffer is inactive due to the endpoint being set to USB_EP_STAT_RX_NAK. 

That is of course with this fix. There is room for improvement. Really, I am tempted to write a new cdc core all together now that I am more familiar with it. I might just try to port over the teensy core. Based on benchmark tests, it is 3 times faster for  single byte reads and almost 5 times faster for readBytes.[http://www.pjrc.com/teensy/benchmark_usb_serial_receive.html](url). I am trying to write images at max speed, and the usb transfer speed is slower than the spi transfer speed, even without DMA. I find this bottleneck unacceptable.

stevstrong:
I think that your logic is not quite complete.
First, you are not allowed to wait for a maximum size of the buffer. In my implementation ([see above](https://github.com/rogerclarkmelbourne/Arduino_STM32/pull/217#issuecomment-254262318)) it is 128+64=196 max bytes, recommended max. 128.
Second, you cannot get the buffer full at all, you are limited to the mentioned value (196 bytes).
That means, if you wait for (let's say) 100 bytes, and the buffer has 99 bytes available, no problem: the missing byte will be stored by the next USB Rx packet reception.

jaretburkett:
@stevstrong I tested your new usb_cdcacm.c and I can confirm that not only does it solve both issues, but it actually gives a speed boost. I ran it through Paul's Teensy usb speed test script and I am consistently getting over 300kBps with single byte reads and readBytes reads. With the previous version I never saw over 250 kBps, and readbytes did not work on my modification at all. So I admit, you were right, I was wrong. I cannot fully test your image program as I do not have the pushColors function in any of my libs. Would you mind posting the function you are using in your custom lib? I really want to see this at 600kBps. 

I think this one is ready to be added to the core, as the current implementation  (my idea/my bad) solved a bug and created another. This implementation is faster, and solves both bugs. Good work. @stevstrong.

stevstrong:
@jaretburkett 
You were not always wrong :)
I finally got your point with the waiting in the Serial.read(buf, size) function. It may make sense if the given size is more than 128, it cannot be checked before if the bytes are there and is not going to be received.

I admit, the fix for overflow is not 100%, as it strongly depends on the USB host behavior on NAKs.
As I wrote, for this fix the size of the Rx bytes at which the RX is ACK-ed again should be set as low as possible. But then the "buffering" is effectively lost, no matter how large the USB buffer is.
You could try to set the USB buffer size down to 128 bytes, and see if this fixes the issue.

For underflow this has no effect as underflow always runs on max. 64 bytes buffered at a time, and the USB buffer is mostly emptied long before it gets filled again. To check the underflow behavior (without displaying anything, just check the Processing output console), comment out the line with the call to "pushColor()".

Btw, I am using the [pushColors()](https://github.com/adafruit/TFTLCD-Library/blob/master/Adafruit_TFTLCD.cpp#L695) function from the original [Adafruit lib](https://github.com/adafruit/TFTLCD-Library) for 8 bit parallel access, with small changes in attempt to make it more universal, you can check it [here](https://github.com/stevstrong/Adafruit_TFTLCD_8bit_STM32).

stevstrong:
Complement info regarding the behavior in buffer overflow situation, it is actually a kind of hysteresis:
The "nr_unread" check in `vcomDataRxCb()` sets the upper value where no more bytes are accepted, the following RX packets are NAK-ed.
The "nr_unread" check in `usb_cdcacm_rx()` sets the lower value where the bytes are accepted again by the USB slave (remove NAK).

PR #218, opened 2016-10-21
Added new method to wait for the end of actual transmission

fabriciopk:
This new metohd was tested in STM32F1xx.

### DIFF
--- a/STM32F1/cores/maple/HardwareSerial.cpp
+++ b/STM32F1/cores/maple/HardwareSerial.cpp
@@ -197,3 +197,13 @@ void HardwareSerial::flush(void) {
     usart_reset_rx(this->usart_device);
     usart_reset_tx(this->usart_device);
 }
+
+/*
+* Wait for the end of actual transmission
+*
+*/
+void HardwareSerial::waitDataToBeSent(void)
+{
+    //Not this will be cleared only after a read from SR and then write to DR (Therefore after this operation, it will still be set)
+    while(!(this->usart_device->regs->SR & USART_SR_TC));
+}
--- a/STM32F1/cores/maple/HardwareSerial.h
+++ b/STM32F1/cores/maple/HardwareSerial.h
@@ -138,6 +138,7 @@ public:
     virtual int read(void);
     int availableForWrite(void);
     virtual void flush(void);
+    virtual void waitDataToBeSent(void);
     virtual size_t write(uint8_t);
     inline size_t write(unsigned long n) { return write((uint8_t)n); }
     inline size_t write(long n) { return write((uint8_t)n); }
--- a/STM32F1/cores/maple/boards.h
+++ b/STM32F1/cores/maple/boards.h
@@ -168,7 +168,7 @@ bool boardUsesPin(uint8 pin);
  * @brief Feature test: nonzero iff the board has SerialUSB.
  */
  //Roger Clark. Change so that BOARD_HAVE_SERIALUSB is always true, so that it can be controller by -DSERIAL_USB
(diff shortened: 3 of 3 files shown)


rogerclarkmelbourne:
Is this compatible with the TX buffering system ?

stevstrong:
@fabriciopk 
What kind of issue is solved with this PR?
Why is this function needed?
How to test?

fabriciopk:
@rogerclarkmelbourne  Yes, this is only compatible with the TX, for the RX we have other functions like available and flush .You can use this function when you want send some data and you want to have sure that all data was sent before you do something else.

I use this to solve problen in a board that uses serial485 <-> UART, the way that I find to send data was put a delay after my Serial.write() or wait for the registers to have sure that all data was sent.  I test the code on STM32F103C8T6 chip in a generic board that I use for series elastic actuator(use in robots).

stevstrong:
I don't think that the removal of serial USB is of general interest.

fabriciopk:
Sorry, this last commit was not to be in the pull request

rogerclarkmelbourne:
in

STM32F1/cores/maple/boards.h 



#define BOARD_HAVE_SERIALUSB 0

Is likely to break the USB isnt it ?

Do you use Serial USB ?

stevstrong:
Roger, this only works with serial 1, 2 or 3, not with USB serial. Normally this has nothing in common with serial USB.
I also have an application using a RS485 network, set up around a year ago, have to check how exactly I detected the end of transmission.

rogerclarkmelbourne:
@stevstrong 

I know the code is for USART but I think the PR permanently disables USB Serial by defining

#define BOARD_HAVE_SERIALUSB 0

which will probably override the setting from the compiler, which comes from boards.txt via platform.txt

stevstrong:
Roger, as @fabriciopk mentioned, the latest [commit](https://github.com/rogerclarkmelbourne/Arduino_STM32/pull/218/commits/ec058eee62aa1367c27cb42fe86d0713d5ba5940) should not be part of the PR.

fabriciopk:
I revert the last commit, this was just for my application work. I think it's a good ideia disable USB for the generic boards, becouse in some case we have to use the USB pins for other purposes.

rogerclarkmelbourne:
??
You think its a good idea to disable USB for all users ???

I think you are in a minority.

stevstrong:
In "some" cases may USB be suprfluous, but in "most" of the cases it is the main communicatin channel.
But thanks anyway that you have revert that change.
Roger, I think the rest can now be commited.

rogerclarkmelbourne:
Thanks Stev

I'll test it later and hopefully merge it

rogerclarkmelbourne:
I've checked the Arduino API but I can't see any mention of this function, so I can't really add it to the core.

Can't the same thing be done in the sketch by 

```
#include <libmaple/usart.h>
...
...
while(!(USART1->regs->SR & USART_SR_TC)
```);

(Or whatever USART you are using ?)

fabriciopk:
It really can be done like this. This function does not exist in Arduino AP and it makes sense not to have different functions.

stevstrong:
I tested the function, it works correctly, as expected.

I think it is valuable because the actual implementation of "Serial.flush()", which I used before on AVR systems, here differs from the official Arduino specification:
```Waits for the transmission of outgoing serial data to complete. (Prior to Arduino 1.0, this instead removed any buffered incoming serial data.)```
I our core, flush() will reset the input and output buffers, which is not conform with the new spec. Instead, it should just use this piece of code suggested by @fabriciopk 

This inconsistency should be removed.

PR #221, opened 2016-10-22
Fix interrupt race condition in usb_cdcacm_rx() and vcomDataRxCb()

Rasmusfk:
This patches an interrupt race condition usb_cdcacm_rx() and vcomDataRxCb(), which only occurred during high-bandwidth usb-to-serial transmission.

### DIFF
--- a/STM32F1/cores/maple/libmaple/usb/stm32f1/usb_cdcacm.c
+++ b/STM32F1/cores/maple/libmaple/usb/stm32f1/usb_cdcacm.c
@@ -68,6 +68,7 @@
 
 static void vcomDataTxCb(void);
 static void vcomDataRxCb(void);
+uint32 usb_cdcacm_peek_ex_noint(uint8* buf, uint32 offset, uint32 len);
 static uint8* vcomGetSetLineCoding(uint16);
 
 static void usbInit(void);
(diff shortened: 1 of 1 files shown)


jaretburkett:
There are two PRs that will conflict. I have tested both Stevstrong's and Rasmusfk's. They both completly solve both issues in the CDC core and use two different methods to solve them. I have tested with multiple tests from images, single bytes, and readbytes into a buffer. The one by stevstrong is a little faster on my read bytes testing. Stevstrong's version averages ~316 kBps on reads and Rasmusfk's version averages ~ 278 kBps on reads. Both are improvements on the current version which is only ~242 kBps. I am running tests for 2 different image to tft programs I have, as well as the test files provided at [PJRC http://www.pjrc.com/teensy/benchmark_usb_serial_receive.html](http://www.pjrc.com/teensy/benchmark_usb_serial_receive.html). I think others should run tests on both (I am testing in a linux environment), but my tests lead me to recommend Stevstrong's version. (I am posting this on both PRs.)

Rasmusfk:
Please note that my PR solves the race condition, I'm not convinced that Stevstrong's does, though I might be wrong.

stevstrong:
This is the classical way to resolve race condition, disable the involved interrupt.
The drawback is, you disable thereby also the Tx channel. So full duplex will suffer because of this.
My version is one which i was using successfully using during several years i was working professionally in the SoC field. I don't want to defend my work, but i am confident it shall work. And according to my and Jaret's test, it does work.
Additionally, the Tx channel is also improved with my PR.
But it is up to Roger to decide which version is he happy with.
If this fix would have come before i start investigating the problem, i would may be happy with it.

Rasmusfk:
If stevstrong's version doesn't have any race condition, I think we should go with that - I've tried it, and it works for me, and the tx buf change is a nice speedup.

Unrelated to the race thing: Would it be an idea dynamically allocate the rx/tx buffers in the USBSerial::begin() method, so that programs that doesn't need usbserial can save the buffer memory?

PR #224, opened 2016-11-03
Fix usb_serial reading bug

dydwo92:
I'm using maple with rosserial. But there was a strange bug that when maple reads continuous data through usb, it goes fail mode and program stop after a few minutes.

I figured out this error happens in read function in usb_serial library. After change the reading character variable to class member, the problem get solved. It would be some internal memory problem..


stevstrong:
Do you have any sketch how to reproduce the problem and test the solution?

rogerclarkmelbourne:
Try getting the latest version, it now has @stevstrong 's fixes

PR #225, opened 2016-11-05
Fix for SREG error - with this flow sensor is working

goran-mahovlic:
http://www.stm32duino.com/viewtopic.php?f=27&t=1167&sid=aba6dee79a24c8ad667381a999c4c08c

https://github.com/sekdiy/FlowMeter

Other sensors that are using SREG will probably work to. 
When tested more you can add it to F3 and F4 boards.

### DIFF
--- a/STM32F1/cores/maple/wirish.h
+++ b/STM32F1/cores/maple/wirish.h
@@ -105,3 +105,27 @@ typedef unsigned int word;
 
 #endif
 
+class SREGemulation
+{
+public:
+   operator int () const __attribute__((always_inline)) {
+      uint32_t primask;
+      asm volatile("mrs %0, primask\n" : "=r" (primask)::);
+      if (primask) return 0;
+      return (1<<7);
+   }
+   inline SREGemulation & operator = (int val) __attribute__((always_inline)) {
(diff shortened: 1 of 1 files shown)


rogerclarkmelbourne:
Looks OK. But I need to check blink etc compiles OK

rogerclarkmelbourne:
Wont compile if you include SPI.h :-(

goran-mahovlic:
ok, I will check

tnx

victorpv:
There is another pull including this pinToInterrupt function, but does it a different way.
https://github.com/rogerclarkmelbourne/Arduino_STM32/pull/225

I think would be positive if one of you post in the forum in stm32duino  a new thread about this to collect all the input and submit a single solution.

rogerclarkmelbourne:

rogerclarkmelbourne:
Whats the status of this?

Looks like no one uses it etc so perhaps I should close this due to lack of interest ???

goran-mahovlic:
I did try now and will close it SREG error does not exist any more. 
And to fix sei() cli() error we just need to add  
#define sei() interrupts();
#define cli() noInterrupts();
at the end of wirish.h

Issue #200, opened 2016-08-13
Max RAM indication not present


Testato:
On Arduino at the end of a build, in consolle appear the used/maximum flash size and the used/maximum RAM size

On STM32duino the maximum RAM do not appear

```
Sketch uses 12,316 bytes (11%) of program storage space. Maximum is 110,592 bytes.
Global variables use 2,560 bytes of dynamic memory.
```

rogerclarkmelbourne:
Please can you suggest a solution and submit via PR

rogerclarkmelbourne:
FYI.
This appears to be a data parameter that's optional in boards.txt

upload.maximum_data_size

Introduced in 

https://github.com/arduino/Arduino/commit/7c87db3532155c051b902ac83c1feea78eb732ee#diff-4ead488680453362ffec01d74609315c

Please can you update boards.txt to add this value to all boards (including F1 and F4 repos)

Thanks

Testato:
Yes i can prepare a PR,
But i checked also on Arduino Zero and this bug is present too.
Very strange

rogerclarkmelbourne:
You could do a quick test by updating one board in boards.txt and check it works.

I'm not sure why they omitted this for the Zero, I suspect they did what we did, and used an old file which doesn't have that data in it.

It doesnt seem to be documented in the Arduino Third Party Boards specification.
https://github.com/arduino/Arduino/wiki/Arduino-IDE-1.5-3rd-party-Hardware-specification

(I'm not surprised as Arduino don't seem to spend much time on documentation :-(  )

BTW. While I remember. I think some of the sizes defined in boards.txt are wrong, I can't remember if its just the Flash size thats wrong or whether the Ram size is wrong as well

rogerclarkmelbourne:
closed due to inactivity

Testato:
i agree that this issue have inactivity, but i think it will be a very good upgrade for the core.
It is very usefull have this info during compilation.
This is an example from the Uno
```
Sketch uses 444 bytes (1%) of program storage space. Maximum is 32,256 bytes.
Global variables use 9 bytes (0%) of dynamic memory, leaving 2,039 bytes for local variables. Maximum is 2,048 bytes.
```

So may you remain it opened ?

rogerclarkmelbourne:
@Testato 

Please send a PR for this

Testato:
@rogerclarkmelbourne 
i created 3 separated PR for the 3 board.txt correction

[...]

PR #238, opened 2016-12-17
Update boards.txt

Testato:
PR for issue https://github.com/rogerclarkmelbourne/Arduino_STM32/issues/200

- replaced `upload.flash.maximum_size` and `upload.ram.maximum_size` by the new arduino standard `upload.maximum_size` and `upload.maximum_data_size`
- corrected Flash and Ram value
- Tested by builded every MCU on every Variant

### DIFF
--- a/STM32F4/boards.txt
+++ b/STM32F4/boards.txt
@@ -8,9 +8,8 @@ discovery_f407.upload.protocol=stlink
 
 #discovery_f407.upload.use_1200bps_touch=false
 discovery_f407.upload.file_type=bin
-discovery_f407.upload.ram.maximum_size=17000
-discovery_f407.upload.flash.maximum_size=1048576
 discovery_f407.upload.maximum_size=1048576
+discovery_f407.upload.maximum_data_size=196608
 
 #discovery_f407.upload.usbID=1EAF:0003
 #discovery_f407.upload.altID=1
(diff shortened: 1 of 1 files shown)


rogerclarkmelbourne:

Testato:
the link is not working,
i see you merged my first of three  PR. Is there some problem with the other two ?

Testato:
link not working

stevstrong:
Correct link: http://www.stm32duino.com/viewtopic.php?f=53&t=2290

rogerclarkmelbourne:
@stevstrong 

Thanks. I was cutting and pasting many links, I think I got out of sync

stevstrong:
I see low risk (F4 related) and big benefit.

victorpv:
Looks good to me too. Seems like Arduino changed those labels, so it's good to change them, and some sizes were wrong and look better now.

rogerclarkmelbourne:
Thanks

victorpv:
@Testato the current linker scripts do not allocate anything to the CCM memory, so the MCU can't use it.
Reporting 192KB of available RAM when the MCU in fact can only use 128KB would not be appropiate. Unless you are planning to modify the linker scripts, it would be better to report the memory in those MCU as the normal RAM capacity excluding CCM.
Could you correct and resubmit?

Testato:
@victorpv 
https://github.com/rogerclarkmelbourne/Arduino_STM32/pull/303

victorpv:
Thanks Tetato. I think it makes more sense like that until a decission is made on how to use CCM and we make it usable.
There is a thread in the forum discussing on that if you want to join.

PR #244, opened 2017-01-16
Avoiding conditional directives that break statements

RomeroMalaquias:
Hello, this change is only aesthetic, it's a suggestion to avoid the use of conditional directives that break statements i think that kind of concern improve the code readability do you agree?


rogerclarkmelbourne:
@RomeroMalaquias 

The F3 core isn't used very much, so I'm happy to include the change as its low risk.

BTW. 
I'm not aware of anyone else who is actively using this F3 core. 
I only added it, to my STM32 repo as the original author seemed to have abandoned it.

I have not done any updates to the F3 code, but the F1 gets a lot of updates and PR's and the F4 gets occasional PRs.

I think that the F3 code will eventually be superseded by STM's own Arduino core github.com/stm32duino , but they have not written the F3 version yet.

PR #246, opened 2017-01-20
Fix bug in serial communication during SPI debug

LieBtrau:
When enabling SPI-debugging in SPI.cpp, the code crashed.  This was because the SPI-debug functionality tried to write to the serial port when it was not opened yet.
This has been discovered using the Nucleo F103RB.

The change on STM32F1/variants/nucleo_f103rb/board.cpp was from another pull request and makes the USB virtual COM-port of the NucleF103RB function as Serial-port.

### DIFF
--- a/STM32F1/libraries/SPI/src/SPI.cpp
+++ b/STM32F1/libraries/SPI/src/SPI.cpp
@@ -31,7 +31,7 @@
 
 #include "SPI.h"
 
-//#define SPI_DEBUG
+#define SPI_DEBUG
 
 #include <libmaple/timer.h>
 #include <libmaple/util.h>
@@ -187,7 +187,7 @@ void SPIClass::end(void) {
 void SPIClass::setClockDivider(uint32_t clockDivider)
 {
 	#ifdef SPI_DEBUG
-	Serial.print("Clock divider set to "); Serial.println(clockDivider);
+    Serial.print("Clock divider set to "); Serial.println(clockDivider, DEC);
 	#endif
 	_currentSetting->clockDivider = clockDivider;
 	uint32 cr1 = _currentSetting->spi_d->regs->CR1 & ~(SPI_CR1_BR);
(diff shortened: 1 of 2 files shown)


rogerclarkmelbourne:
I didnt even know those serial.print statements where in the code. 

I don't think they should be there at all, as people can add them if they really need to debug the library

Could you submit a PR where they are removed completely ?

stevstrong:
I knew about those print lines, but did not want to touch them as never used.
The default state of it is disabled, so in normal case does not harm.

@LieBtrau , why did you enabled it? Do you have problems with SPI?

LieBtrau:
I was implementing the [RadioHead library](http://www.airspayce.com/mikem/arduino/RadioHead/) on the Nucleo F103RB for communicating with the XL1276-D01 – LoRa 433MHz Wireless Transceiver Module.  
SPI communication wasn't working, so I turned SPI-debugging info on, which didn't help me, because the code crashed.  That's the bug that is reported in this pull request.
In the end, I found out that I had made in mistake in wiring up the wireless module to the Nucleo.  After making the correct connections, SPI communication works great.

It's up to you what you do with this pull request.  People will only encounter problems when they're turning SPI debugging on.  As SPI communication is working now, there's no need to turn debugging on.
When lines 622, 623 and 624 in SPI.cpp are removed, the bug is fixed.

stevstrong:
I have nothing against changing those lines.

rogerclarkmelbourne:
OK. I will pull this to get the Nucleo stuff then when I get time I think I will remove the debug from the SPI library, because even though its ifdef'ed out, I think its not normal for the Arduino libraries to contain debugging code, 

I should remove all of the debug print code (I can't remember if it was me that added it. I may have done, and didnt realise I'd  not removed it)

stevstrong:
Roger, you added that as a patch from timschuerewegen, but it does not matter.

rogerclarkmelbourne:
LOL

Thanks @stevestrong

I thought it was possible that I added it. But I usually try to remove debug code once a bug has been fixed.

rogerclarkmelbourne:

victorpv:
Roger I think you should close this PR. The SPI library has gotten many changes (to add callback, async, etc) and Steve is doing a general cleanup of dead code, so this PR can't be merged unless you edit the files manually, which should not be needed anyway after Steve's clean up

rogerclarkmelbourne:
@victorpv 

Good point.

I will close this as it's about to be superseded

PR #247, opened 2017-01-22
Initial Standard Peripheral Library Support

Matheus-Garbelini:
Hello, I've tested the Standard libraries in my bluepill stm32f103 and it seems to work fine. For now, I provided an example with Free RTOS and Led blink and Timer3 example with serial printing to show no conflict with the core. It include CMSYS and can work together with libmaple, so users can finally use in a easy way manipulating commom or stm cortex M3 registers.

I also included a Getting Started PDF. A very good book from indiana.edu that explain very well how to use stm32 with a lot of peripherals using standard library. My test shows no conflit with the core, thus, if it is supported in the core, the library could provide CAN, and other SLAVE mode for peripherals like SPI and I2C in a easier way then libmaple implementation.

Interrupts work when using libmaple functions to attach interrupts to a desireble function. Is it possible to add macros do disable functions handler such as timer library does, so that the user can implement a clean interrupt handler? when I add a timer interrupt without using libmaple approach, I get duplicate declaration error because the handler is already defined.

### DIFF
--- a/STM32F1/libraries/Standard_Library/examples/FreeRTOS_Blink/FreeRTOS_Blink.ino
+++ b/STM32F1/libraries/Standard_Library/examples/FreeRTOS_Blink/FreeRTOS_Blink.ino
@@ -1,0 +1,40 @@
+//#include <wirish/wirish.h>
+//#include "libraries/FreeRTOS/MapleFreeRTOS.h"
+#include <MapleFreeRTOS821.h>
+
+#include <stm32f10x_gpio.h>
+
+static void vLEDFlashTask(void *pvParameters) {
+	for (;;) {
+		vTaskDelay(1000);
+		GPIO_WriteBit(GPIOC, GPIO_Pin_13, Bit_RESET);
+		Serial.println(bitRead(GPIOC->IDR, 13), BIN);
+
+		vTaskDelay(1000);
+		GPIO_WriteBit(GPIOC, GPIO_Pin_13, Bit_SET);
+		Serial.println(bitRead(GPIOC->IDR, 13), BIN);
+	}
+}
+
+void setup() {
+	// initialize the digital pin as an output:
+	Serial.begin(115200);
+
+	GPIO_InitTypeDef GPIOInitStructure;
+	GPIOInitStructure.GPIO_Mode = GPIO_Mode_Out_PP;
+	GPIOInitStructure.GPIO_Pin = GPIO_Pin_13;
+	GPIOInitStructure.GPIO_Speed = GPIO_Speed_50MHz;
+	GPIO_Init(GPIOC, &GPIOInitStructure);
+
+	xTaskCreate(vLEDFlashTask,
+		"Task1",
+		configMINIMAL_STACK_SIZE,
+		NULL,
+		tskIDLE_PRIORITY + 2,
+		NULL);
+	vTaskStartScheduler();
+}
+
+void loop() {
+	// Insert background code here
+}
--- a/STM32F1/libraries/Standard_Library/examples/Timer3_TimerBase/Timer3_TimerBase.ino
+++ b/STM32F1/libraries/Standard_Library/examples/Timer3_TimerBase/Timer3_TimerBase.ino
@@ -1,0 +1,57 @@
+//#include <wirish/wirish.h>
+//#include "libraries/FreeRTOS/MapleFreeRTOS.h"
+#include <MapleFreeRTOS821.h>
+
+#include <stm32f10x_gpio.h>
+#include <misc.h>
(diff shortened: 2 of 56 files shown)


rogerclarkmelbourne:
Thanks.

I don't use RTOS, so I won't be able to test it, but perhaps someone else who watches this repo uses RTOS and can test it.

I will probably accept the PR, as it purely adds things, so I don't think it will affect existing users, but I'll wait in can anyone else wants to comment

Matheus-Garbelini:
Thank you, I'm using the already included RTOS from this branch, so the first example should work out of the box. Anyway, I will include some more examples next week.

rogerclarkmelbourne:
OK

If I get time I may see if it compiles, but I have pressing deadlines for work, which mean I have very little free time (as I work freelance and not 9 - 5)

Matheus-Garbelini:
OK. Thank you. Should i post this in the forum? If yes, in wich topic category? (I'm new to the community)

rogerclarkmelbourne:
Just post to General Discussion if you can't find a suitable section.

General discussion is also likely to get more views

Matheus-Garbelini:
@rogerclarkmelbourne thank you. I'm discussing with other members. Right now i'm also managed to compile HAL library with libmaple, but did not tested in the device yet. As soon as i tested and run some tests and figurout the simplest way to work with HAL interrupts, i will add in this repository.
The final idea is that the STM32f10x arduino users could configure entireli the pins and devices using STM32CubeMX to facilitate more complex projects.

rogerclarkmelbourne:
@Matheus-Garbelini 

Are you aware that we already have a HAL based core for the F103, which has been written my STM

see github.com/stm32duino

Matheus-Garbelini:
I didn't know. This is awesome. I saw it is under system/drivers in that Repo. 
But i didnt find any example or documentation in how to using it. 
There's nees to include something in the sketch or can we just call a HAL funcion?

LMESTM:
@Matheus-Garbelini  
Here's a link to HAL port install https://github.com/stm32duino/wiki/wiki
About using HAL from sketch I don't think there would be an issue, even if most of the services should be mapped to standard Arduino APIs. 
@fpistm @wi6labs

RickKimball:
This seems like it is a standalone library that hasn't had any real world use. It seems premature to just add it to the list of libraries. I think it would be better if this was created as a standalone library at least until it has some actual flight time. When you compile it, it has an amazing number of warning that haven't really been addressed. If you actually include the stm32f10x_adc.h in a script it will fail to compile because it has a name collision with wirish.h.

If you deal with this as it is,  a library, it could become something that is updated like all libraris, with the library tool.

rogerclarkmelbourne:
Thanks Rick.

I looked into the possibility of using the HAL with LibMaple a while ago and it was a can of worms.

Matheus-Garbelini:
@RickKimball I agree with you, a standalone project is the way to do.

Matheus-Garbelini:
ah, @rogerclarkmelbourne , could move my topic under general to libraries?
[http://www.stm32duino.com/viewtopic.php?f=3&t=1737&p=22944#p22944](url)

beginner3456789:
This adc sketch works with the SPL install after fixing these compiler errors:

1. Change the ADCx in the pinmap table at variants/*/board.cpp to 0xFF or NULL. Either works.
2. Comment out the define ADCx = 0xFF in the boards/maple/wirish_types.h file.

I couldn't find anyplace else the ADCx is used in the STM32F1 except the board.cpp file.
There could be other reasons for the ADCx define yet unknown.

The sketch does works in a blue pill type using this SPL Library after the above changes.
I copied and pasted this code from numerous examples available yet is certainly not optimal.

[Bluepill-ADC-SPL.ino.txt](https://github.com/rogerclarkmelbourne/Arduino_STM32/files/763883/Bluepill-ADC-SPL.ino.txt)

rogerclarkmelbourne:

victorpv:
@beginner3456789 changing ADCx to NULL would not be a good idea, since NULL is normally taken as 0 by the compiler, and 0 would indicate ADC Channel 0, which is not what ADCx is meant to represent (it means "No ADC Channel". Perhaps changing it to 0xFF, or better change the ADCx name to something else more meanfull or what it represents, such as "NoADC" or "NullADC" or something like that.

rogerclarkmelbourne:
This PR seems to cover multiple areas

The Standard Peripheral Library stuff is problematic to add because of potential licensing issues
Also it has been superseded by STM's HAL which is used in their own official core.

And. I think its conflicting and confusing for a single core to contain multiple implementations of the low level functions.

Hence I will not be able to merge this PR

PR #249, opened 2017-02-08
Add EEPROM.update() function

### LINKED ISSUE #248: [Feature request]Add update() function to EEPROM
Would it be possible to add an update() function to the EEPROM library to improve compatibility with the original Arduino lib? 

Purpose of the function is that it will pass an address and data to the standard EEPROM.write() function, but only if the data in the requested EEPROM address is different from what is in the call to update().
Intent on the arduino is to reduce unneeded writes to the EEPROM. 

See: https://www.arduino.cc/en/Reference/EEPROMUpdate

I'm happy to put together a PR for this, but thought I'd raise it here first to see whether it will be considered.

Thanks.

noisymime:
Basic EEPROM.update() function. Equivalent to original Arduino function, it checks the existing value of the address and only performs a write if the value differs from the one requested. 

Closes #248

### DIFF
--- a/STM32F1/libraries/EEPROM/EEPROM.cpp
+++ b/STM32F1/libraries/EEPROM/EEPROM.cpp
@@ -522,6 +522,28 @@ uint16 EEPROMClass::write(uint16 Address, uint16 Data)
 }
 
 /**
+  * @brief  Writes/upadtes variable data in EEPROM.
+            The value is written only if differs from the one already saved at the same address.
+  * @param  VirtAddress: Variable virtual address
+  * @param  Data: 16 bit data to be written
+  * @retval Success or error status:
+  *			- EEPROM_SAME_VALUE: If new Data matches existing EEPROM Data
+  *			- FLASH_COMPLETE: on success
+  *			- EEPROM_BAD_ADDRESS: if address = 0xFFFF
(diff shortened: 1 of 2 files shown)


rogerclarkmelbourne:
Thanks. Looks good, but I will need to test it before merging

noisymime:
Cheers. No rush on my side. 

I'll probably have a few more small compatibility type things dribble in as I work through porting a project over

rogerclarkmelbourne:
OK.

Thanks

noisymime:
Sorry to hassle, but any feedback on this one?

lacklustrlabs:
Just a thought.. is there any reason why Arduino legacy `write()` doesn't do what `update()` does in the first place? i.e. are there any valid use-cases for nuking the flash memory with unnecessary writes?
(`write()/update()` would then delegate to a private/protected `_real_write()` method.)

And here is the real question: if `write()` and `update()` at one time in the future will do the same thing, maybe we could make the transition easier by giving the methods identical return codes? 
(skip `EEPROM_SAME_VALUE`, just return `EEPROM_OK`)?

rogerclarkmelbourne:
Seems to be OK. So I've merged.

Thanks

PR #192, opened 2016-07-20
F1 usart   buffered interrupt based tx (static tx buffer)

edogaldo:
Minimal changes to implement buffered interrupt-based TX with static buffer.

PR #260, opened 2017-03-25
8/16 Mhz support with proper wait-states

tomtor:
This adds 8/16 Mhz support.

Note that these changes are currently only done for STM32F1, they should be copied to the STM32F2, etc. trees.

I'm willing to tidy this up if you have interest in this pull


edogaldo:
Can't get the point for "make usart_putc really unbuffered", why?!

rogerclarkmelbourne:
I'm afraid this PR makes so many changes including compiler settings in platform.txt, that I can't risk pulling it

I like some of the changes like changing 72 to F_CPU, but the other changes are too radical for the main stream

tomtor:
@edogaldo The original master had a broken flush(), so either that had to be fixed or the output had to be made unbuffered. Note that the old master claimed that usart_putc was unbuffered, but it was not...

Current master appears to have a working flush() (I did not test it), so that's indeed the better solution.

edogaldo:
Ok, about the claim that usart_putc is unbuffered: originally USART TX was effectively unbuffered and blocking then, little less than 1 year ago, I pushed PR #192 to introduce buffered TX but missed that comment as my modifications didn't directly impact "usart_putc()".
Anyway, as a user, I can't see any benefit in going back to unbuffered & blocking USART TX..

tomtor:
@edogaldo I agree, only fixing the comment is needed

rogerclarkmelbourne:
I'll investigate making these changes a bit more generic and in separate commits

e.g. the change to the compiler setting to enable gnu11 would need to be a commit in its own right and widely tested.

The change to control flash wait states is also an independent change as it can already be applied to the existing 48Mhz setting.

So I'll do those 2 first and then come back to this PR and see what is left over.

tomtor:
Please note that gnu11 is the default for gcc5 and higher, so dealing with
it is more or less mandatory in the (near) future.

I can remove the usart_putc() changes (no longer needed) and gnu11 changes
from this commit, that would just leave the Mhz changes. Would that make it
easier for you?

Op vr 26 mei 2017 00:31 schreef Roger Clark <notifications@github.com>:

rogerclarkmelbourne:
I'm going add the gnu11 stuff today.

It was discussed on the forum http://stm32duino.com/viewtopic.php?f=3&t=1241&p=27885&hilit=gnu11#p27885

But I didnt want to make the change until I'd switched to using Releases so that people could go back to a non gnu11 version if necessary

rogerclarkmelbourne:
Added gnu11 and gnu++11 in

https://github.com/rogerclarkmelbourne/Arduino_STM32/commit/66579ef8ef822fd188c11efd4e682d75b95126d8

rogerclarkmelbourne:
Done the CYCLES_PER_MICROSECOND based on F_CPU change

https://github.com/rogerclarkmelbourne/Arduino_STM32/commit/f2c2323afde16917b5ba7059c677c54ef9ee0e8c

(Note, I didn't do this for the F4 as F_CPU may not be defined correctly for all boards)

rogerclarkmelbourne:
I've posted a question to the forum about the wait states calculation.

It looks OK to me, but I've not read the spec in detail to see what STM advises.

rogerclarkmelbourne:
Just one question

Why did you want a mode where the F_CPU is 16 via 8Mhz HSE x 2 ?

AFIK, you can have HSE x 1, by changing the clock input MUX to bypass the HSE PLL

tomtor:
@rogerclarkmelbourne I don't understand your question about F_CPU 16? F_CPU 16 is implemented with HSE x2 which is the correct way? What could be done differently is the way I implemented F_CPU 8Mhz. I used 8Mhz HSI RC-internal for that (the default boot mode).
You suggest bypassing the HSE PLL to get an 8Mhz clock based on the external crystal. That would be more accurate but also consume a bit more current. The advantage of using HSI is that after a wakeup from sleep, the clock used by the hardware is HSI. So on 8Mhz configuration I leave the clock unchanged after a wakeup. If we would use HSE x1 the wakeup code would have the restore that HSE clock configuration, which is ofcourse possible, but not tested by me. (I did test 16Mhz and 72Mhz with sleep and restoring the clock, that works OK)

I use the following code after wakeup:

#if F_CPU == 8000000UL
  // nothing to do, using about 8 mA
#elif F_CPU == 16000000UL
  rcc_clk_init(RCC_CLKSRC_HSI, RCC_PLLSRC_HSE , RCC_PLLMUL_2);
#elif F_CPU == 48000000UL
  rcc_clk_init(RCC_CLKSRC_HSI, RCC_PLLSRC_HSE , RCC_PLLMUL_6);
#elif F_CPU == 72000000UL
  rcc_clk_init(RCC_CLKSRC_HSI, RCC_PLLSRC_HSE , RCC_PLLMUL_9);
#else
#error "Unknown F_CPU!?"
#endif

For me the current 8Mhz from HSI looks like the more logical implementation in settings where the user opts for the low power 8Mhz clock combined with sleep modes.

rogerclarkmelbourne:
I can't quite see if you code is doing this, but if its wwitching to HSI just because F_CPU is set to 8Mhz , it could be problematic

I have a personal project where I use HSE 8Mhz, as I need accurate timekeeping but low power consumption, and I may not be alone in needing this configuration.

tomtor:
Yes, on boot I leave the clock on the default F_CPU 8 mhz HSI.

I can understand that F_CPU 8 mhz HSE is preferable in other contexts.
Giving the users a choice is probably the best.

Op vr 26 mei 2017 10:29 schreef Roger Clark <notifications@github.com>:

rogerclarkmelbourne:
I think we'd need to add a flag for HSI / HSE, to achieve that.

Also I'm not sure whether the option of 16Mhz would be necessary, because we'd get into the situation where we'd need all the multipliers.

I can see some people would also want to overclock. And Boards.txt ends up getting very bloated as we'd need to add the menu selection for all devices not just the F103C

I'm happy to put ifdefs in the core, and let advanced users edit board.txt or platform.txt etc to enable of disable features, but I have to weigh up the complication to add a feature to the menu's that only 1/1000 users will use.

tomtor:
I have never used the 16mhz myself, I just thought it was a nice step
between 8 and 48 mhz and for those occasions were 8mhz HSI was too
inaccurate.

I noticed just one time with LORA timing that HSI was less accurate. I had
to increase the clock tolerance to 1%, so the HSI is still quite accurate
in most cases. I did test the 8mhz HSI with I2C and SPI setups.

If you find a way to offer 8mhz HSE then an API call should be made
available to reset to this clock, eg with rcc_clk_init(), after a wakeup
from sleep.

Op vr 26 mei 2017 13:03 schreef Roger Clark <notifications@github.com>:

edogaldo:
For what's worth, my opinion is that having too many clock options is a
complication that goes beyond the Arduino spirit.

rogerclarkmelbourne:
http://www.stm32duino.com/viewtopic.php?f=53&t=2286

rogerclarkmelbourne:
This PR now has conflicts due to subsequent changes.

I don't think its worth adding 8 and 16 Mhz options to the menu, as they would be so rarely used.

However I should implement your change to set the correct flash wait states for 48Mhz, which is now an option.

Also, now that we have a 128Mhz overclock option, I suspect the wait states is incorrect as it will be using the same wait states as for 72Mhz


Perhaps there is some other way to implement these changes e.g. as a library or code snippet that people could use to select 8 or 16Mhz if they want.

tomtor:
Note that adjusting the waitstates needs an additional change to maintain correct timing, also for current 48mhz clock:

See
https://github.com/tomtor/Arduino_STM32/commit/2ab804224b813b419b3a9faa5877448e33590962

I fail to see the advantage of offering the current 48mhz clock. The change in current consumption is minimal.

rogerclarkmelbourne:
"I fail to see the advantage of offering the current 48mhz clock. The change in current consumption is minimal."

Yes. Probably not worth it.

I'll remove it when I get chance

rogerclarkmelbourne:
I've had feedback from various people on the forum, indicating that 48Mhz does have its use.
I can't quite remember what the use was, I think it may be something to do with being able to run SPI at 24Mhz, rather than 36Mhz or 18Mhz etc

So I'm keeping the 48Mhz option.


I like the changes to the wait states, but the changes would need to be integrated deeper in the core rather than in just one variant

rogerclarkmelbourne:
I've manually merged your change for wait state optimisation

https://github.com/rogerclarkmelbourne/Arduino_STM32/commit/7ef6e71a781270f69b50e0c113b8ed9e1aa8c15a

rogerclarkmelbourne:
@tomtor 

I'm not sure why you did this

+#if F_CPU > 8000000
     // Enable HSE, and wait until it's ready.
     // Enable HSE, and wait until it's ready.
     rcc_turn_on_clk(RCC_CLK_HSE);
     rcc_turn_on_clk(RCC_CLK_HSE);
     while (!rcc_is_clk_ready(RCC_CLK_HSE))
     while (!rcc_is_clk_ready(RCC_CLK_HSE))
         ;
         ;
+#endif

Won't this completely disable the HSE for frequencies 8Mhz or lower ?

I though the purpose of the PR was to allow 8MHz operation by bypassing the PLL, but this seems to do something different

tomtor:
@rogerclarkmelbourne Yes, it does disable the HSE for 8Mhz and lower.

At the time I wrote the code I was not aware of the option to bypass the PLL. Disabling the HSE for low frequency operation should however also reduce the power consumption, and reducing the power consumption is the reason for lowering the frequency in the first place.
On the other hand the internal clock is less accurate, so in some cases using HSE with the crystal is better. Offering a choice to the user would be the best?

rogerclarkmelbourne:
I can add the option to disable HSE but it can't be disabled based on FCPU frequency

It will need to be a separate # define 

As would the option to bypass the PLL

I'm also not sure if simply not turning on the HSE does enough to make the HSI work, but perhaps they are HSI is enabled by default ?

rogerclarkmelbourne:
Its something to do with the Leaflabs docs generator and various other things that we generally dont use

tomtor:
HSI is always enabled on boot. I remember now a second reason, if you wake up from deep sleep then you  are always in 8Mhz HSI mode. So this 8Mhz/HSI mode is the most natural low power mode.

I use this code:

    void AlarmFunction () {
      // We always wake up with the 8Mhz HSI clock!
      // So adjust the clock if needed...
    
      #if F_CPU == 8000000UL
        // nothing to do, using about 8 mA
      #elif F_CPU == 16000000UL
        rcc_clk_init(RCC_CLKSRC_HSI, RCC_PLLSRC_HSE , RCC_PLLMUL_2);
      #elif F_CPU == 48000000UL
        rcc_clk_init(RCC_CLKSRC_HSI, RCC_PLLSRC_HSE , RCC_PLLMUL_6);
      #elif F_CPU == 72000000UL
        rcc_clk_init(RCC_CLKSRC_HSI, RCC_PLLSRC_HSE , RCC_PLLMUL_9);
      #else
        #error "Unknown F_CPU!?"
      #endif

      extern volatile uint32 systick_uptime_millis;
      noInterrupts();
      systick_uptime_millis += sleepTime;
      interrupts();
    }

stevstrong:
As this PR has got not more activity, no more users have missed this, and it does not seem to solve any real problem, I will close this without merging.
Any special HSI handling can be done anyway on individual basis.

PR #261, opened 2017-03-25
Free rtos900

victorpv:
This is the port of FreeRTOS v9.00
I have tested it with several sketches successfully.
There should not be any changes to any other files in other folders, so it's safe if you pull and add for everyone to use.


victorpv:
Roger, this has been tested now by at least me, Pito, and Michael, and seems to work fine in every case, even improved performance slightly in Pito's case. I think it is safe to merge, besides doesn't modify anything existing.

PR #263, opened 2017-03-27
Add PB1 as ADC and PWM pin

mmonsterski:


stevstrong:
Seems fine for me.

mmonsterski:
Sorry but this pin in used as default as ERROR_PIN. It shouldn't be used as ADC or PWM pin. Sorry for my mistake

PR #270, opened 2017-04-18
Fix if serial

SukkoPera:
The [Arduino API uses ```if (!Serial)```](https://www.arduino.cc/en/Serial/IfSerial) to check if a serial connection is established on USB CDC ports, like on The Leonardo/Micro or Due, while libmaple was using ```Serial.isConnected()```.

This patch switches the latter to the former and adds a (deprected) wrapper method to maintain backwards compatibility with sketches that use the libmaple interface.

(See http://www.stm32duino.com/viewtopic.php?f=28&t=117&p=26487#p26425 and subsequents posts.)

### DIFF
--- a/STM32F1/cores/maple/usb_serial.cpp
+++ b/STM32F1/cores/maple/usb_serial.cpp
@@ -186,7 +186,7 @@ uint8 USBSerial::getRTS(void) {
     return usb_cdcacm_get_rts();
 }
 
-uint8 USBSerial::operator bool() {
+USBSerial::operator bool() {
     return usb_is_connected(USBLIB) && usb_is_configured(USBLIB) && usb_cdcacm_get_dtr();
 }
 
--- a/STM32F1/cores/maple/usb_serial.cpp
+++ b/STM32F1/cores/maple/usb_serial.cpp
@@ -178,16 +178,16 @@ uint8 USBSerial::pending(void) {
     return usb_cdcacm_get_pending();
 }
 
-uint8 USBSerial::isConnected(void) {
-    return usb_is_connected(USBLIB) && usb_is_configured(USBLIB) && usb_cdcacm_get_dtr();
-}
-
 uint8 USBSerial::getDTR(void) {
     return usb_cdcacm_get_dtr();
 }
 
 uint8 USBSerial::getRTS(void) {
     return usb_cdcacm_get_rts();
+}
+
+uint8 USBSerial::operator bool() {
+    return usb_is_connected(USBLIB) && usb_is_configured(USBLIB) && usb_cdcacm_get_dtr();
 }
 
 #if BOARD_HAVE_SERIALUSB
--- a/STM32F1/cores/maple/usb_serial.h
+++ b/STM32F1/cores/maple/usb_serial.h
@@ -44,19 +44,19 @@ public:
 
     void begin(void);
 
-	// Roger Clark. Added dummy function so that existing Arduino sketches which specify baud rate will compile.
-	void begin(unsigned long);
-	void begin(unsigned long, uint8_t);
+    // Roger Clark. Added dummy function so that existing Arduino sketches which specify baud rate will compile.
+    void begin(unsigned long);
+    void begin(unsigned long, uint8_t);
     void end(void);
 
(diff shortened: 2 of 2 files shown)


SukkoPera:
I second that. I say isConnected() is deprecated in the comment preceding it, I didn't know there was an attribute for the purpose.

victorpv:
From GCC documentation:
====================
deprecated
deprecated (msg)
The deprecated attribute results in a warning if the function is used anywhere in the source file. This is useful when identifying functions that are expected to be removed in a future version of a program.
====================
Since SukkoPera is adding another method for the same purpose, which is the one compatible with the official API (!Serial), I think it would be good to add the deprecated attribute but with a message, to guide user to use the official Arduino method.

__attribute__((deprecated(Use !Serial instead.)))

That will result in:
1.- We keep compatibility with libmaple Serial.isConnected
2.- We add compatibility with Arduino !Serial
3.- We warn users of isConnected to rather use the official API, without breaking any sketch.

That will provide a meaninful warning on why is the function deprecated and what to use instead.
The lack of !serial pops frequently in the forum, pretty much every time a new user of the core tries to port the first sketch from an Arduino core, so I think it's important to add it.

Regarding getting extra warnings, we currently have several messages about functions being deprecated (multiple of dma ones), done ages ago by libmaple, but since the warning doesn't say what to use instead, everyone just ignores them.
Using the message parameter we can guide people how to avoid the message rather than ignore it.

rogerclarkmelbourne:
@victorpv 

OK.

Sounds good.

Can @SukkoPera update the PR so that the original function is retained by given the deprecated attribute and message as indicated by @victorpv

SukkoPera:
@victorpv: Be careful, I don't think that !this calls the boolean conversion operator. I think it merely checks if this != NULL, which is hopefully always true :). That's why I implemented isConnected() as  ```return (bool) *this;```.

@rogerclarkmelbourne: The function had already been kept for backwards compatibility in my first PR proposal. I think I only need to add the deprecated attribute, which I will do tonight.

stevstrong:
+1 for this commit, I already integrated the solution in my black_F4 branch, working nicely.

victorpv:
Sukko I'll check in the debugger what exactly does to confirm.
In any case we need a change in the write function as it wouldn't be wise to use a function we call deprecated.

SukkoPera:
Added attribute, removed usage of isConnected() and tested everything, works fine :).

For the record, this is the message I get from GCC if I use isConnected():

> warning: 'uint8 USBSerial::isConnected()' is deprecated (declared at /home/sukko/Development/arduino/hardware/Arduino_STM32/STM32F1/cores/maple/usb_serial.h:83): Use !Serial instead [-Wdeprecated-declarations]
>    while (!Serial.isConnected())

rogerclarkmelbourne:
I'll pull to a local branch and test.

BTW. As isConnected() is used in the examples e.g.

https://github.com/rogerclarkmelbourne/Arduino_STM32/search?utf8=%E2%9C%93&q=isConnected%28%29&type=

They would all need to be updated at the same time to match this change.

SukkoPera:
I will fix the examples too, ASAP.

edogaldo:
+1 also for me.

Best, E.

rogerclarkmelbourne:
I don't see a problem with keeping isConnected(), but I think we should add the bool operator, as looking at the latest Arduino examples that come as part of the IDE a lot of them use while (!Serial);

I will attempt to integrate the bool operator tomorrow, but I've not had time to do it today, as I've processed about 6 other changes ;-)

edogaldo:
Just to clarify, I'm with the bool operayor too, just wanted to point out
that I'm also with keeping the isConnected () too.

Il 25 Mag 2017 13:17, "Roger Clark" <notifications@github.com> ha scritto:

SukkoPera:
I will probably be able to work on this I'm the weekend, too. Sorry but I've been pretty busy in the last few days.

SukkoPera:
Examples fixed.
Tested a few, seem to work fine.
Also fixed a conflict due to recent changes upstream.

rogerclarkmelbourne:
Thanks

PR #271, opened 2017-04-22
Fixed utoa() function definition

grafalex82:
utoa accepts unsigned _long_ for some reason, while its counterpart itoa accepts _int_.

I figured this out when fixing a warning on my project, but I do not remember which one

### DIFF
--- a/STM32F1/cores/maple/itoa.c
+++ b/STM32F1/cores/maple/itoa.c
@@ -121,7 +121,7 @@ extern char* ltoa( long value, char *string, int radix )
   return string;
 }
 
-extern char* utoa( unsigned long value, char *string, int radix )
+extern char* utoa( unsigned value, char *string, int radix )
 {
   return ultoa( value, string, radix ) ;
 }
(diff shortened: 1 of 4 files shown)


lacklustrlabs:
I've tested this on the Arduino provied gcc (arm-none-eabi-gcc/4.8.3-2014q1) as well as [gcc 6.3 bleeding-edge-toolchain-170314](http://www.freddiechopin.info/en/download/category/11-bleeding-edge-toolchain) and it works just fine. 

On gcc 6.3 -Wall -Wextra this fix is required.

(Disclaimer: I've only tested it on maple mini)

rogerclarkmelbourne:
http://www.stm32duino.com/viewtopic.php?f=53&t=2284

rogerclarkmelbourne:
Looks like I will manually need to merge this as Git claims there are conflicts

rogerclarkmelbourne:
I manually updated the files in 

https://github.com/rogerclarkmelbourne/Arduino_STM32/commit/2c33a0237134df4d5337fca38c17b9d1d74a4b8a

But, GitHub still thinks there is a conflict that makes your files un-mergable

Please check I've updated the files correctly

PR #275, opened 2017-04-27
Added the macro digitalPinToInterrupt()

lacklustrlabs:
As I understand it the interrupt numbers and pin numbers are the same on stm32duino.
So this macro does nothing, it just provides a little bit of Arduino compatibility.

### DIFF
--- a/STM32F1/variants/generic_stm32f103c/variant.h
+++ b/STM32F1/variants/generic_stm32f103c/variant.h
@@ -1,6 +1,10 @@
 #ifndef _VARIANT_ARDUINO_STM32_
 #define _VARIANT_ARDUINO_STM32_
 
+#ifndef digitalPinToInterrupt
+#define digitalPinToInterrupt(P)   ( P )
+#endif 
+
 #define digitalPinToPort(P)        ( PIN_MAP[P].gpio_device )
 #define digitalPinToBitMask(P)     ( BIT(PIN_MAP[P].gpio_bit) )
 #define portOutputRegister(port)   ( &(port->regs->ODR) )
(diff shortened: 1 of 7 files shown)


victorpv:
There is another pull including this same function, but does it a different way.
https://github.com/rogerclarkmelbourne/Arduino_STM32/pull/225

I think would be positive if one of you post in the forum in stm32duino  a new thread about this to collect all the input and submit a single solution for that.

rogerclarkmelbourne:

stevstrong:
Here is the correct link: http://www.stm32duino.com/viewtopic.php?f=53&t=2281

rogerclarkmelbourne:
I think the problem with this PR is that its adding the same code in the multiple variants when I think it needs to be added somewhere central instead

rogerclarkmelbourne:
I have actioned this in a different way, by putting the definition into wirish.h

stevstrong:
It looks to me that the variant.h files contain identical information. Couldn't we put these info to another file which is common to all variants? For example util.c?

rogerclarkmelbourne:
I've already put it in wirish.h

https://github.com/rogerclarkmelbourne/Arduino_STM32/commit/c5caa5d288ca6c4a08fb8fe7e2125e6988cd1065

stevstrong:
Yes, but I meant all the defines like digitalPinToPort(), they are all common to all variants.
There is no real "variant" specific information in it.

rogerclarkmelbourne:
Does it need to change from variant to variant ?

I've tried it with both a Maple mini and a Blue Pill using the different naming schemes and it seems to work for both.

But perhaps I don't understand what the pin number that is passed to the macro

I thought it was just the pin name / number e.g. PA8 ??

stevstrong:
Sorry, I think I could not make it clear, it is not specially about this single define, but I was saying about all the defines which are duplicated in all those variant.h files, they are all identical information (expect now the LED_BUILTIN).
I thought they could me moved to a common file.

PR #277, opened 2017-04-28
Maple mini was missing BOARD_BUTTON_PIN

lacklustrlabs:
This will make the examples WhileStatementConditional, QASlave, InteractiveTest, Button, StateChangeDetection, StateChangeDetection and Debounce compile w/o any modifications on maple mini.

So far i've compiled all of the examples and tested Button, StateChangeDetection & Debounce on maple mini.


lacklustrlabs:
Ok, i see. BOARD_LED_PIN was removed from those examples while BOARD_BUTTON_PIN managed to sneak past review.

PR #278, opened 2017-05-06
To use STM32ADC in continuous mode triggered by an internal timer

syfre:
No sure for the compile error in util_adc.c but ADC_CR2_TSEREFE don't exists

### DIFF
--- a/STM32F1/cores/maple/HardwareTimer.cpp
+++ b/STM32F1/cores/maple/HardwareTimer.cpp
@@ -142,6 +142,11 @@ void HardwareTimer::refresh(void) {
     timer_generate_update(this->dev);
 }
 
+void HardwareTimer::setMasterModeTrGo(uint32_t mode) {
+	this->dev->regs.bas->CR2 &= ~TIMER_CR2_MMS;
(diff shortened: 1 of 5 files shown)


rogerclarkmelbourne:
Thanks.

I'll need to get this tested before I can merge the PR

victorpv:
There is a thread about ADC_CR2_TSEREFE, apparently was a flag that was named wrong and it was corrected last year, that's why it does not exist any more:
http://www.stm32duino.com/viewtopic.php?t=1317&start=20

So renaming that to TSV... is the right thing to do, since looks like that was just 1 miss from last time.

rogerclarkmelbourne:
@victorpv 

Thanks

What about the change 

cr2 |= ADC_CR2_EXTTRIG;

?

stevstrong:
`cr2 |= ADC_CR2_EXTTRIG;`
is needed in order to select the external event which triggers the ADC via EXTSEL.
Looks fine for me.

rogerclarkmelbourne:
OK. I will merge...

Thanks

PR #286, opened 2017-05-15
Compiler fixes for Link Time Optimization

grafalex82:
I have custom (CMake based) build system. I tried to use -flto (Link Time Optimization) switch, but faced 2 issues

The first one is a warning like this:
```
D:\GRAFAL~1\Hobbies\Prog\MYPROJ~1\GPSLOG~3\Libs\STM32D~1\system\libmaple\include/libmaple/timer.h:138:18: warning: type of 'timer1' does not match original declaration
 extern timer_dev timer1;
                  ^
D:\Graf Alex\Hobbies\Prog\MyProjects\GPSLogger3\Libs\STM32duino\cores\maple\libmaple\timer.c:53:11: note: previously declared here
 timer_dev timer1 = ADVANCED_TIMER(1);
```
This object is used in boards.cpp file only, each reference is getting address of this object. Fortunately there is a TIMER1 pointer already defined as a pointer to timer1. So the solution is pretty straightforward: use TIMERx pointers instead of &timerx in boards.cpp. This is not an issue for F3 and F4 - only for F1.

I tested this change with blue pill board only (F103C), but the change is pretty simple so I could blindly make it for other borads as well.

The second problem is inability to link vTaskSwitchContext() from FreeRTOS. This function is referenced twice - one in .c file, but from unused function, second one from a piece of assembler code. For some reason linker can't figure out the second case, throws vTaskSwitchContext() function out as unused, and then reports for an unresolved symbol from an assembler function.

The solution is to mark the function with __attribute__((used)). Well, this is gcc only attribute, but we are not using other compilers anyway. So I think the change is ok for now. I am using FreeRTOS 8.2.1, but I also duplicated this change for another FreeRTOS copy as well.

### DIFF
--- a/STM32F1/cores/maple/libmaple/adc_f1.c
+++ b/STM32F1/cores/maple/libmaple/adc_f1.c
@@ -40,10 +40,12 @@
  * Devices
  */
 
+
+voidFuncPtr adc1_handlers[] = {[0 ... 2]=0}; //added by bubulindo. EOC, JEOC, AWD
 adc_dev adc1 = {
     .regs   = ADC1_BASE,
     .clk_id = RCC_ADC1,
-    .handlers = {[3]=0}, //added by bubulindo. EOC, JEOC, AWD
+    .handlers = adc1_handlers, 
     .irq_num = NVIC_ADC_1_2,
 };
 /** ADC1 device. */
@@ -57,10 +59,11 @@ adc_dev adc2 = {
 adc_dev *ADC2 = &adc2;
 
 #if defined(STM32_HIGH_DENSITY) || defined(STM32_XL_DENSITY)
+voidFuncPtr adc3_handlers[] = {[0 ... 2]=0}; //added by bubulindo. EOC, JEOC, AWD
 adc_dev adc3 = {
     .regs   = ADC3_BASE,
     .clk_id = RCC_ADC3,
-    .handlers = {[3]=0}, //added by bubulindo. EOC, JEOC, AWD
+    .handlers = adc3_handlers,
     .irq_num = NVIC_ADC3,//added by bubulindo. 
 };
 /** ADC3 device. */
--- a/STM32F1/cores/maple/libmaple/gpio_f1.c
+++ b/STM32F1/cores/maple/libmaple/gpio_f1.c
@@ -142,7 +142,6 @@ gpio_pin_mode gpio_get_mode(gpio_dev *dev, uint8 pin) {
     gpio_reg_map *regs = dev->regs;
     __io uint32 *cr = &regs->CRL + (pin >> 3);
     uint32 shift = (pin & 0x7) * 4;
-    uint32 tmp = *cr;
 
 	uint32 crMode = (*cr>>shift) & 0x0F;
 	
--- a/STM32F1/cores/maple/libmaple/gpio_f1.c
+++ b/STM32F1/cores/maple/libmaple/gpio_f1.c
@@ -142,6 +142,7 @@ gpio_pin_mode gpio_get_mode(gpio_dev *dev, uint8 pin) {
     gpio_reg_map *regs = dev->regs;
     __io uint32 *cr = &regs->CRL + (pin >> 3);
     uint32 shift = (pin & 0x7) * 4;
+    uint32 tmp = *cr;
 
 	uint32 crMode = (*cr>>shift) & 0x0F;
 	
--- a/STM32F1/cores/maple/libmaple/timer.c
+++ b/STM32F1/cores/maple/libmaple/timer.c
@@ -50,72 +50,72 @@ static inline void enable_irq(timer_dev *dev, timer_interrupt_id iid);
  */
 
 #if STM32_HAVE_TIMER(1)
-timer_dev timer1 = ADVANCED_TIMER(1);
+DEFINE_ADVANCED_TIMER(timer1, 1);
 /** Timer 1 device (advanced) */
 timer_dev *const TIMER1 = &timer1;
 #endif
 #if STM32_HAVE_TIMER(2)
-timer_dev timer2 = GENERAL_TIMER(2);
+DEFINE_GENERAL_TIMER(timer2, 2);
 /** Timer 2 device (general-purpose) */
 timer_dev *const TIMER2 = &timer2;
 #endif
 #if STM32_HAVE_TIMER(3)
-timer_dev timer3 = GENERAL_TIMER(3);
+DEFINE_GENERAL_TIMER(timer3, 3);
 /** Timer 3 device (general-purpose) */
 timer_dev *const TIMER3 = &timer3;
 #endif
 #if STM32_HAVE_TIMER(4)
-timer_dev timer4 = GENERAL_TIMER(4);
+DEFINE_GENERAL_TIMER(timer4, 4);
 /** Timer 4 device (general-purpose) */
 timer_dev *const TIMER4 = &timer4;
 #endif
 #if STM32_HAVE_TIMER(5)
-timer_dev timer5 = GENERAL_TIMER(5);
+DEFINE_GENERAL_TIMER(timer5, 5);
 /** Timer 5 device (general-purpose) */
 timer_dev *const TIMER5 = &timer5;
 #endif
 #if STM32_HAVE_TIMER(6)
-timer_dev timer6 = BASIC_TIMER(6);
+DEFINE_BASIC_TIMER(timer6, 6);
 /** Timer 6 device (basic) */
 timer_dev *const TIMER6 = &timer6;
 #endif
 #if STM32_HAVE_TIMER(7)
-timer_dev timer7 = BASIC_TIMER(7);
+DEFINE_BASIC_TIMER(timer7, 7);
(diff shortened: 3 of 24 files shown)


stevstrong:
I don't think the changes are ok.
If you replace `&timer1` by `&TIMER1`, then the table will include a pointer to a pointer of a timer device, not a pointer to a timer device.

grafalex82:
Good catch. I have not pushed a commit, that fixes that

rogerclarkmelbourne:
Does this change the binary size or RAM usage.?

I think @victor_pv tried making similar changes in the past to attempt to move more of the data structures out of the RAM, but it ended up making things worse.

Also, we have issues where globally instantiated constructors (normally in libraries) attempt to use pinMode, but this fails because the PIN_MAP global structure is not initialised before the library constructor is called.

AFIK the issue with globally instantiated constructors still exists but, as the order of initialisation of global structs is not defined by the C language, the current Core code seems to be compiled in a way that works for most libraries.

This change however may change the order of initialisation so that some libraries no longer work.

Hence, as this is such a major change, it would need to be extensively tested before being released.

grafalex82:
Hmmmm. Surprisingly you are right. While overall binary size remains the same, it moved PIN_MAP from .rodata to RAM. 

I'll try to search for some other change

rogerclarkmelbourne:
@grafalex82 

No worries.  We've been here before ;-)

You could try contacting @victorpv  to see if he can remember what he tried doing, when he was investigating a similar change.

victorpv:
@grafalex82 Roger is partially correct, but at the end we succeeded in moving the whole PIN_MAP structure to flash as a constant, and that was thanks precisely to using those pointers.
If you change them, PIN_MAP goes to RAM as you found.
Not sure why the linker does that, but it does, and the workaround was using &timer1 etc. Basically it creates a PIN_MAP table in RAM that's missing a lot of values, and those values have to be populated later on at run time, so the linker places it in RAM to have write access

If you find another work around that keeps PIN_MAP in flash, please test, and ideally open a thread in the forum so more people can test and confirm stability, then it could be pushed.

About the second issue, if your solution doesn't affect anything else, I think would be mostly safe to push. Worst case would be that the function gets included even if not called, correct?
Has you checked if the FreeRTOS v9 has the same issue? I sent a PR for it, not sure if it is merged yet, but if not, you can pull the code from the branch the PR came from.

grafalex82:
Aha, I figured out the problem and seems to have a solution (not tested yet)

Now I took a deeper look an see what the problem is. The cause is that timer_dev structure has variable size due to handlers[] array. This means that specific timers can be created with different number of handlers in the list. Compiler (actually linker) generates the warning because it can't calculate exact object size just by extern declaration.

I have a solution. It seems that timer1 object is statically allocated, no one is copying it. Everyone else in the code is using TIMER1, which is pointer to timer1. Perhaps we could use some kind of structure inheritance: structure is split into public part (first 3 fields) and a private part (handlers array) which is visible only in timer.c

It could look like this:
Header
```
/** Timer device type */
typedef struct timer_dev {
    timer_reg_map regs;         /**< Register map */
    rcc_clk_id clk_id;          /**< RCC clock information */
    timer_type type;            /**< Timer's type */
} timer_dev;
```


timer.c 
```
/** Timer device type */
typedef struct timer_dev_pvt 
{
    timer_dev dev; // Not a pointer - whole stucture is placed here
    voidFuncPtr handlers[];     /**<
                                 * Don't touch these. Use these instead:
                                 * @see timer_attach_interrupt()
                                 * @see timer_detach_interrupt() */
} timer_dev_pvt;

timer_dev_pvt timer1 = ADVANCED_TIMER(1);
timer_dev *const TIMER1 = &timer1;

```
timer_dev is a fixed size structure, so Link Time Optimization applies well. All the code which needs the handlers field will just cast passed timer_dev pointer to timer_dev_pvt one. 

As you suggested, I created a forum topic: http://stm32duino.com/viewtopic.php?f=3&t=2092

rogerclarkmelbourne:
@victorpv 
Thanks.
I couldn't remember whether your changes were 100% effective at moving the PIN_MAP to flash.

grafalex82:
Please check out my changes. I did timer_dev and adc_dev structures as fixed size. Instead of array of handlers now these structures have **a pointer** to array of handlers.

In case of timers I had to change initialization macros. Macro were renamed from ADVANCED_TIMER to DEFINE_ADVANCED_TIMER (an so on for other timers). This change should reflect nature of the macro. Now they define variables (timer1, timer2, ...), not just initialization structure for them.  Handler arrays are also defined in those macros.

Despite I could generate variable name inside the macro (e.g. timer##num), I decided that it would be better to accept variable name explicitly - this will help searching for a definition by variable name.

I also fixed array initialization (in case of adc handlers there was an array overrun). Now it explicitly specify elements range that is initialized.

I tested on a F103CB/BluePill. Blink example (using timer) works fine. Let me know how I can easily to run a deeper test.

As for memory questions. 
- PIN_MAP is still in the flash
- .text segment is around the same (+8 bytes)
- .data segment reduced by 112 bytes
- .bss segment increased by 128 bytes

Last 2 items means that some of the objects (particularly handler arrays - 108 bytes) moved from data to bss. Total .data + .bss change is +16 bytes due to extra pointers. 

Hope this is not a big deal. Esspecially as this change opens new opportunity - define timer_dev objects as constants and move them to flash as well (handler arrays remain non-constant). This will also imply making TIMERx pointers constant as well, mark all functions accepting timer_dev pointer to accept constant pointer. Perhaps this will also allow compiler to better optimize code around this. Though this should be a separate change.

rogerclarkmelbourne:
Thanks.

I noticed you changed from freertos code. So that would need to be tested as well.
I don't personally use freertos but I think @victorpv  does.

Re: Extra 16 bytes in RAM.
I'm sure some people will complain about this on the forum, (yes really they will)

However IMHO 16 bytes is not a big deal, if it enhances the core.

What I still dont understand is what you were trying to compile they required link time optimisation ( -flto )

I know you have not changed platform.txt, so enable it for normal builds, but I suspect it would be too high a risk for me to do this, because there are probably some libraries that would not work with that option enabled.

Re: Testing

You'd need to test anything which uses timers and also anything that uses ADC
e.g The first thing that springs to mind is PWM which uses the timers.

For ADC, @ahull's pig-o-scope would be a good test, as it uses ADC with DMA and well as SPI

There is a massive thread on the forum for the pig-o-scope, (in fact I think there are multiple threads on oscilloscopes as there are enhanced pig-o-scope code being developed), so that's probably he the best starting point to get the latest code.

And FreeRTOS would need to be tested - Its not something I use but there are examples in the core

victorpv:
I gave a test sketch that uses freertos + 2 spi + timer + dma. Doesn't use adc though.
I need to check the changes, but why was freertos modified? It doesn't use any of the timer or adc functions as far as I remember.

rogerclarkmelbourne:
@victorpv 
Thanks

If I get chance I'll see if I can get the pig-o-scope to work on one of the display boards which I use for displaying home power (mains) usage. As those boards have a Maple mini and a ILI9341

grafalex82:
> I noticed you changed from freertos code. So that would need to be tested as well.
Tested with and without lto

> What I still dont understand is what you were trying to compile they required link time optimisation ( -flto )
From what I see LTO works well on C code. For instance it works great on various initialization routines which implements multiple cases/ifs, while in fact used just with 1 values. In case of my project (~100 files, ~60k resulting binary) it saves about 3.5k. Optimization mostly happens in libmaple.

> I know you have not changed platform.txt, so enable it for normal builds, but I suspect it would be too high a risk for me to do this, because there are probably some libraries that would not work with that option enabled.
I'd like to make build system changes as a separate PR. I still have difficulties to make a good cross platform configuration. In the meantime we can focus on code part, these changes are low risk and can be done independently.

grafalex82:
Surprisingly turning LTO on for arduino build system went smooth. I just added -flto flag and it compiled/linked well. However it may require a full rebuild.

Stats for a simple blink over timer ISR program (flash/RAM)
Without LTO: 14204/2864
With LTO: 12548/2832

I tested my changes in the following scenarios:
- Blink using timer ISR
- Timer generated PWM
- ADC in single conversion mode signalin end of conversion with interrupt

I think I can't test with pig-o-matic as I do not have necessary hw items.

Anythin I can do to move forward with this PR?

rogerclarkmelbourne:
Interesting.

If you compile something more complex e.g. with SPI and other libraries, how does the size compare.

grafalex82:
My project uses SPI, I2C, UART, FreeRTOS, AdafruitSSD1306 and NeoGPS (not bundled with stm32duino). Total optimization is about ~3.5k on a ~60k firmware.

rogerclarkmelbourne:
This change is a bit too risky to me merged at the moment. 

Please don't delete it. and I'll take a look at it when there is less flux in the repo

victorpv:
Roger can you merge this one to a separate branch?
I'll download and test it with my wav player code. That uses DMA in 2 SPI ports with interrupts and callbacks, timers with DMA and interrupts, and a button. Doesn't use any ADC though.

rogerclarkmelbourne:
This has been added by the Optimisation menu

PR #288, opened 2017-05-20
Pinmode timer

MicroBahner:
Hi,
as I mentioned in the forum [http://www.stm32duino.com/viewtopic.php?f=49&t=2079](url), the pinMode() function disables timer IRQs if setting non pwm modes on pwm capable pins. 
This little fix prevents this. It disconnects the timer from the pin, but leaves all other timer configurations as they are.

### DIFF
--- a/STM32F1/cores/maple/stm32f1/wirish_digital_f1.cpp
+++ b/STM32F1/cores/maple/stm32f1/wirish_digital_f1.cpp
@@ -79,10 +79,14 @@ void pinMode(uint8 pin, WiringPinMode mode) {
 
     gpio_set_mode(PIN_MAP[pin].gpio_device, PIN_MAP[pin].gpio_bit, outputMode);
 
-    if (pwm && PIN_MAP[pin].timer_device != NULL) {
-        /* Enable/disable timer channels if we're switching into PWM. */
+    if (PIN_MAP[pin].timer_device != NULL) {
+        if ( pwm ) { // we're switching into PWM, enable timer channels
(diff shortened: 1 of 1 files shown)


MicroBahner:
Do you really think so? If you call the pinMode function with a non PWM mode, you probably don't want a PWM signal at that pin. And the only thing that is done now, is to clear the output enable Flag for that channel. That means, the timer channel is disconnected from that pin - nothing else. Every thing else ist left as it is. The main problem with the current code is, that it detaches the channel interrupts.

stevstrong:
Seems feasible.

edogaldo:
I'm not very good with timers so I'm probably wrong, I was wondering if a scenario where 2 pins can map to the same couple {timer, channel} could be possible (even in the future).
In this case when using one pin as pwm and you change the pinmode of the other you could break the first..

stevstrong:
I don't think it is possible to map 2 different pins to the same timer/channel combination simultaneously.
If any further (not yet known/used scenario) would need another change, it could be still done in the future.
But for now, I think this is safe.

MicroBahner:
It's definitely not possible to map 2 pins to the same timer/channel simultaneously. But it's possible to have different peripherals connected to one pin alternatively. That's why I decided to disconnect the timer from the pin in non PWM mode.
But edogaldo is not completely wrong: You can map timerchannels to another pin alternatively. But if you do so, there are other problems, because until now the core assumes a fixed relationship between pins and timerchannels. If you change that, even setting the PWM mode will not work correctly.

rogerclarkmelbourne:
Thanks

PR #220, opened 2016-10-21
improved USB serial module

stevstrong:
- improved Rx routine with bugfix for loosing bytes when reading very slowly or very fast the Rx buffer (see http://www.stm32duino.com/viewtopic.php?f=9&t=1451)
- reworked Tx routine - implemented buffered transmission, which improves significantly the Tx speed

PR #290, opened 2017-05-22
F1 spi development

stevstrong:
I tried to collect only the SPI changes, so that this PR should supersede #220 .

Tested with the SdFatEX bench without DMA:
```
File size 5 MB
Buffer size 512 bytes
Starting write test, please wait.

write speed and latency
speed,max,min,avg
KB/Sec,usec,usec,usec
2385.34,15396,205,213
2422.33,15326,205,209

Starting read test, please wait.

read speed and latency
speed,max,min,avg
KB/Sec,usec,usec,usec
1905.37,1103,266,267
1905.37,1105,266,267
```
and with DMA:
```
write speed and latency
speed,max,min,avg
KB/Sec,usec,usec,usec
3720.00,13837,131,136
3282.78,181529,131,154

Starting read test, please wait.

read speed and latency
speed,max,min,avg
KB/Sec,usec,usec,usec
3225.60,1512,157,157
3227.68,1184,156,157
```
For more info see the comments of [last commit](https://github.com/rogerclarkmelbourne/Arduino_STM32/pull/290/commits/f1608abe0b8e250d22999a804fc105f391974305)


rogerclarkmelbourne:
I've tested with my ILI9341 display, and the graphic test is working, but some activities are 10% slower in the new version

First column is the activity, second column is the current time for that activity (master branch), the third column is the new SPI time, the last column is the time increase in perentage

Activity | time for master branch | time for new PR code | time increase in percentage for PR code
------------ | -------------
Screen fill | 170786|170798|0.007026337
Text	                                43086		46578		1.081047208
Lines	                                242465		268270		1.106427732
Horiz/Vert Lines	        15811		16085		1.017329707
Rectangles (outline)	        11569		11954		1.033278589
Rectangles (filled)	        355069		355159		1.000253472
Circles (filled)	                143286		156881		1.09488017
Circles (outline)	                172906		189445		1.095653129
Triangles (outline)	        59728		66434		1.11227565
Triangles (filled)	                166040		173251		1.043429294
Rounded rects (outline)	59772		64818		1.084420799
Rounded rects (filled)	415712		419931		1.010148853

stevstrong:
Yes, it is about the function setDataSize(), originally it was:
````
void SPIClass::setDataSize(uint32 datasize)
 {
 	_currentSetting->dataSize = datasize;
	uint32 cr1 = _currentSetting->spi_d->regs->CR1 & ~(SPI_CR1_DFF); 
	_currentSetting->spi_d->regs->CR1 = cr1 | (datasize & SPI_CR1_DFF);
}
````
the new version has one more line and additional stuff:
````
void SPIClass::setDataSize(uint32 datasize)
 {
	uint32 cr1 = _currentSetting->spi_d->regs->CR1 & ~(SPI_CR1_DFF);
	uint8 en = spi_is_enabled(_currentSetting->spi_d);
	spi_peripheral_disable(_currentSetting->spi_d);
	_currentSetting->spi_d->regs->CR1 = cr1 | (datasize & SPI_CR1_DFF) | en;
 }
````
According to the reference manual RM0008 (Rev. 16), chapter 25.5.1, page 745, the data format bit DFF should be written only when SPI is disabled (SPE=0) for correct operation. And this is exactly what the original version was missing, what I have implemented. I don't know whether it worked correctly before and why...

Now, the ILI9341 lib is using this function very intensively. I am not sure whether this is necessary and whether it can be optimized, but I can imagine that the run-time difference is only due to this change.

As test, you could try to change locally this function back to original version and check the run-times again.

Meanwhile I will have a look on the ILI94341 lib to see what is going on there.

stevstrong:
Having a look in ILI94341 lib, it seems that 8 bit mode and 16 bit mode are alternatively used. 8 bit mode is used to write the commands, 16 bit mode used to write (color) data.

I will try to rewrite the lib to use exclusively 16 bit mode, meaning writing the command in 16 bit mode, too, so that no mode switch is required. This will, on one side, increase the run-time due to added extra byte (command hi_byte, which is actually 0), but will, on the other side reduce the run-time because no mode switch is required anymore.
I am curious to see which one has the greatest influence on the run-time.

rogerclarkmelbourne:
@stevstrong 

I don't mean to make more work for you. But I know that some people e.g. the guys using the pig-o-scope will not be happy if I post an update which slows down the performance of SPI by 10% even if it fixes bugs

stevstrong:
I managed to test it on an SPI display, and the increase is indeed because of the changed setDataSize() function. I got similar values as yours.
OTOH, interestingly, the old version works, too, without disabling before the SPI interface.
So I can actually revert this change. Would the PR then be acceptable?

rogerclarkmelbourne:
@stevstrong 

I think if there is a not a big reduction in speed it would be OK.

I know some speed reduction was necessary to fix bugs, but 10% ore more speed reduction I think would cause people to question if the bugs needed to be fixed

stevstrong:
Roger, I have reworked the Adafruit library for 16 bit SPI register access.
As a positive side effect, the benchmark is now better, even with the setDataSize() bugfix, because there are significantly less calls to that function now than before. Only the setRotation() function needs it still.

ili9341 benchmark

I also tested the ILI9341 lib with the original GFX library instead of GFX_AS from the repo and I mostly achieved better bench results, so I leaved this way and used as further reference. It can be changed back if not wanted.

I hope this speed improvement finally gives green light for the commit.

rogerclarkmelbourne:
@stevstrong 

Thanks for all your hard work.

I'll retest later today.

rogerclarkmelbourne:
BTW. I'm now starting to use the github "Releases" system, which gives me more flexibility to update the Master HEAD of the repo with new code for experienced users to test, while still not upsetting people who were happy with what they had. (see the announcement on the forum)

This will make it more acceptable (to users) for me to accept PR's

stevstrong:
Makes sense.

rogerclarkmelbourne:
@stevstrong 

I pulled this to my local repo, but it now fails to compile the graphic test

The errors seem to be because of multiple definitions of 

multiple definition of 
Adafruit_GFX::drawLine(short, short, short, short, unsigned short)
Adafruit_GFX::drawRect(short, short, short, short, unsigned short)
Adafruit_GFX::drawFastVLine(short, short, short, unsigned short)'
etc

in Adafruit_GFX.cpp

I can get rid of the error by not including

#include <Adafruit_GFX_AS.h>    // Core graphics library, with extra fonts.

in the example, but I think its normal to include that file

So I suspect perhaps that file needs to be updated with the changes you made to the ILI9341_STM lib ??

rogerclarkmelbourne:
Umm.
I just realised that I have 2 copies of the Adafruit_GFX library, so perhaps our copy works and the official copy doesn't. I'll need to delete the official copy and try again, but I don't think it will fix things.

stevstrong:
The problem is that if you have already included into the sketch the GFX_AS header, the ILI9341 header includes now the GFX header, so you have two GFX libs.

The solution is to either remove the GFX_AS header include from the sketch, or to revert my change where I included the GFX header instead of GFX_AS header.

I think I will revert my change so that noone else will complain :).

rogerclarkmelbourne:
OK.

I think the example sketches include GFX_AS

Its getting a bit too late to pull and test today (sorry its after 9pm and I've already spent 3 hours on other changes)

But I'll look first thing tomorrow.

stevstrong:
No hurry, and, more important, DON`T PANIC!
:o)

rogerclarkmelbourne:
Hi Steve

I've pulled you code again and run the graph test, and it now seems to run OK

All your SPI code seems to run faster than the old code

Activity | time for master branch | time for new PR code | Time taken by PR code as a fraction of Master time
------------ | ------------- | ------------- | -------------
Screen fill|170786|170777|0.999947302
Text|43086|39766|0.922944808
Lines|242465|223304|0.920974161
Horiz/Vert Lines|15811|15624|0.988172791
Rectangles (outline)|11569|11293|0.976143141
Rectangles (filled) |355069|354995|0.99979159
Circles (filled)|143286|133995|0.935157657
Circles (outline)|172906|157524|0.911038368
Triangles (outline)|59728|54880|0.918832039
Triangles (filled)|166040|161165|0.970639605
Rounded rects (outline)|59772|55095|0.92175266
Rounded rects (filled)|415712|412647|0.992627107

So I'm going to merge it into the master branch, and then leave it for the community to test ;-)

Thanks again

PR #292, opened 2017-05-23
wait few seconds (3) after dfu upload, tools/{os}/maple_upload shell …

ag88:
…script, to prevent arduino ide's serial console connecting prematurely

discussed in this thread
http://www.stm32duino.com/viewtopic.php?f=28&t=2107
fixes contributed by hanyazou, danieleff, ag123

### DIFF
--- a/tools/linux/maple_upload
+++ b/tools/linux/maple_upload
@@ -37,4 +37,13 @@ if [ ! -x "${DFU_UTIL}" ]; then
     exit 2
 fi
 
+echo "${DFU_UTIL}" -d ${usbID} -a ${altID} -D ${binfile} ${dfuse_addr} -R
 "${DFU_UTIL}" -d ${usbID} -a ${altID} -D ${binfile} ${dfuse_addr} -R
+
+# take a breath waiting for restarting the target device //hanyazou, ag123
+echo -n wait for ${dummy_port_fullpath}...
+while [ ! -c ${dummy_port_fullpath} ]; do
+    sleep 0.1
+done
+sleep 0.3
+echo done.
--- a/tools/linux64/maple_upload
+++ b/tools/linux64/maple_upload
@@ -37,4 +37,13 @@ if [ ! -x "${DFU_UTIL}" ]; then
     exit 2
 fi
 
+echo "${DFU_UTIL}" -d ${usbID} -a ${altID} -D ${binfile} ${dfuse_addr} -R
 "${DFU_UTIL}" -d ${usbID} -a ${altID} -D ${binfile} ${dfuse_addr} -R
+
+# take a breath waiting for restarting the target device //hanyazou, ag123
+echo -n wait for ${dummy_port_fullpath}...
+while [ ! -c ${dummy_port_fullpath} ]; do
+    sleep 0.1
+done
+sleep 0.3
+echo done.
--- a/tools/macosx/maple_upload
+++ b/tools/macosx/maple_upload
@@ -50,4 +50,13 @@ if [ ! -x ${DFU_UTIL} ]; then
     exit 2
 fi
 
+echo ${DFU_UTIL} -d ${usbID} -a ${altID} -D ${binfile} -R ${dfuse_addr} -R
 ${DFU_UTIL} -d ${usbID} -a ${altID} -D ${binfile} -R ${dfuse_addr} -R
+
+# take a breath waiting for restarting the target device //hanyazou 
+echo -n wait for ${dummy_port_fullpath}...
+while [ ! -c ${dummy_port_fullpath} ]; do
+    sleep 0.1
+done
+sleep 0.3
+echo done.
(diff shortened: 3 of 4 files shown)


rogerclarkmelbourne:
This PR seems to add delays even when they are not needed by the user.

I'm afraid with boards like the Maple Mini and BP which only have 1 MCU which has to handle both the upload (via its bootloader) and the main application, that some people will have problems on their specific machines.

However, the vast majority of people do not have problems, and would object to delays being inserted in the upload process,

I'd advise anyone who requires a delay for there specific set-up to work to edit the upload file and add a delay that suits their on system.

Note. This is a long standing issue on OSX which seems to take ages to enumerate USB on some models, especially where people connect via the keyboard / hub

I recall adding some optional delays in the OSX upload scripts, but these were optional and the user has to enable (and tune the delay) to suit their system.

ag88:
i'm thinking for windows, it might be possible to shorten the delay to 1s (i.e. a single ping)
the main trouble it seemed is that the host (i.e. arduino ide serial console) turns around faster than do the target (maple mini / blue pill) have enough time to complete the reset and initialise the serial device (i.e. enumerate)

the end result is that arduino ide serial console tries to connect even before the serial device has enumeriated and ends up throwing and error as the device does not exist
http://www.stm32duino.com/viewtopic.php?f=28&t=2107#p28433

this really is a race condition in terms of the serial console connecting to the usb device and the blue pill/maple mini trying to complete the reset, boot and enumerate usb.
this delay is necessary to provide just enough time for blue pill / maple mini to complete the reset, boot, and enumerate usb so that a valid device shows up on the host
so that once the arduino ide serial console tries to connect, blue pill / maple mini has enumerated and the usb port is valid for the arduino ide serial console to connect. i think this avert many difficult / marginal situations for the turn around between flashing the sketch to reset and enumerating usb

rogerclarkmelbourne:
The problem with the Serial Console getting messed up if you leave it open, has been there from the start., and this issue seems to only niggle one or two people. 

Adding a small delay will only work on fast machines. People on old slower machines would need more delay, and some people may need no delay at all.

Probably a better solution is a system which waits and checks for the serial device to become available before the upload script returns, but on Windows changing it is a problem as it uses a JAR to which we don't have the sources.

danieleff:
I added the following modified version to my repo, and they seem to work. Tested on linux, win7, win10. (I actually also added it to stlink upload too)
These wait for serial, or timeout in a 4 seconds.
Now the serial console can be open at all times.

Linux:
```
#!/bin/bash

echo -n Waiting for ${dummy_port_fullpath} serial...

COUNTER=0
while [ ! -c ${dummy_port_fullpath} ] && ((COUNTER++ < 40)); do
    sleep 0.1
done

echo Done

```

Win:
```echo Waiting for %1 serial...

for /l %%x in (1, 1, 40) do (
  ping -w 100 -n 1 192.0.2.1 > nul
  mode %1 > nul
  if ERRORLEVEL 0 goto comPortFound
)

echo timeout waiting for %1 serial

:comPortFound
```

rogerclarkmelbourne:
Thanks Daniel

It seems like a good solution, as its not dependant on a fixed timeout.

I presume the Mac version would look very similar to the Linux version.

I will try it later

rogerclarkmelbourne:
Umm.

Almost works perfectly but not quite

```
int c=0;
void setup() {
Serial.println("Hello world");
}
void loop() {
delay(250);
Serial.println(c++);
}
```

gives

2
3
4
5
6
7
8
9
10
etc

ie it doesn't show

Hello world 
0
1

But is a lot better than we had before

RickKimball:
I have an issue sometimes that when my bluepill re-enumerates, it comes back as a different tty .. This changes seems like it would cause a hang on my linux setup and bluepill.

rogerclarkmelbourne:
Doesn't daniel's code have a timeout  or 40 x 0.1 secs, so worst case is 4 seconds

We could probably reduce that to 2 or 3 secs

RickKimball:
I was looking at the linux pull code not the comments.

rogerclarkmelbourne:
@RickKimball 

@danieleff's solution in the comments looks the best.

I would not use @ag88 's version as I don't like fixed timeouts.

ag88:
i think i'm seeing rick's point about it, but i'd think daniel's code could avoid a possible 'hang', yup i think we could use 2 secs as timeout instead of 4

rogerclarkmelbourne:
Whats the current status of this.?

ag88:
hi roger, my bad, it has been a while and i couldn't really remember what i changed,
examining 3 'files changed', i noted i added some echo statements for the dfu-util command lines in the linux and mac os scripts. that basically echos the dfu-util command line that's run
i'd guess that may help provide some visible feedback to the user should there be issues

 it seemed the rest of the commits are merged and do not show any differences

rogerclarkmelbourne:
I'm thinking of enabling the "hold in bootloader" feature where the core puts a magic number in one of the battery backup registers (which survive a reboot), but I don't know if this would effect this PR

PS. The code is already in the bootloader, I just need to add it to the core as part of the reboot for upload, code

PR #294, opened 2017-05-25
Fix documentation for usart_putc()

tomtor:
See discussion at https://github.com/rogerclarkmelbourne/Arduino_STM32/pull/260

usart_putc() blocks until space is available in the ring buffer, not until the character has been transmitted.


rogerclarkmelbourne:
Thanks.
BTW. I didnt realise initially this was a documentation fix.  I've have actioned it sooner if I had realised ;-)

PR #298, opened 2017-06-02
Avoid initializing USB serial twice

hanyazou:
This patch avoid initializing the USB serial HW of F103C twice in the bootstrapping process. It may hangup if you initialize the HW while the USB serial port is using by host.

Please refer to discussion in the topic " USB Observations" at  http://stm32duino.com/viewtopic.php?f=28&t=2107

There is another patch which will fix this problem also.
http://stm32duino.com/viewtopic.php?f=28&t=2107&start=80#p29087

### DIFF
--- a/STM32F1/cores/maple/usb_serial.cpp
+++ b/STM32F1/cores/maple/usb_serial.cpp
@@ -55,6 +55,7 @@ static void ifaceSetupHook(unsigned, void*);
 
 #define USB_TIMEOUT 50
 
+bool USBSerial::_hasBegun = false;
 USBSerial::USBSerial(void) {
 #if !BOARD_HAVE_SERIALUSB
     ASSERT(0);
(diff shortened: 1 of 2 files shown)


rogerclarkmelbourne:
Looks like a good, and simple, solution

BTW. I though you also had a much more complicated solution involving a major modification to the whole of the startup / init system, but I prefer this solution in the short term as it is low risk and should fix the issue

rogerclarkmelbourne:
I tested on my Maple mini and it fixed the issue, so I've merged your PR into the master.

BTW. Strangle I did not get the problem if I used Serial.begin(115200); rather than Serial.begin();

So I looked at the code and there is a bug (in what I did a long time ago), because the Serial.begin(xxx) and Serial.begin(xxx,yyy) functions don't ultimately call Serial.begin()

```
//Roger Clark. Two new begin functions has been added so that normal Arduino Sketches that use Serial.begin(xxx) will compile.
void USBSerial::begin(unsigned long ignoreBaud) 
{
volatile unsigned long removeCompilerWarningsIgnoreBaud=ignoreBaud;

	ignoreBaud=removeCompilerWarningsIgnoreBaud;
}
void USBSerial::begin(unsigned long ignoreBaud, uint8_t ignore)
{
volatile unsigned long removeCompilerWarningsIgnoreBaud=ignoreBaud;
volatile uint8_t removeCompilerWarningsIgnore=ignore;

	ignoreBaud=removeCompilerWarningsIgnoreBaud;
	ignore=removeCompilerWarningsIgnore;
}
```

So I think they need to be updated to call ::begin();

PR #302, opened 2017-06-27
Adding i2C repeated start condition

bmarquismarkail:


rogerclarkmelbourne:
Thanks..

Whats the best way to test this ??

bmarquismarkail:
The best way to test this is to utilize a device with an i2c interface that both supports and does not support repeated stop.

I used an NXP MPL3115A2 to test the repeated start functionality, and a Silicon Labs Si5351A to test its legacy mode ( endTransmission(void) ).

They both can be bought as modules from adafruit:
https://www.adafruit.com/product/2045
https://www.adafruit.com/product/1893

rogerclarkmelbourne:
umm
I don't have those.

I guess that as long as it works with the same devices it used to work with, e.g. BMP280 then it would be OK for me to accept the PR and if someone raises an issue because the repeated start doesn't work for them, we'd have to come back to it and bug fix

bmarquismarkail:
I will be more than happy to!

PR #303, opened 2017-07-04
Update boards.txt

Testato:
- Exluded CCM Ram


victorpv:
Thanks Testato, I think it makes more sense like that.
On another note, there is a discussion going on in the forum about the CCM, in case you want to discuss that subject there.
There are some options to use CCM. It has some limitations (no DMA and no code), so I'm not sure we will all come to an agreement about it, but perhaps we can find a way to conditionally use it depending on options or something.

rogerclarkmelbourne:
There is now a merge conflict with @stevstrong latest changes to F4 boards.txt, can someone investigate?

Testato:
Solved

rogerclarkmelbourne:
Thanks

PR #305, opened 2017-07-05
update SDIO - fix for CL4 cards

stevstrong:
CL4 cards are slower, therefore the timeout for DMA end had to be increased.


rogerclarkmelbourne:
Was this intended to be changes to the F4 only  ?

I'm seeing some changes in the F1 USB as well as the F4 changes

stevstrong:
Roger, sorry, is again my lack of github knowledge, I hoped to create a small PR for only this change, and this is what it became.
Anyway, the F1 USB changes relate to your fix for PB10 (USB DISC PIN, https://github.com/rogerclarkmelbourne/Arduino_STM32/commit/f7a576f2e0efcee958bdfd737d9af7a2f6c76051). The compiler generates a warning when NULL is passed for the second parameter which is expected to be uin8_t.

In addition, I removed the compiler warning "USB CDC ACM relies on LeafLabs board-specific configuration.", it was just pointless, we are using the USB CDC anyway on several boards.

The rest of the changes relate to F4.

rogerclarkmelbourne:
Steve

Don't worry.

I'll manually do the F1 changes and commit them myself.

rogerclarkmelbourne:
I've manually updated to fix the issues with USB Serial, but at the moment it seems to think your F1 changes are still needed.

Perhaps you can try pulling the repo to your repo and see if those changes then go away

stevstrong:
I'll check it later today.

stevstrong:
Just updated freshly from your master branch.

rogerclarkmelbourne:
Steve

I've locally merged this.

I didn't do it manually, because it only seemed to change the F4 files as far as I could tell

PR #307, opened 2017-07-08
Spi callback

victorpv:
Added SPI Callback functionality to the SPI library as discussed in the forum thread.
Tested functionality in SPI1 and SPI2


victorpv:
Cancelling this one.

PR #308, opened 2017-07-08
Added USBSerial implementation of readBytes.

victorpv:
As discussed in thread http://stm32duino.com/viewtopic.php?f=53&t=2299&p=31008#p31008
The Stream class has a readBytes function to read blocks, but it implements so by repeatedly reading 1 byte a time with calls to read().
This change adds a faster implementation in the USBSerial class, increasing speed for large reads from 300KB/s to 500KB/s max.


rogerclarkmelbourne:
Thanks @victorpv

HexTank:
This code might be more performant if the millis() call is avoided if data has been read.

It also highlights that data has to be transferred in the _timeout phase, regardless if data is being received or not.

HexTank:
Is there any reason the types differ as well? char* rather than uint8, and why is 'len' a reference?

victorpv:
I can't remember the details, but the coffee was mostly copied from either the Arduino or the teensy implementation, most likely the types are the same.

victorpv:
Feel free to improve it, pjrc posted a tool for Windows he wrote and used to test the speed, I used the same tool for testing

HexTank:
Ah cool. Do you have a link where I can find the tool? Cheers.

stevstrong:
I think this is the site @victorpv was referring to: https://www.pjrc.com/teensy/benchmark_usb_serial_receive.html
There you can find a package including the used host software.

HexTank:
I had a play with it, on Windows at least, the numbers fluctuated enough to not be able to measure enough of a gain to be considered a stable performance increase, so I'll leave it as interesting research session. :)

PR #312, opened 2017-07-10
Spi callbacks

victorpv:
This has the addition of:
dmaSendSet
dmaSendRepeat
dmaTransferSet
dmaTransferRepeat
onTransmit
onReceive
EventCallback

I have added a few variables to the class to hold the "this" pointer for each port, so they should work fine with all 3 ports (I have tested SPI1 and SPI2 simultaneously, with different callback function for each).
It should also work with setModule, since the "this" pointer is in the settings class.

Changed TABs to spaces, previously was a mix of both, and I believe the preference is for spaces. If that's not the case I will resubmit with Tabs.

Finally, the only change to any other function is in dmaSendAsync, where I changed the variable used to hold whether there is a DMA transfer going on, so if dmaSend is used with callbacks (non-blocking) and you try to send with sendAsync, sendAsync will wait for the previous dma to be finished.
Thinking on it we should probably add the same check to the other functions.

### DIFF
--- a/STM32F1/libraries/SPI/src/SPI.cpp
+++ b/STM32F1/libraries/SPI/src/SPI.cpp
@@ -101,16 +101,19 @@ SPIClass::SPIClass(uint32 spi_num) {
 #if BOARD_NR_SPI >= 1
     case 1:
         _currentSetting->spi_d = SPI1;
+		_spi1_this = (void*) this;						  
         break;
 #endif
 #if BOARD_NR_SPI >= 2
     case 2:
         _currentSetting->spi_d = SPI2;
+        _spi2_this = (void*) this;								  
         break;
 #endif
 #if BOARD_NR_SPI >= 3
     case 3:
         _currentSetting->spi_d = SPI3;
+        _spi3_this = (void*) this;								  
         break;
 #endif
     default:
@@ -137,6 +140,8 @@ SPIClass::SPIClass(uint32 spi_num) {
 	_settings[2].spiRxDmaChannel = DMA_CH1;
 #endif
 
+    // added for DMA callbacks.
+    _currentSetting->state = SPI_STATE_IDLE;
 }
 
 /*
@@ -154,6 +159,8 @@ void SPIClass::begin(void) {
     spi_init(_currentSetting->spi_d);
     configure_gpios(_currentSetting->spi_d, 1);
     updateSettings();
+    // added for DMA callbacks.
+    _currentSetting->state = SPI_STATE_READY;
 }
 
 void SPIClass::beginSlave(void) {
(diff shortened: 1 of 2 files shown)


victorpv:
Furthermore, we could slightly modify the dmaSendAsync function to set the DMA ISR to EventCallback too.
That way when the DMA is over EventCallback will change the state variable to Ready (and call a user callback if one is set), and dmaSendAsync doesn't need to wait checking when the transfer is over, that will remove overhead in that function.

rogerclarkmelbourne:
I've tested these changes with the graphics test sketch and it seems to still work fine

So I will merge the changes

stevstrong:
Sorry to resurrect this but I am trying to port this code to F4 and have some issues when using dmaSendAsync() and dmaSend() in the same application, without specifying any callback for transfer/send.
I just wonder whether we could easy the application part if dmaSendAsync() could have an input parameter a variable which would be set when the transfer/send is done. This way the application could just poll that variable and do what needed in that case.
Otherwise the application should have its own function which sets that variable when the send/transfer is finished. More than that I cannot imagine that is meaningful to do in the attached event as callback function.
Or do you have some other concerns regarding this, Victor?

victorpv:
I added a volatile in the code,  _state, that we can leverage for that.
I see multiple easy solutions:
1.- Add a new function to return whether the port is active or not. I need to check if the DMA ISR is always set, or only when there is a callback set. If the later, we can change it to the former. So the SPI DMA ISR can change that _state variable if is not doing so already.
2.- Add a new parameter to dmaSendAsync. That poses a risk of breaking existing code, so I kind of lean towards adding a new function that returns the state (states can be idle, ready-after begin() is called, TX when doing a TX, and RX when doing an RX). the function could also just return a bool, with true for TX or RX states, and false for anything else.
3.- Implement a callback function that controls a variable within the user sketch.

Roger uses the dmaSendAsync in a led driver library, so if we make any change to it, we need to test with that. There may be other people using the library or other code using the same function. As long as we can keep the code compatible with the existing calls, I don't see a reason to not make changes.

stevstrong:
OK, 1 or 3 should be doable.
My problem during evaluating the live OV7670 project was that I had to call tft.setAddressWindow() without knowing the state of the async DMA called before, so it was sometimes conflicting with that.
Accordingly, an additional function which would return the state of the DMA (finished or not) would be really helpful. This would involve the existence of the SPI driver internal callback which sets the internal _state, right?

victorpv:
Right. There is an internal callback within the SPI driver, but I don't remember if it's enabled or only when the user code sets a callback, I'll check that up later today and make the necessary changes.

stevstrong:
OK, thaks.
In the same context, I think there is a potential problem with using dmaSendAsync() and dmaSend().
Even having defined a user callback, the dmaSendSet() will be executed regardless of internal _state, which I think is not wanted.
In my test version I made a separate function `dmaSendWaitCompletion()` which waits the internal _state to get ready, and I call this function at the beginning of each DMA send function, even for dmaSendSet(). This involves a small overhead, but I think this eliminates any conflicting situation, so it maybe makes sense.  What do you think?

victorpv:
I think if we just checked the _state value instead should reduce the overhead from the calls, and still not take much more space.
But either way, I agree it should be guarded.

PR #314, opened 2017-07-19
Fixed headings

hemalchevli:
Readme now look much cleaner.


rogerclarkmelbourne:
Thanks

I'd noticed that problem but did not have time to fix it

PR #315, opened 2017-07-19
Reworked RTC for F4

stevstrong:
- no need for extra time library.
I took code snippets from Teensy's Time lib.

- as discussed in the forum: http://www.stm32duino.com/viewtopic.php?f=39&t=775&p=31218#p31218


rogerclarkmelbourne:
Thanks..

PR #317, opened 2017-07-28
F103C8 128kB flash upload

Testato:
Added F103C8 128kB flash upload, functionality largely accepted as working and present on the wiki.

### DIFF
--- a/STM32F1/boards.txt
+++ b/STM32F1/boards.txt
@@ -178,12 +178,19 @@ genericSTM32F103C.menu.device_variant.STM32F103C8.build.ldscript=ld/jtag_c8.ld
 genericSTM32F103C.menu.device_variant.STM32F103C8.upload.maximum_size=65536
 genericSTM32F103C.menu.device_variant.STM32F103C8.upload.maximum_data_size=20480
 
+## STM32F103C8 -------------------------
+genericSTM32F103C.menu.device_variant.STM32F103C8x=STM32F103C8 (20k RAM. 128k Flash)
+genericSTM32F103C.menu.device_variant.STM32F103C8x.build.cpu_flags=-DMCU_STM32F103C8
+genericSTM32F103C.menu.device_variant.STM32F103C8x.build.ldscript=ld/jtag.ld
+genericSTM32F103C.menu.device_variant.STM32F103C8x.upload.maximum_size=131072
+genericSTM32F103C.menu.device_variant.STM32F103C8x.upload.maximum_data_size=20480
+
 ## STM32F103CB -------------------------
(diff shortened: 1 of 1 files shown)


rogerclarkmelbourne:
We already have a menu option for the 128k F103CB

Although currently most STM32F103C8 boards seem to have 128k, this is not guaranteed to continue, as STM may change their manufacturing process and disable any Flash beyond the spec of the board, so I don't think adding a C8x would be appropriate.

If people want to use 128k of the flash they can just select the 103CB variant

Testato:
Yes I agree that ST may change it's production, but all the actual and future C8 cpu will remain in the field.
Also the bootloader message on this C8 cpu officially speaks about 128KiB flash space, so i think: "Why not ?"

In the future, if ST will really change their manufacturing, the people will see in the bootloader message that it have a 64KiB flash and will choose 64KiB board.txt. But nobody will remove from the field the thousand/million of "Cx" version.

It is more user friendly and clear for the people found two C8 line in the board menu, because normally he never choose CB board without know that he can do.
Also i do not know if by choosing the CB board is correct, because i see that the cpu_flag is different, and maybe can be something of thifferent in the core, i do not know, but it is more correct use the correct cpu_flag: cpu_flags=-DMCU_STM32F103CB vs -DMCU_STM32F103C8

I committed a new name for the board so it is more clear that this is a particular case:
`STM32F103C8 Experimental (20k RAM. 128k Flash)`

rogerclarkmelbourne:
I saw you'd fixed that typo where a 8 should be a B or vice versa

rogerclarkmelbourne:
I'll create a topic on the forum to discuss this PR (Like I do for the other PRs)

Testato:
Yes, it is the best way.
Please put also here the link to the topic

rogerclarkmelbourne:
I've committed a change to fix the typo in the RAM size on the F103CB.

I've also created a topic to discuss this change on the forum

http://www.stm32duino.com/viewtopic.php?f=53&t=2402

Please update your fork of my repo as there have been a lot of changes to board.txt for other reasons

victorpv:
Honestly I think this should not be merged for 2 reasons:
1.- Not all C8 MCUs have 128KB, and the ones that have are virtually a CB sold as a C8, so should be programmed as CB (otherwise things like end of flash in the EEPROM library would not be correct).
2.-Newbies not knowing what's about may buy a C8 board, then seeing the option for 128KB try to use it, and fail, because their C8 did not have 128KB, so would just cause confusion to people that doesn't now about the subject. For anyone knowing about this, better use the CB as explained in point 1.

rogerclarkmelbourne:
I agree.

If anyone wants to try using 128k on a 64k device they can just pick the F103CB instead of the C8

PR #319, opened 2017-07-30
Add openocd upload( J-Link OB) and STLink V2 type for 103C8, 103V

giveup2001:
Signed-off-by: wang <w.limin@qq.com>


rogerclarkmelbourne:
This PR seems to make changes that do not relate to openocd 

Also, we already have STLink upload, so I'm not sure why another type of STLink upload is needed

Please write in more detail precisely what this PR aims to achieve

PR #322, opened 2017-08-01
linux fix for gd32 variant in boards.txt

mtiutiu:
Linux is case sensitive so it matters when it comes to files/folder names. This fixes GD32F103xx variants in boards.txt to pick the right folder at compile time.

Otherwise this happens:
```
...
#include <board/board.h>
                         ^
compilation terminated.
exit status 1
Error compiling for board Generic GD32F103C series.
```


rogerclarkmelbourne:
Thanks

PR #325, opened 2017-08-03
Correcting FreeRTOS900 examples to include right version

victorpv:
The examples had not been edited to use the FreeRTOS 9.0 version, so correcting that now.


rogerclarkmelbourne:
Thanks

PR #327, opened 2017-08-18
Adding new.cpp as in F1 core.

victorpv:
This is to avoid libstdc++ being pulled in when new() is used.
Already included in the F1 in 2016.
Discussed on this thread:
http://www.stm32duino.com/viewtopic.php?f=39&t=2460&p=33163

### DIFF
--- a/STM32F4/cores/maple/new.cpp
+++ b/STM32F4/cores/maple/new.cpp
@@ -1,0 +1,35 @@
+/*
+  Copyright (c) 2014 Arduino.  All right reserved.
+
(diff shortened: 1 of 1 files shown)


rogerclarkmelbourne:
Thanks.  Adding new.cpp to the F1 did not seem to cause any problems, so I'm happy to merge this PR

victorpv:
Thanks. Should be easy to revert in case it does, being a single file, but I think we are safe. Both me and Frederic compiled with it and seemed to work fine.

Issue #332, opened 2017-08-28
Add a new RTC example


csnol:
[https://github.com/csnol/STM32-Examples](url)

It's a part of my internet clock without the codes about LCD and ESP8266.

If useful, my pleasure!

rogerclarkmelbourne:
Thanks

BTW. The link in your issue, did not work, but found it was 

https://github.com/csnol/STM32-Examples

I'll see if I have time to add it to the other examples (as we have a folder for examples)

RickKimball:
Can we let this thing have some flight time before blindly adding it to the core libraries?

csnol:
Hi @rogerclarkmelbourne  and @RickKimball 
I just wanna add the Timelib library to STM32F1 system. 
because the RTClock.h hasn't  include year(), month() day() hour() minute() and second() function in the STM32F1 of this repo. (I don't know why them be included in  RTClock.h in  STM32F4)

On the other hand,  I can't sure whether  or not its license  can be used in this Arduino_STM32 repo.
Its  license is  GNU Lesser General Public License,  NOT MIT of Arduino_STM32

```
  time.c - low level time and date functions
  Copyright (c) Michael Margolis 2009-2014

  This library is free software; you can redistribute it and/or
  modify it under the terms of the GNU Lesser General Public
  License as published by the Free Software Foundation; either
  version 2.1 of the License, or (at your option) any later version.

  This library is distributed in the hope that it will be useful,
  but WITHOUT ANY WARRANTY; without even the implied warranty of
  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
  Lesser General Public License for more details.

  You should have received a copy of the GNU Lesser General Public
  License along with this library; if not, write to the Free Software
  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
  ```
I really don't understand about these freeware license.

Thanks guys!
BR.
C.S Nol.

stevstrong:
You may have a look to the RTC lib for F4, I have reworked it currently to be independent from any external time lib: https://github.com/rogerclarkmelbourne/Arduino_STM32/tree/master/STM32F4/libraries/RTClock/src

csnol:
Thanks @stevstrong .
I've gotten it. very nice codes . and I'll try to copy some lines for F103.

csnol:
I've renew the RTColck library, Don't dependents external time lib anymore.
[...]

PR #334, opened 2017-08-29
Update F103 RTClock library for RTC sketch.

csnol:
Add. time and date function.


victorpv:
Seems like you added quite a bit of code, and features.
Did you see the other PR from Steve related to that library?
https://github.com/rogerclarkmelbourne/Arduino_STM32/pull/333

stevstrong:
Victor, my PR is only for F4, not relevant here because of different alarm structure.

stevstrong:
I would give green light for this PR.

csnol:
@victorpv  
Yes， in fact, many codes of this library porting from @stevstrong 's library for F4.
F103 still is a popular chip,  and RTC is important function. there will always people  update this library.  So I do.  just wanna do something for this repo. wanna share good codes. that's all.
and the story about this RTC library, you can get from [#332](https://github.com/rogerclarkmelbourne/Arduino_STM32/issues/332)

finally, Thanks @stevstrong again.   It is really useful without "timelib.h".  I've changed my clock from Pro mini to F103, and removed a DS1302 RTC chip.

victorpv:
Thanks for the clarification to both, I only wanted to refer one another since I thought you were both working on the same RTC files.
I don't have anything to object to the changes on either PR.

pkourany:
Roger, this PR looks good.  Can you please merge it!

RickKimball:
Has this been tested on different stm32f103 boards?

rogerclarkmelbourne:
I don't think anyone except the OP has tested it.

It can't be merged until its been independently tested by a few people who have the necessary hardware setup

pkourany:
Did some testing last night.  All the examples work but when I try both forms of `createAlarm()`, I could not get the alarm interrupt to fire.  I didn't get to debugging.

csnol:
@pkourany  thanks for your test. 
createAlarm() is useful to wake up MCU from halt/sleep.
I've added 2 functions,  could you please help to test for me?  
I can't test them now the boards not in my hand.

csnol:
@rogerclarkmelbourne  and @RickKimball 

I don't change the core library rtc-util in /util/,   The programmer need totally understanding  the features of F10X.
you know RTClock just is a repackaged C++ library of rtc-util., and I only added some date function and modified errors into the RTClock.  
It is very easy to build for C++ coders even like me a STM32 beginner.
and I searched STM32F10X series in cubeMX,  found them have a same feature on RCT function.

You are right,     **No tested No truth.** 
maybe it is a good idea that we can wait somedays for test it on different F10X boards.

pkourany:
I was able to test an alarm using createAlarm() and it works find.  I will be testing on wake from sleep and will also test new added functions.  All tests done on custom STM32F103RCT6 board with 32KHz crystal for RTC and processor running at 8MHz (HSI).

I wonder if adding a function to set the timezone for the time calculations is worth including?

pkourany:
Tested createAlarm() to wake from STANDBY and STOP and both worked.  Also, new functions work as expected.  For me this PR is a go though still would like comments on timezone idea.

csnol:
@pkourany  Thanks.
About timezone, up to now I've no idea.  because of the DST.  The start/stop date of DST be affirmed by governments.   maybe someday you find a hour early/late from your clock.    

``` c++
#ifndef TimeZone
#define TimeZone  0
#endif
#ifdef  DST
#define DSTstart_month  4
#define DSTstart_day  1
#define DSTstop_month  10
#define DSTstop_day  1
#endif
```
Look, that's why I don't wanna include it.    it make the  library more complicated.  
I think it better choice that Unix epoch time as the standard and write some codes about timezone / DST in sketch.

pkourany:
@csnol, I was thinking more along the lines that you have a function to set the timezone (user calculated) and it is used internally for time calculations.  A convenience more than anything.  As for DST, that's another headache altogether!

csnol:
Haaaa, I haven't headache. No DST in China.   but I can try to add timezone in RTClock.

pkourany:
@csnol, I haven't tested the TimeZone function but it is very simple.  I am satisfied that this library is good to go.

csnol:
All new function tested on BluePill. and BlackPill.
Everyone can download BluePill_RTC_test.ino to test your boards
Change LED_PIN and timezone to yours.

RickKimball:
@pkourany, @rogerclarkmelbourne  I don't understand this urgent need to force libraries into the core stm32 code. This code is obviously fresh and only lightly tested by a handful of users.  It seems to me that libraries that are included in the core should have been tested by more than 2 people and should have some time as a standalone usage before being considered for the core.

stevstrong:
People interested in this lib, well they did test it and they want to have it a.s.a.p. in repo.
Those who don't use it, and are not willing to test it, should at least not delay the merging process.
Anyway, in case of any further issue, if they come up, can be easily handled, so **why not merge it**?

pkourany:
@RickKimball, this PR is "safe" in that it doesn't break existing functionality.  What's the point of having PRs if they don't get merged!  If we wait for this code to be tested by many, it may never be merged.  I am using this code in concert with low power modes (not a library at this time) and it is working very well.  @rogerclarkmelbourne can't do everything :)

csnol:
Easy guys. Rigorous is right. and Enjoy sharing right too.
they must find it who interesting on RTC of STM32. so it isn't important if merge this PR.

RickKimball:
I think that code included in the core should be of high quality and tested with all the boards and not just looked at and deemed "safe".  I don't think testing with two boards by two people is enough warrant inclusion.

Each new library we add to the core is giving that library a sort of stm32duino blessing. There is an implicit suggestion that it is of high quality and tested with all boards.  When these core library have issues, they cause support questions that will just suck up Roger's time and could be better handled by the original author.  When issues are fixed, that means Roger is again in the middle being impatiently pushed by people to update the library in the core.  The original author can't just post a fix in his github, he has to get Roger involved. 

Just looking at the some of sample code included, it doesn't even use LED_BUILTIN. How is that high quality code?  Granted, I spent five seconds looking at this code. However, it seems like it needs more time to be tested and used on a variety of boards.

pkourany:
@RickKimball, the library is for the F103 only.  It is difficult to imagine that the library would not work on any F103 board.  @csnol has tested on the two most common F103 platforms and I have tested on an F103 custom variant.  Nonetheless, your comments regarding the sample code is good.  However, @csnol expanded an existing library to add functionality so it stands to reason that he shouldn't reinvent the examples, only add new ones.  It would be good to have a new example that demonstrates the new features for sure.  Can you contribute to the library testing and getting a solid example together?

csnol:
@RickKimball 
Coding really is a philosophy,  Thanks for your 5 seconds..  I can't say anymore.  Good-night!

RickKimball:
OK I checked out the pull request... First thing I notice ... BlutPill-RTClock-test ...

maybe call it Blue?

victorpv:
I think both opinions have merit. We can't hold PRs waiting forever if no one else is willing to test, since then the core won't ever advance or get bugfixes, and at the same time code added to the core needs to be well tested so we don't break something.
Since @csnol sent the PR, a couple of weeks have passed, but now seems to be getting the attention it deserved. Since there are comments about adding timezone and what not (in which I agree with csnol that should be left out and leave the library in UTC, the user can manage timezones if he needs to), perhaps it should just stay open a few more days until the discussions are finished, so we don't make Roger merge, and then merge again in a few days, and then again for a patch.
About coding quality or standards, the only one I have seen is the old ones from leaflabs, to which I try to adhere when I remember, but there is all kinds of code merged (mine is some of the worst for sure), but let's take it as constructive criticism and we try to keep the code clean, readable, etc.
@RickKimball if there is some sections of the code in particular that you think don't fit please point to those lines, that will help find issues.
A few weeks ago Roger and me discussed how some PRs stay here for a long time without anyone looking at them, and it's hard for him to check each one of them, and this is one of those.
So now that has attention let's provide constructive feedback so it gets to be as good as possible.

As I said above, I think the timezone stuff should be left out, since that varies so much, and countries change rules every once in a while. I'll give a shot at the one @csnol pointed together with this RTC code and see if it works fine with our core.

pkourany:
@victorpv well put and I'm in full agreement.  The single timezone function is really a helper function, nothing more and I agree that it should come out.

After playing with the library, I believe there is some fuzziness around the createAlarm() vs the attachAlarmInterrupt() usage.   The former sets the `RTC_ALARM_SPECIFIC_INTERRUPT` type for waking the processor from STANDBY/STOP while the latter sets the `RTC_ALARM_GLOBAL_INTERRUPT`.  Perhaps an overload on createAlarm() (so as to not affect existing library users) with an extra parameter specifying usage (eg. WAKE_FROM_SLEEP and GENERAL).  The `detachAlarmInterrupt()` would need to handle both cases as well.

csnol:
@pkourany  Thanks, you are right, I can't rename existed function even over 2yrs.  but excessively repackage will make difficult for user. 
you know , if use enough parameters, every functions can be included in 1. 
and for the functions generality, I must use same function name and usage of F4 RTC library @stevstrong . like attachAlarmInterrupt() and detachAlarmInterrupt() etc.. that is why I can't change these 2 function.  maybe I can add a discreateAlarm() for existing library users. 
Thanks again for your suggestion.

pkourany:
@csnol, understood.  Perhaps `removeAlarm()` is a better name (?).  Thanks for your hard work!

rogerclarkmelbourne:
Guys

Whats the current status of this.

stevstrong:
I just tested the blue pill example, it worked out of the box.
So it looks fine for me.

csnol:
@stevstrong Thanks for tested.

rogerclarkmelbourne:
Thanks. I did a test compile of a few projects and it seemed OK.

PR #337, opened 2017-08-31
add JLink upload support for OSX

janchrillesen:
This is tested using a chinese JLink clone with Arduino 1.8.3 and a unmodified Blue Pill board
Requires the JLink binaries pre-installed from https://www.segger.com/downloads/jlink/

This is identical to the linux/jlink_upload except for the path to the binary


rogerclarkmelbourne:
Thanks

PR #338, opened 2017-09-04
Correction to upload issue in Sloeber

victorpv:
As discussed in http://stm32duino.com/viewtopic.php?f=41&t=2535
Changes to the Generic Boards options so the maple dfu option is the default one and Sloeber uploads correctly.


rogerclarkmelbourne:
Thanks.

PR #339, opened 2017-09-12
F1 SDIO first version

victorpv:
To enable 4bit wide mode, uncomment from this line to line 496:
https://github.com/victorpv/Arduino_STM32/blob/patch-3/STM32F1/libraries/SDIO/SdioF1.cpp#L491

In line 513 in that same file you can adjust the speed. I have tested from 1Mhz (1000) up, but with my card reader doesn't seem to work fine past 1Mhz. I have checked with a logic analyzer and the frequencies look right,

### DIFF
--- a/STM32F1/cores/maple/sdio.cpp
+++ b/STM32F1/cores/maple/sdio.cpp
@@ -1,0 +1,198 @@
+/******************************************************************************
+  * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
+ * BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
+ * ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *****************************************************************************/
+
+/**
+ * @file libmaple/sdio.c
+ * @author stevstrong
+ * @brief Secure digital input/output interface.
+ */
+
+#include <libmaple/sdio.h>
+#include <libmaple/gpio.h>
+#include <boards.h>
+#include "wirish.h"
+
+
+//#include <libmaple/libmaple.h>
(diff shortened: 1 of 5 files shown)


rogerclarkmelbourne:
Tried to compile for my F103Z board and I get

BOARD_SDIO_D0

So i guess that definition is missing 

Which board did you select ?

victorpv:
I submited the change only for 1 board, the generic 103r, didn't want to make changes to a ton of files until we have the final code, you can add the same lines to the Z:
https://github.com/rogerclarkmelbourne/Arduino_STM32/pull/339/files#diff-010bbb9d6bf0c10efd6e8415cb230b28

victorpv:
I bet you can also compile for the R, and upload to the Z and should work.

rogerclarkmelbourne:
Hi Victor

OK. The Z is a superset of the R so I think it should work.

I'll try building it later

BTW. Do you have any example sketch code to try this with ?

Because at the moment all I'd be able to confirm is that it compiles for me.

Or did you post a sketch to the forum ?

Ideally we'd add an example of how to use this as well as the changes to the core

victorpv:
I was using the Bench.ino example from the latest SDFat. The use is similar to what Steve did with the F4 (copying it here with the changes for F1):
- get the SdFat lib from https://github.com/greiman/SdFat
- insert into line 164 of SdFatConfig:
CODE: SELECT ALL
#elif defined(__STM32F1__)
#define ENABLE_SDIO_CLASS 1
- use in example sketches
CODE: SELECT ALL
#include "SdioF1.h" // instead of #include "SdFat"

rogerclarkmelbourne:
@victorpv 

OK.

I'll try to install the latest SDFat etc, later today

rogerclarkmelbourne:
In bench.ino, do I need to change

// Set USE_SDIO to zero for SPI card access. 
#define USE_SDIO 0

to

// Set USE_SDIO to zero for SPI card access. 
#define USE_SDIO 1

??

But if I do that I get

SDIO_bench:39: error: 'SdFatSdioEX' does not name a type

:-(

rogerclarkmelbourne:
I'm going to merge this, as its mainly additional files and won't have any negative effects to the F103R by adding those defines.

Then perhaps a few more people can try it. ;-)

victorpv:
Hopefully so :)
I need to order a new 103 board that should have the sdio with all pins connected, hopefully that will help me test further too.

cheeta1:
hi there @rogerclarkmelbourne   did you managed to run th SDfat library with SDIO on stm32f103Z? 
also can you answer your own question given below

"
In bench.ino, do I need to change

// Set USE_SDIO to zero for SPI card access.
#define USE_SDIO 0

to

// Set USE_SDIO to zero for SPI card access.
#define USE_SDIO 1

??

But if I do that I get

SDIO_bench:39: error: 'SdFatSdioEX' does not name a type

:-(

"

thanks for the awesome library.

PR #340, opened 2017-09-12
Correct issue with I2C port speed not changing after begin()

victorpv:
The current implementation will not change speed if Wire.setClock is called after begin() since the flags are only applied when the port is enabled from being disabled.
Corrected that by adding 2 lines to disable the port, and then enable it again with the new settings.
Tested and confirmed the new speed is applied.

### DIFF
--- a/STM32F1/libraries/Wire/Wire.cpp
+++ b/STM32F1/libraries/Wire/Wire.cpp
@@ -97,8 +97,10 @@ void HardWire::setClock(uint32_t frequencyHz)
 			dev_flags &= ~I2C_FAST_MODE;// clear FAST_MODE bit
 			break;
 	}
-	i2c_disable(sel_hard);
-	i2c_master_enable(sel_hard, dev_flags);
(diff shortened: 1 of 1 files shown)


rogerclarkmelbourne:
Thanks Victor.

Makes sense....  (same as SPI, which needs to be disabled and re-enabled when some settings are changed)

I will test

victorpv:
Hold on, sending a further change. The code as sent did work fine for me, but to cover any possibility, I added a check to confirm whether the peripheral was already enabled, and if so, then disabled and re-enable. If the peripheral was not enabled, either because begin() had not been called yet, or the it had been disabled for something, then it keeps it disabled as it was.
Sending new commit to the same PR.

victorpv:
Now it has that check. Tested with eeprom library by inserting it before and after begin(), then did a full write/read of the eeprom.

rogerclarkmelbourne:
OK

rogerclarkmelbourne:
Thanks

PR #345, opened 2017-09-22
Leadingzero support for STM32F1 print and println

Tijgerd:
Added an extra optional argument for print and println
Default value = PRINT_NOARG
When the value is PRINT_LEADINGZERO, all HEX values will have a leading zero :+1:

### DIFF
--- a/STM32F1/cores/maple/Print.cpp
+++ b/STM32F1/cores/maple/Print.cpp
@@ -65,8 +65,8 @@ size_t Print::write(const void *buffer, uint32 size) {
 	return n;
 }
 
-size_t Print::print(uint8 b, int base) {
-    return print((uint64)b, base);
+size_t Print::print(uint8 b, int base, int arg) {
+    return print((uint64)b, base, arg);
 }
 
 size_t Print::print(const String &s)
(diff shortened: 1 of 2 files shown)


Tijgerd:
I believe the Print.h and Print.cpp are the same for the other STM32 devices but I don't have those devices to test the code :(

Tijgerd:
@rogerclarkmelbourne , Can this be tested?

stevstrong:
Can you provide a simple test sketch (for with and without your changes)?

Tijgerd:
@stevstrong 

void setup() {
  Serial.begin(9600);
  
  delay(4000); //my Serial Monitor doesn't always "catch" a serial command right after sending
  
  Serial.print(0x05, HEX, PRINT_LEADINGZERO);   // should print: 05
  Serial.print(0xF5, HEX, PRINT_LEADINGZERO);   // should print: F5 (no leading zero)
  Serial.println();  
  Serial.print(0x06, HEX);                      //should print: 6
  Serial.println();
  Serial.print(0x07, DEC, PRINTLEADINGZERO);    //should print: 7 (since the base is not HEX)
  Serial.println();
  Serial.println(0x08, HEX, PRINT_LEADINGZERO); //should print: 08 and a newline
  Serial.println(0x09, DEC, PRINT_LEADINGZERO); //should print: 9 and a newline
}

void loop() {
}

RickKimball:
Please don't even consider this change.

Tijgerd:
Why not? 
It is an addition, the code will hold backward compatability

RickKimball:
1.) It is incompatible with every other Arduino port out there. Any code you write that uses this extension will not work on any other Arduino port.

2.) You can use an { char buffer[16]; sprintf(buffer, "%02x",value); Serial.print(buffer); } to get the same effect

3.) You could use the Streaming library to do the same thing with a custom filter http://forum.arduino.cc/index.php?topic=131283.0/

4.) .. many more reasons

stevstrong:
Well, I have s lightly different opinion.
As long as Arduino (AVR) compatibility is given, I have no objection, it brings a nice feature, specially for printing HEX bytes. Of course it can be solved in other ways, too, but this is included in the printing class itself, so no need to use other (complicated) ways.
By the usage I don't like to write the define wording (it's long), but it should also work if one use the "1" as second argument.

Tijgerd:
Agreed :)

rogerclarkmelbourne:
I think Rick is correct.

Although this is backwards compatible, it only adds a few small features which seem very specific to one user.

I think its just going to cause confusion and supportability problems.

PR #351, opened 2017-09-22
SPI add const qualifier

greiman:
Add const qualifier to SPI member function arguments for write and transmit buffers.

Qualifying the write or transfer source buffer with const avoid compile errors or casting away const when calling a SPI member function with a const source buffer.


rogerclarkmelbourne:
Thanks

I few people e.g. Victor and Steve are the SPI expects, so I'll wait for them to take a look

And I'll try doing a local pull to confirm if I get any errors.

stevstrong:
SdInfo builds with the latest SdFat without any warning/error message, and it works as expected (for F1).

So from my side, it is safe to pull.

rogerclarkmelbourne:
Tested with the graphics test for ILI9341 and it compiled and worked OK.

So I've merged this change.

PR #352, opened 2017-09-28
Removed deprecated functionality from Print class for BYTE "base"

sweetlilmre:
Several sketches:
- ASCIITable.ino
- Midi.ino
- SerialCallResponse.ino
- SerialPassthrough.ino
- USB_ASCII.ino

make use of the deprecated Serial.print(value, BYTE).
This can be replaced by Serial.write() in the sketches, seems only to be used in these sketches and results in conflicts when BYTE is defined as something else e.g. in the Elm-chan FatFS library.

Please consider this pull request as a small cleanup operation.

### DIFF
--- a/STM32F1/cores/maple/Print.cpp
+++ b/STM32F1/cores/maple/Print.cpp
@@ -99,10 +99,6 @@ size_t Print::print(unsigned long n, int base) {
 }
 
 size_t Print::print(long long n, int base) {
-    if (base == BYTE) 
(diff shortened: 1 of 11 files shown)


rogerclarkmelbourne:
Thanks

We'll need to check this before I can merge.

rogerclarkmelbourne:
Looks OK.

I'll merge..

Thanks

PR #358, opened 2017-10-15
Merge MIDI HID branch

arpruss:
Roger,

After a couple of hours of work, I think I finally managed to get the addMidiHID branch to the point where it merges into master. I may have screwed up somewhere and the only device I have for testing is the stm32f103c8.

It would be really nice to have MIDI and HID support together with all the latest changes.


arpruss:
If you don't want to push it to master, you might want to push it to addMidiHID. This fixes various bugs in addMidiHID by updating it to the master as of yesterday. For instance, it fixes lacks of includes for header files, and it fixes a nasty bug somewhere in addMidiHID which broke the C++ `new` operator.

rogerclarkmelbourne:
If it will auto merge with Master, I do not have a problem with that, but it would need a lot of testing.

I'll need to start a thread on the forum, as I'll need to get multiple people to test this, not just me

arpruss:
It should auto merge, as long as master doesn't diverge in the meanwhile, but I certainly agree about the need for a lot of testing first. I do think it will add a lot of usefulness to STM32F1 boards to have keyboard / mouse / joystick emulation.

rogerclarkmelbourne:
It says it will merge at the moment, and only I can change the master, so I'll make this a priority so that it gets checked before the master changes again

rogerclarkmelbourne:
Just so I understand what has changed...

How did you create your PR..

Did start with the current master and update it with files and changes from the addMidiHID branch, or did you use the addMidiHID branch as your base, then merge in various things from master ?

The reason I ask, is I don't know how git handles these sorts of merges, where one branch is a long way behind but then gets pulled into Master

i.e if there are changes to a file, I don't know how git decided which change to go with

arpruss:
I went with a modified addMidiHID, then merged in Master. I had to manually resolve some conflicts. The resulting code still didn't have working usb. I narrowed down the issue to a few usb-related files. It turned out that the merge replaced those with files from Master that broke things in addMidiHID. I then then either edited the files manually to fix or replaced them with the versions in addMidiHID until it worked.

arpruss:
Specifically, there were two files which I had to simply replace with the addMidiHID versions. This lost any Master changes to them so that part should be looked at closely. Here is the commit which replaces the two files with the addMidiHID ones: https://github.com/rogerclarkmelbourne/Arduino_STM32/commit/734aa2041f4aa96f2e8dbed5975fa1b177c6665f

arpruss:
I'm now looking at that commit to make sure that by replacing the two files I didn't lose any bug fixes in Master. One of them -- usb_reg_map.c -- is OK: all that happened was that after addMidiHID was branched, some code that addMidiHID needed but that wasn't needed for other things got commented out, so I ended up uncommenting it.

The other one has a lot of differences, and it's the one that needs to be looked at carefully to see if there are overwritten bug fixes: STM32F1/cores/maple/libmaple/usb/stm32f1/usb_cdcacm.c

arpruss:
OK, I went through the Master history for usb_cdcacm.c since the time the branch diverged, and I manually entered in all the changes. The PR has been updated to include these fixes and it still works. So now the updated PR has all the fixes to usb_cdcacm.c that Master had.

Here's something else I just discovered I had to do: I had to update the PR to change BOARD_USB_DISC_* in a lot of variant board.h files to the correct pin number as per the data sheet.

Here's the long story of my last two hours. 

You have commit https://github.com/rogerclarkmelbourne/Arduino_STM32/commit/f7a576f2e0efcee958bdfd737d9af7a2f6c76051#diff-434f84a6f545ef9caa9cae5e5d85dcab that fixes an issue where PB10 is used for USB disconnection on hardware (e.g., blue pill) where PB10 is not connected to the right USB pin.

Unfortunately, when I updated addMidiHID with this commit, it broke USB HID and USB MIDI. The initial version of my PR basically dropped this commit. But now I investigated further.

As you may recall, that commit fixed an issue where the USB code used PB10 on hardware where PB10 isn't connected to USB. 

OK, so I investigated further. The usb_cdcacm.c comments say that the disconnection is done via the USBDP connection. I then looked at the data sheet for the stm32f103x8/B and found that USBDP is actually PA12 (see page 31). So whoever wrote the board.h files put in the wrong definition. It should have been PA12 instead of PB10 all along.

(I don't know why it still worked on my black pill with the wrong definition--maybe PB10 gets routed to PA12 or something--but in any case it works perfectly when I change it to PA12.)

I think what happened may have been a simple mistake by whoever wrote the initial BOARD_USB_DISC_* definitions. The first column in the table on page 31 of the medium density data sheet on the line USBDP line says "B10". So maybe someone assumed that the USBDP pin is PB10.  But it's not. The first column lists the physical pin number on LFBGA100 packages (which anyway aren't the relevant ones!) rather than the logical pin number, which is PA12 (explicitly listed in the "Pin name" column).

I checked four data sheets: low/medium/high density STM32F103, as well as GD32F103, and they all put USBDP at PA12. So on all the board variants explicitly labeled STM32F103* or GD32F103*, I changed the BOARD_USB_DISC_* definition to PA12. I also did this for the microduino variant, because the web says that's an STM32F103xC.  I added this to the PR.

I left unchanged these boards: Maple which had C12 in board.h, Maple Mini which had B9, Maple Ret6 which had C12 and STM32VLD which had B10. The datasheet says the STM32VLD has no USB hardware, so maybe it should be NULL for that. I don't know anything about these boards, or what hardware they have, but anyway both your commit and my PR leave them unchanged. Maybe we should change some of them to PA12 eventually.

RickKimball:
There is no true disconnect circuitry on most of the generic boards (Blue Pill/ Red Pill/ Black Pill). The stm32duino code **manipulates the actual USBDP pin aka PA12** to make it appear to work like a true disconnect circuit.  The one board I know that has disconnect logic is the Maple Mini, which is why the code has BOARD_USB_DISC_* defines exist at all.  The usb circuit works completely different for the maple mini compared to the generic boards.

* the generic disconnect logic for PA12 from variants/generic_stm32f103c/wirish/boards_setup.cpp
```
__weak void board_setup_usb(void) {
#ifdef SERIAL_USB
    #ifdef GENERIC_BOOTLOADER
        //Reset the USB interface on generic boards - developed by Victor PV
        gpio_set_mode(PIN_MAP[PA12].gpio_device, PIN_MAP[PA12].gpio_bit, GPIO_OUTPUT_PP);
        gpio_write_bit(PIN_MAP[PA12].gpio_device, PIN_MAP[PA12].gpio_bit,0);

        // Only small delay seems to be needed, and USB pins will get configured in Serial.begin
        for(volatile unsigned int i=0;i<512;i++);
        gpio_set_mode(PIN_MAP[PA12].gpio_device, PIN_MAP[PA12].gpio_bit, GPIO_INPUT_FLOATING);

    #endif

    // Roger Clark. Changed SerialUSB to Serial for Arduino sketch compatibility
    Serial.begin();
#endif
}
```
see @stevstrong link for a deeper discussion

arpruss:
OK, I was ignorant of that thread in my previous comment. Thanks for helping me. I think I can revert my last two additions to the PR.

I also think I understand why https://github.com/rogerclarkmelbourne/Arduino_STM32/commit/f7a576f2e0efcee958bdfd737d9af7a2f6c76051 was incompatible with HID and MIDI. The issue was that the changes the commit made to the connect/disconnect code in usb_cdcacm.c also needed to be made in usb_hid.c and usb_midi_hid.c, as otherwise they were written to a NULL device. 

I'll have to try things out at home to make sure it all works before I change the PR. Or I may close the PR and try again from scratch.

arpruss:
I am closing this PR, since I have a made a much cleaner one. The cleaner one leaves intact all the improved disconnection/connection code from Master, and just makes sure that HID and MIDI do the right thing when BOARD_USB_DISC_* is NULL.

PR #359, opened 2017-10-18
Half-hour TimeZone supported.

csnol:
Maybe it isn't best way. 
Suggestions needs.

Here is Timezone list. 
[https://www.epochconverter.com/timezones](url)


csnol:
The PR be closed.
Suggest use below function:
``` c++
time_t TimeZone(time_t t, double TZ) { return ( t + (TZ * SECS_PER_HOUR)); } // usage: localtime = TimeZone(UnixTime, 8);
```

PR #360, opened 2017-10-18
Added missing SDIO pins for F103V and F103Z

Squonk42:
Hi Roger,
Please review this simple patch!


rogerclarkmelbourne:
Thanks

I'm sure this will be OK to merge.

PS. The SDIO stuff has not been fully tested.

It works for Victor but I've not been able to get it to work yet.

Victor is waiting for another board to arrive so he can do more testing

PR #361, opened 2017-10-19
Add midi hid

arpruss:
I have cleaned up my last pull request for merging the MIDI HID branch. 

Here's what I did to generate this pull request. 
1. Fork your addMidiHID branch
2. Make some fixes/enhancements to the HID support: add missing `#include`s; make the joystick axes be x/y/rx/ry which makes more sense for most games; allow for batched joystick HID sending (sending one button at a time is very slow; fix connect/disconnect code for devices without external disconnect hardware; added a message to the upload tool; fixed a potential missing vtable warning; see: https://github.com/omegacentaurisoftware/Arduino_STM32/commit/bd09f78d279fb9835eecc273a6329b890e968a6a
3. Merge master into branch. Almost everything can merge automatically, except for STM32F1/cores/maple/usb_serial.h and STM32/boards.txt, which have to be manually reconciled.
4. The resulting merge doesn't compile until further small changes to STM32F1/cores/maple/usb_serial.h and STM32F1/cores/maple/libmaple/usb/stm32f1/usb_reg_map.c are made; see: https://github.com/omegacentaurisoftware/Arduino_STM32/commit/badfa0a4aae9d7224ceee4a2d7494b489d5ffb40
5. Test with my [CapacitivePiano](https://github.com/arpruss/ADCTouchSensor/tree/master/examples/CapacitivePiano) (MIDI) and [GameCubeAdapter](https://github.com/arpruss/gamecube-usb-adapter) (HID and, separately, serial) sketches on my Black Pill.
6. Tweaked the boards.txt and platforms.txt files so that USB mode is not tied to the upload method.

It looks to me that this merge does a much better job picking up the fixes from Master after the divergence point. Doing things in the above sequence resulted in almost no manual fix-ups being needed.


arpruss:
As mentioned before, if you don't want to merge this into master, we could at least replace the current (broken) addMidiHID branch with this PR.

arpruss:
I think we should abandon this pull request, in light of the libraries I talk about here: http://stm32duino.com/viewtopic.php?f=13&p=37937

It would make some sense to include these libraries in the core, but I won't generate a pull request until they are more thoroughly tested.

PR #365, opened 2017-10-26
Fixing readTemp unit conversions

Shaffer319:
The original not doing any unit conversions. For example let result=1700: 
   (1.54V - 1700unitless) / (4.34 mV/C) + 25C
This now converts the numerator to mV assuming STM32F103XXXX has a 12 bit ADC  and ranges from 0V to 3.6V.
Also the VCC should be changed too. I'm not sure where the 1.1 scale factor came from.

Can someone create an example to use this function with polling. I have tried and it causes my cpu board to hang. OLIMEXINO-STM32 development board

STM32F103X8
STM32F103XB
Datasheet
http://www.st.com/content/ccc/resource/technical/document/datasheet/33/d4/6f/1d/df/0b/4c/6d/CD00161566.pdf/files/CD00161566.pdf/jcr:content/translations/en.CD00161566.pdf

### DIFF
--- a/STM32F1/libraries/STM32ADC/src/STM32ADC.cpp
+++ b/STM32F1/libraries/STM32ADC/src/STM32ADC.cpp
@@ -67,8 +67,11 @@
     float STM32ADC::readTemp(){
         unsigned int result = 0;
         float temperature = 0.0;
-        result = adc_read(_dev, 16);
-        temperature = (float)((_V25-result)/_AverageSlope)+ 25.0; 
+        result = adc_read(_dev, 16); // raw ADC value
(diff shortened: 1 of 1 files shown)


rogerclarkmelbourne:
Interesting...

I presume this applies to other MCUs in the F103 series, e.g.  e.g F103VET F103ZET, F103VGT etc, and may also apply to the smaller versions (e.g. F103T ) ??

rogerclarkmelbourne:
I can't get any of the examples in this lib to work.

Please can you post the code you used to test this change

rogerclarkmelbourne:
Closed due to lack of response from OP

PR #367, opened 2017-10-29
Update HardwareTimer.cpp

stevstrong:
- added DMA enable/disable, part 2


rogerclarkmelbourne:
Steve, it would be better if these PRs were done is one commit, so I'm going to manually add them to the master myself, so that it can be a single commit

stevstrong:
You're right, sorry, but this way my PR is at least clean, otherwise I would most probably end up with a lot of non-relevant stuff included due to my incompetence  ... :/

rogerclarkmelbourne:
Actioned in https://github.com/rogerclarkmelbourne/Arduino_STM32/commit/4a3e23949b3efde74bf786cf91f78cb6315fed04

PR #375, opened 2017-11-13
Update timer.h

stevstrong:
- clear the IRQ flag before enabling the IRQ

This solves the problem of generating interrupt right after enabling it if the IF flag was previously set.


rogerclarkmelbourne:
Steve

Thanks

I'd noticed that annoying problem, but put code in my sketch to ignore the first trigger.

I thought it was probably unavoidable, but its good that you have fixed it

Issue #121, opened 2015-10-22
Missing _BV definition


sweetlilmre:
Hardly a burning issue but _BV seems to be undefined in the default set of includes and I couldn't seem to find it elsewhere. I'm trying to maintain a 'portable' set of code between a Maple Mini and Arduino Nano platform and this is a minor inconvenience. As this is a fairly common macro, perhaps it should be included somewhere?

sweetlilmre:
More on this. Seems its a known Due / ARM etc. core issue as described here:
https://github.com/arduino/Arduino/issues/1834

No movement on this since mid 2014 :(

rogerclarkmelbourne:
What does it do ?

I we can probably add it into one of the headers.

We have had to add various other macro's etc, as the original codebase we use is based on Arduino 0022, and as there is no overall API spec for the Arduino API, we only tend to know about omissions when people try to use things.

phenom128:
from atmel
http://www.atmel.com/webdoc/AVRLibcReferenceManual/group__avr__sfr_1ga11643f271076024c395a93800b3d9546.html

#define _BV(bit) \
	(1 << (bit))

#include <avr/io.h>

rogerclarkmelbourne:
Can you test that

#define _BV(bit) \
	(1 << (bit)) 

Does do what Atmel claim it does

"Converts a bit number into a byte value."

I can't see the on an ARM platform that this is going to result in a byte value.

More like an int, which is 32 bits

If this is supposed to return a byte, then I suspect the macro would need to cast the result back to a byte

phenom128:
well it was atmel avr libc,
apols fuzzy thinking or lack thereof

is'nt it returning a bit mask that the user has to apply in some way?

srp

rogerclarkmelbourne:
Yes, The result of this could be used as a bit mask

I think there are some other macros that we already have that do the same thing, but I understand that if you have a lib that uses this macro that it needs to be in the core

BTW. Does you code include avr/io.h ? i.e do we need a file with that name (and we'd stick the single marco in it, as using the whole AVR header could be problematic and cause conflicts

phenom128:
if 'our' code already has this macro albeit by another name and that macro
is 'visible' within
'our' include path, then a simple wrapper in that file would that suffice?

stephen

rogerclarkmelbourne:
My mistake 

The macros we have, don't do this.

I was thinking of this set

[...]

PR #377, opened 2017-11-13
Solve race condition with arduino IDE monitoring serial port

albertoa:
I encountered a race condition upon which, even though the udev rules were working and uploads succeed, the serial monitor can not re-establish communications.

If the serial monitor is left open when doing an upload the error shown is:

[...]
Copying data from PC to DFU device
Starting download: [##################################################] finished!
state(8) = dfuMANIFEST-WAIT-RESET, status(0) = No error condition is present
Done!
Resetting USB to switch back to runtime mode
Waiting for /dev/ttyACM0 serial...Done
processing.app.SerialException: Error opening serial port '/dev/ttyACM0'. Try consulting the documentation at http://playground.arduino.cc/Linux/All#Permission
	at processing.app.Serial.<init>(Serial.java:145)
	at processing.app.Serial.<init>(Serial.java:82)
	at processing.app.SerialMonitor$4.<init>(SerialMonitor.java:101)
	at processing.app.SerialMonitor.open(SerialMonitor.java:101)
	at processing.app.AbstractMonitor.resume(AbstractMonitor.java:104)
	at processing.app.Editor.resumeOrCloseSerialMonitor(Editor.java:2218)
	at processing.app.Editor.access$2200(Editor.java:79)
	at processing.app.Editor$DefaultExportHandler.run(Editor.java:2196)
	at java.lang.Thread.run(Thread.java:748)
Error opening serial port '/dev/ttyACM0'. Try consulting the documentation at http://playground.arduino.cc/Linux/All#Permission

I traced it to the device being opened prior to the permissions/ownership changes.

When testing via -c the upload script waits for the character device creation, but at that point the udev rules have not changed the permissions. By testing the readability of the device via the -r test, we wait until the new permissions have been applied.

I imagine this is specific to the Linux kernel and the speed of the machine being used. In my case:
arduino-1.8.5
Linux 4.4.0-98-generic #121-Ubuntu SMP
but the change should work on any system with proper udev rules / permissions.

### DIFF
--- a/tools/linux/maple_upload
+++ b/tools/linux/maple_upload
@@ -42,7 +42,7 @@ fi
 echo -n Waiting for ${dummy_port_fullpath} serial...
 
 COUNTER=0
-while [ ! -c ${dummy_port_fullpath} ] && ((COUNTER++ < 40)); do
+while [ ! -r ${dummy_port_fullpath} ] && ((COUNTER++ < 40)); do
     sleep 0.1
 done
 


rogerclarkmelbourne:
Thanks

PR #274, opened 2017-04-27
Fixed a few STM32F1 compiler warnings..

lacklustrlabs:
..by adding gcc specific tags e.g. `__attribute__((unused))`
~~Also replaced tab with spaces in some files.~~

PR #380, opened 2017-11-16
Suppressing 'warning: unused parameter ... [-Wunused-parameter]'

lacklustrlabs:
Same as #274 but only suppressing -Wunused-parameter.
I've only tested this on MapleMini (but compiled all of the boards). 

I think it's a good practice to let the compiler do whatever it can do identify bugs in my code, so I always use `-Wall` when compiling. But it serves no purpose to be met with a wall of red text when the platform is re-compiled from scratch. Therefore I think it's acceptable to suppress these (inevitable) false positives.

### DIFF
--- a/STM32F1/cores/maple/libmaple/adc_f1.c
+++ b/STM32F1/cores/maple/libmaple/adc_f1.c
@@ -203,7 +203,7 @@ void adc_foreach(void (*fn)(adc_dev*)) {
 #endif
 }
 
-void adc_config_gpio(adc_dev *ignored, gpio_dev *gdev, uint8 bit) {
+void adc_config_gpio(adc_dev *ignored __attribute__((unused)), gpio_dev *gdev, uint8 bit) {
     gpio_set_mode(gdev, bit, GPIO_INPUT_ANALOG);
 }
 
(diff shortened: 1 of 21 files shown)


lacklustrlabs:
I've compiled these changes with every available board except "Maple (Ret 6)" - that board does not compile vanilla from rogerclarkmelbourne:master.

The gcc compilers I've tested the changes on are 4.8.3-2014q1, 6.3.0 and 7.2

rogerclarkmelbourne:
Thanks.

As this changes loads of existing files I'll need to start a thread on the forum to discuss it before I can potentially merge it

rogerclarkmelbourne:
See http://www.stm32duino.com/viewtopic.php?f=53&t=2851

I think some of these are actually bugs, e.g. pulseIn does not seem to conform to the Arduino API any more

I think rather than masking this.

A lot of these should be fixed

rogerclarkmelbourne:
Looks like Wire.begin() is another api call which is not implemented correctly by the core.

There should be 2 versions, one with an argument and one without.

The one with the argument is the slave address and this should be used.

I don't have time to go though these one by one.

But please check all the API calls you have changed, and remove those where the core does not match with the latest Arduino API

lacklustrlabs:
I started to fix the I2C begin() situation but it quickly escalated out of scope of this PR. I'll just revert the WireBase.begin() change and make a separate PR for the I2c master/slave situation.

I've been though the list twice now, and WireBase.begin() is the only method that's remotely part of the Arduino API, am i missing something?

rogerclarkmelbourne:
Thanks

Did a local test build etc

Seems OK ;-)

PR #381, opened 2017-11-17
Fixing 'missing field initializers' warning

lacklustrlabs:
Same as #274 but only fixing the 'missing field initializers' warning.

The type stm32_pin_info has 7 fields, not 6. Now the last field (pinMode) will be initialized to 0.

### DIFF
--- a/STM32F1/cores/maple/boards_private.h
+++ b/STM32F1/cores/maple/boards_private.h
@@ -42,7 +42,7 @@
 
 /* Makes the PIN_MAP rows more human-readable. */
 #define PMAP_ROW(gpio_dev, gpio_bit, timer_dev, timer_ch, adc_dev, adc_ch) \
-    { gpio_dev, timer_dev, adc_dev, gpio_bit, timer_ch, adc_ch }
+    { gpio_dev, timer_dev, adc_dev, gpio_bit, timer_ch, adc_ch, 0}
 
 namespace wirish {
     namespace priv {
--- a/STM32F1/cores/maple/boards_private.h
+++ b/STM32F1/cores/maple/boards_private.h
@@ -42,7 +42,7 @@
 
 /* Makes the PIN_MAP rows more human-readable. */
 #define PMAP_ROW(gpio_dev, gpio_bit, timer_dev, timer_ch, adc_dev, adc_ch) \
-    { gpio_dev, timer_dev, adc_dev, gpio_bit, timer_ch, adc_ch, 0}
+    { gpio_dev, timer_dev, adc_dev, gpio_bit, timer_ch, adc_ch }
 
 namespace wirish {
     namespace priv {
--- a/STM32F1/cores/maple/wirish_types.h
+++ b/STM32F1/cores/maple/wirish_types.h
@@ -55,7 +55,6 @@ typedef struct stm32_pin_info {
     uint8 gpio_bit;             /**< Pin's GPIO port bit. */
     uint8 timer_channel;        /**< Timer channel, or 0 if none. */
     uint8 adc_channel;          /**< Pin ADC channel, or ADCx if none. */
-	uint8 pinMode; 				/**< mode specific by pinMode call (Roger Clark added to optimize compatibility with Arduino API*/
 } stm32_pin_info;
 
 /**
--- a/STM32F1/variants/STM32VLD/board.cpp
+++ b/STM32F1/variants/STM32VLD/board.cpp
@@ -51,7 +51,8 @@ void boardInit(void) {
 }
 
 // Note. See the enum of pin names in board.h
-
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmissing-field-initializers"
 extern const stm32_pin_info PIN_MAP[BOARD_NR_GPIO_PINS] = {
 
 
@@ -113,6 +114,7 @@ extern const stm32_pin_info PIN_MAP[BOARD_NR_GPIO_PINS] = {
 	{&gpiod,   NULL, NULL, 1, 0, ADCx},  /* PD2 */
 	{&gpiod,   NULL, NULL, 2, 0, ADCx},  /* PD2 */
 };
+#pragma GCC diagnostic pop
 
 extern const uint8 boardPWMPins[BOARD_NR_PWM_PINS] __FLASH__ = {
     PA0, PA1, PA2, PA3, PA6, PA7, PA8, PA9, PA10, PB0, PB1, PB6, PB7, PB8, PB9
--- a/STM32F1/variants/STM32VLD/board.cpp
+++ b/STM32F1/variants/STM32VLD/board.cpp
@@ -51,70 +51,67 @@ void boardInit(void) {
 }
 
 // Note. See the enum of pin names in board.h
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wmissing-field-initializers"
 extern const stm32_pin_info PIN_MAP[BOARD_NR_GPIO_PINS] = {
 
 
-    {&gpioa, &timer2, &adc1,  0, 1,    0}, /* PA0 */
-    {&gpioa, &timer2, &adc1,  1, 2,    1}, /* PA1 */
-    {&gpioa, &timer2, &adc1,  2, 3,    2}, /* PA2 */	
-    {&gpioa, &timer2, &adc1,  3, 4,    3}, /* PA3 */
-    {&gpioa,   	NULL, &adc1,  4, 0,    4}, /* PA4 */
-    {&gpioa,   	NULL, &adc1,  5, 0,    5}, /* PA5 */
-    {&gpioa, &timer3, &adc1,  6, 1,    6}, /* PA6 */
-    {&gpioa, &timer3, &adc1,  7, 2,    7}, /* PA7 */
-    {&gpioa, &timer1, NULL,  8, 1, ADCx}, /* PA8 */	
-    {&gpioa, &timer1, NULL,  9, 2, ADCx}, /* PA9 */	
-    {&gpioa, &timer1, NULL, 10, 3, ADCx}, /* PA10 */
-    {&gpioa,	NULL, NULL, 11, 0, ADCx}, /* PA11 */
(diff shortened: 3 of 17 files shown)


victorpv:
I like this one, but had that field ever been used? I remember looking at this in the past, and couldn't find any place in the core where that field was ever used.
Furthermore, I believe it may have had some usage years ago when PIN_MAP was in RAM, but being in flash, that field can't be modified at run time. So having it all 0 at compile time, and without possibility to change it, wouldn't be better to just remove it?

I think this is something Roger added a long time ago with some intention, but may have never been used. Either that, or was for some compatibility with Arduino, but if it is always 0, I wouldn't think it doesn't help with any compatibility.

stevstrong:
I think it is just a fill up byte as the pin map table is 32 bit aligned, so with or without the last byte the table will have the same size in memory (hopefully in flash).
Another question would be whether all those table could be moved in a single common file since all those map definitions seem to be equal.

lacklustrlabs:
Just wanted to clarify that the compiler implicitly fills missing fields with 0, my fix just kills the warning.

@stevstrong The files are not the same, number of pins differ from board to board.

While we're talking about, hypothetically, redesigning the PIN_MAP table. Wouldn't it be better to make use of the compiler? A static constexp function could reduce the memory footprint quite a bit.

For example:
```
static constexpr uint32_t* getGpioSetRegister(uint8_t pin);
static constexpr uint32_t* getGpioSetRegister(uint8_t pin) {
  return pin==0?reinterpret_cast<uint32_t*>(0x40010C10):
          pin==1?reinterpret_cast<uint32_t*>(0x40010C10):
           pin==2?reinterpret_cast<uint32_t*>(0x40010C10):
            pin==3?reinterpret_cast<uint32_t*>(0x40010C10):
             pin==4?reinterpret_cast<uint32_t*>(0x40010810):
              pin==5?reinterpret_cast<uint32_t*>(0x40010810):
               pin==6?reinterpret_cast<uint32_t*>(0x40010810):
                pin==7?reinterpret_cast<uint32_t*>(0x40010810):
                 pin==8?reinterpret_cast<uint32_t*>(0x40010810):
                  pin==9?reinterpret_cast<uint32_t*>(0x40010810):
                   pin==10?reinterpret_cast<uint32_t*>(0x40010810):
                    pin==11?reinterpret_cast<uint32_t*>(0x40010810):
                     pin==12?reinterpret_cast<uint32_t*>(0x40011010):
                      pin==13?reinterpret_cast<uint32_t*>(0x40011010):
                       pin==14?reinterpret_cast<uint32_t*>(0x40011010):
                        pin==15?reinterpret_cast<uint32_t*>(0x40010C10):
                         pin==16?reinterpret_cast<uint32_t*>(0x40010C10):
                          pin==17?reinterpret_cast<uint32_t*>(0x40010C10):
                           pin==18?reinterpret_cast<uint32_t*>(0x40010C10):
                            pin==19?reinterpret_cast<uint32_t*>(0x40010C10):
                             pin==20?reinterpret_cast<uint32_t*>(0x40010810):
                              pin==21?reinterpret_cast<uint32_t*>(0x40010810):
                               pin==22?reinterpret_cast<uint32_t*>(0x40010810):
                                pin==23?reinterpret_cast<uint32_t*>(0x40010810):
                                 pin==24?reinterpret_cast<uint32_t*>(0x40010810):
                                  pin==25?reinterpret_cast<uint32_t*>(0x40010810):
                                   pin==26?reinterpret_cast<uint32_t*>(0x40010810):
                                    pin==27?reinterpret_cast<uint32_t*>(0x40010810):
                                     pin==28?reinterpret_cast<uint32_t*>(0x40010C10):
                                      pin==29?reinterpret_cast<uint32_t*>(0x40010C10):
                                       pin==30?reinterpret_cast<uint32_t*>(0x40010C10):
                                        pin==31?reinterpret_cast<uint32_t*>(0x40010C10):
                                         pin==32?reinterpret_cast<uint32_t*>(0x40010C10):
                                          pin==33?reinterpret_cast<uint32_t*>(0x40010C10):
                                           nullptr; 
}
```
Gives the address of a GPIO set register per pin for a Maple Mini.  
If the application only uses one gpio then that's the only address stored in flash and/or ram.
I recognize that this is a _big_ change. And it's not really the subject of this PR, just interesting to speculate..

rogerclarkmelbourne:
Its redundant

The original intention was to store the current pinMode so that functions like analogRead and analogWrite would not need to set pinMode each time they were called.

However its actually quicker to set the pinMode (register) than to check for a variable 

BTW. Doesnt this line now need to be removed as well ?

https://github.com/rogerclarkmelbourne/Arduino_STM32/blob/master/STM32F1/cores/maple/wirish_types.h#L58

victorpv:
What I was referring is to pinMode not being needed, doesn't really matter if it has a random value, or is initialized to 0 manually or automatically.
Since pinMode is redundant, my vote goes to delete the line pointed out by Roger, and we don't have to add 0s to the PIN_MAP tables.
The tables are 32bit aligned, but if we remove the pinMode, the warnings will be gone and the compiler will pad whatever is needed.

Son rather than this PR, remove the pinMode line.

On a separate note, PIN_MAP is almost the same, but not quite, in every board. Depends on how many pins the MCU has, it has more lines or less. Also to keep copatibility with original Maple and Maple mini pin numberings, the tables on those 2 are sorted in a different way:
I.e. Maple Mini first entry:
https://github.com/rogerclarkmelbourne/Arduino_STM32/blob/master/STM32F1/variants/maple_mini/board.cpp#L57
Bluepill first entry:
https://github.com/rogerclarkmelbourne/Arduino_STM32/blob/master/STM32F1/variants/generic_stm32f103c/board.cpp#L58

stevstrong:
I also vote for Roger's solution.

Regarding PIN_MAP, what I meant is that e.g timer 4 ch 3 is always on PB8, does not matter which F103 variant is taken.
Of course, the size of the table could be then differentiated with #ifdefs dependent on package pin number.
The order in the table could be kept the same (PA0, PA1, ...).
The IC pin mapping to the board pin is given by [this enum](https://github.com/rogerclarkmelbourne/Arduino_STM32/blob/master/STM32F1/variants/maple_mini/board/board.h#L76) which can vary from variant to variant, but not the internally assigned function of the IC pin given by the table.

rogerclarkmelbourne:
Steve

I agree there is duplication with pin to Timer and pin to ADC etc

But I'm not sure there is a better way of doing this, as the only method I can think of is some sort of lookup for each pin into a different table, which causes larger and slower code :-(

lacklustrlabs:
Removing the pinmode field is the best solution, if it's not used that is. 

> But I'm not sure there is a better way of doing this, as the only method I can think of is some sort of lookup for each pin into a different table, which causes larger and slower code :-(

Now that we are using g++11 we can make use of constexpr functions to build the PIN_MAP table, with  no runtime overhead cost. 

Since we are not yet using c++14, it will be ugly with one single function return allowed.

stevstrong:
The PIN_MAP should be placed in flash, I am not sure whether the run-time solution would do that.

The look up for the single table for each variant would be solved by the [enum](https://github.com/rogerclarkmelbourne/Arduino_STM32/blob/master/STM32F1/variants/maple_mini/board/board.h#L76) as is currently the case, which maps each pin to an entry of the table. So no extra code is needed.

lacklustrlabs:
constexpr functions are not runtime, they are executed at compile time.

lacklustrlabs:
Here is an experiment I made before the PIN_MAP in flash era:

```
// Maple mini
static constexpr uint16_t getGpioBitMask(uint8_t pin) {
  return pin==0?(uint16_t)0x800:
          pin==1?(uint16_t)0x400:
           pin==2?(uint16_t)0x4:
            pin==3?(uint16_t)0x1:
             pin==4?(uint16_t)0x80:
              pin==5?(uint16_t)0x40:
               pin==6?(uint16_t)0x20:
                pin==7?(uint16_t)0x10:
                 pin==8?(uint16_t)0x8:
                  pin==9?(uint16_t)0x4:
                   pin==10?(uint16_t)0x2:
                    pin==11?(uint16_t)0x1:
                     pin==12?(uint16_t)0x8000:
                      pin==13?(uint16_t)0x4000:
                       pin==14?(uint16_t)0x2000:
                        pin==15?(uint16_t)0x80:
                         pin==16?(uint16_t)0x40:
                          pin==17?(uint16_t)0x20:
                           pin==18?(uint16_t)0x10:
                            pin==19?(uint16_t)0x8:
                             pin==20?(uint16_t)0x8000:
                              pin==21?(uint16_t)0x4000:
                               pin==22?(uint16_t)0x2000:
                                pin==23?(uint16_t)0x1000:
                                 pin==24?(uint16_t)0x800:
                                  pin==25?(uint16_t)0x400:
                                   pin==26?(uint16_t)0x200:
                                    pin==27?(uint16_t)0x100:
                                     pin==28?(uint16_t)0x8000:
                                      pin==29?(uint16_t)0x4000:
                                       pin==30?(uint16_t)0x2000:
                                        pin==31?(uint16_t)0x1000:
                                         pin==32?(uint16_t)0x100:
                                          pin==33?(uint16_t)0x2:
                                           (uint16_t)0x0000;
}
```

Permute that by the variant degree and the code will be 100% unreadable :face_with_head_bandage:

stevstrong:
I am curious how does this look in assembly.

lacklustrlabs:
I've never actually verified the assemby, but getGpioBitMask(1) should evaluate to exactly 0x400.

lacklustrlabs:
But then again, the subject of this PR is to remove the compilation warnings. How to consolidate the board.cpp code is kind of off topic. 

I could make a try to remove the `pinMode` field entirely and update this PR, if that's your preferred solution.

rogerclarkmelbourne:
Yes. Remove it entirely because I am pretty sure its now redundant

lacklustrlabs:
My pet projects (using Serial & SPI) builds with these changes for every platform (except Maple RET6)
They run on Maple Mini exactly like code from master.

rogerclarkmelbourne:
Thanks

Tested on BluePill and Maple mini using my normal SPI ILI9341 display test program, and everything still seems to work fine.

PR #383, opened 2017-11-18
Fixed 'warning: comparison between signed and unsigned integer expressions [-Wsign-compare]

lacklustrlabs:
Same as #274 but only fixing -Wsign-compare.

Most of these fixes are self explanatory.

The question is if `rcc_sysclk_src<<2` should be type-casted to `uint32` or `unsigned int`, as I did.

```
typedef enum rcc_sysclk_src {
    RCC_CLKSRC_HSI = 0x0,
    RCC_CLKSRC_HSE = 0x1,
    RCC_CLKSRC_PLL = 0x2,
} rcc_sysclk_src;

```

### DIFF
--- a/STM32F1/cores/maple/libmaple/rcc.c
+++ b/STM32F1/cores/maple/libmaple/rcc.c
@@ -61,7 +61,7 @@ void rcc_switch_sysclk(rcc_sysclk_src sysclk_src) {
     RCC_BASE->CFGR = cfgr;
 
     /* Wait for new source to come into use. */
-    while ((RCC_BASE->CFGR & RCC_CFGR_SWS) != (sysclk_src << 2))
+    while ((RCC_BASE->CFGR & RCC_CFGR_SWS) != (unsigned int)(sysclk_src << 2))
         ;
 }
 
--- a/STM32F1/cores/maple/libmaple/usb/stm32f1/usb_cdcacm.c
+++ b/STM32F1/cores/maple/libmaple/usb/stm32f1/usb_cdcacm.c
@@ -487,7 +487,7 @@ uint32 usb_cdcacm_rx(uint8* buf, uint32 len)
  * Looks at unread bytes without marking them as read. */
 uint32 usb_cdcacm_peek(uint8* buf, uint32 len)
 {
-    int i;
+    uint32 i;
     uint32 tail = rx_tail;
 	uint32 rx_unread = (rx_head-tail) & CDC_SERIAL_RX_BUFFER_SIZE_MASK;
 
@@ -505,7 +505,7 @@ uint32 usb_cdcacm_peek(uint8* buf, uint32 len)
 
 uint32 usb_cdcacm_peek_ex(uint8* buf, uint32 offset, uint32 len)
 {
-    int i;
+    uint32 i;
     uint32 tail = (rx_tail + offset) & CDC_SERIAL_RX_BUFFER_SIZE_MASK ;
 	uint32 rx_unread = (rx_head-tail) & CDC_SERIAL_RX_BUFFER_SIZE_MASK;
 
@@ -583,7 +583,7 @@ static void vcomDataTxCb(void)
 	uint32 *dst = usb_pma_ptr(USB_CDCACM_TX_ADDR);
     uint16 tmp = 0;
 	uint16 val;
-	int i;
+	uint32 i;
 	for (i = 0; i < tx_unsent; i++) {
 		val = vcomBufferTx[tail];
 		tail = (tail + 1) & CDC_SERIAL_TX_BUFFER_SIZE_MASK;
--- a/STM32F1/libraries/OLED_I2C/OLED_I2C.cpp
+++ b/STM32F1/libraries/OLED_I2C/OLED_I2C.cpp
@@ -283,7 +283,7 @@ void OLED::printNumF(double num, byte dec, int x, int y, char divider, int lengt
 
 	if (divider != '.')
 	{
-		for (int i=0; i<sizeof(st); i++)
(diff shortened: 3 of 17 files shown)


stevstrong:
Looks good to me.
Another question would be if we really need separate `syscals.c` for each variant in part?

rogerclarkmelbourne:
Steve

Thats a good point.

I just diff'ed 4 of the variants against the F103C and  diff did not report any difference, even to the Nucleo and the STM32VLD 

Perhaps we should try moving it to the cores/maple  and see if the core still compiles

stevstrong:
>> Perhaps we should try moving it to the cores/maple and see if the core still compiles

Yes, it makes sense.

rogerclarkmelbourne:
I tried moving it, but there is a warning

sbrk.c:(.text._sbrk+0x30): warning: undefined reference to `end'

But the code seems to run OK.
(I tried the ILI9341 graphic test)

I'm still not happy about changing this until I understand what this means.

rogerclarkmelbourne:
I've looked at the verbose output from the IDE and there is no difference in the way that syscall.c is compiled if its in the variant folder or in the core maple/libmaple  folder

But it is linked differently

When syscalls.c is moved to the core its linked into the "archive" of all the other core files, which is in turn linked to the sketch and the libraries etc as the final part of the build process.

When its in the variant folder syscalls.c.o is linked specifically in the final line of build process

So my guess is that the options used to link the "archive" must be removing something from syscalls which is needed later 

Anyway, as we can't  change the way the Arduino build things, the only way to unify syscalls into one file is to put the actual code in a header and include it into each syscall.c file in each variant folder.

And as syscalls.c is not that big and hardly every changes, I don't think its worth making that change

rogerclarkmelbourne:
@lacklustrlabs 

Unfortunately, your changes to remove warnings which I have now pulled, conflict wit this :-(

Can you update so its compatible with your PR (which is now in the Master

Thanks

lacklustrlabs:
@rogerclarkmelbourne Sure, no problem

stevstrong:
The F4 has the syscalls.c in the core/maple/libmaple folder.
But it does not include the _exit() function, which may cause the warning for F1.

lacklustrlabs:
@stevstrong the only warnings i get for F4 are -Wmisleading-indentation and -Wexpansion-to-defined
but i think those only shows up for gcc7.2.
Do you get any Wsign-compare warnings for F4?

stevstrong:
Sorry, I was answering to Roger, regarding moving syscalls.c to the core folder, make it common to all variants.
I tried to move it myself from the variant folder to the core folder, and I didn't get any warning nor error.

rogerclarkmelbourne:
@stevstrong 
Sounds like the F1 and F4 are different, as I get the warning on the F1

stevstrong:
Yea, that is strange, i don't get any warnings for F1.
I tested with my repo, but it should not make any difference though.

stevstrong:
I could finally reproduce the problem mentioned by Roger in [his comment](https://github.com/rogerclarkmelbourne/Arduino_STM32/pull/383#issuecomment-345502980).

The issue is identical with this one from the Arduino forum: https://github.com/arduino/Arduino/issues/1287. Namely, the linker evaluates the object files in a strange, not quite understood order, so that if included in the `core.a` it will not be correctly handled.
I searched and tried 2 days long many solutions on the net, but only these from Arduino team helped.

There are two fixes for that applied by Arduino team, one "temporary" and one "final", although I like the temporary one better than the "final".
- Temporary: - put `syscalls.o` in the linker script before `core.a`.: https://github.com/arduino/Arduino/commit/e43dcd8bedeadb30b16a707f00de52e22a101e7e
This means change the linker [combine pattern in `platforms.txt` line 90](https://github.com/rogerclarkmelbourne/Arduino_STM32/blob/master/STM32F1/platform.txt#L90):
`-Wl,--start-group "{build.path}/core/libmaple/syscalls.c.o" {object_files} "{build.path}/{archive_file}" -Wl,--end-group`
- Final: - add some extra flags specifying each function from `syscalls.o`.
This has to be defined:
`compiler.combine.flags=-u _sbrk -u link -u _close -u _fstat -u _isatty -u _lseek -u _read -u _write -u _exit`
and then add these flags to the same combine pattern line as above:
`-Wl,--start-group {compiler.combine.flags} {object_files} "{build.path}/{archive_file}" -Wl,--end-group`

As a conclusion, there are at least 2 solutions to get rid of the multiple identical `syscalls.c` files from the multiple variant folders. Doable if we want.

rogerclarkmelbourne:
Perhaps its best for me just to merge this PR and then look at the duplication in syscalls later i.e raise it as an issue

stevstrong:
I have tested a blinky sketch using SPI and sprintf with my repo (syscalls made common, remove from variants + linker directive " -specs=nosys.specs -specs=nano.specs") and everything seems to work.
I also adapted the F4 platform.txt with the mentioned linker directive, the code size got impressively reduced while the was code still working.
Here my commit: https://github.com/stevstrong/Arduino_STM32/commit/f7620661971f01b0609462dbd2e75bb2592590f8.

lacklustrlabs:
rebased

rogerclarkmelbourne:
Following the rebasing from Steve's repo, is this now ready for the big time,?

i.e Can I merge it ?

stevstrong:
This PR looks fine to me, although it is still using identical syscalls.c files for each variant.
I think it can be now merged, and then later one can discuss about moving syscalls.c file under core folder or not.

stevstrong:
I would welcome the merging of this issue, after the conflict resolution.

PR #384, opened 2017-11-19
Complete pulseIn() using existing time measure tech

lacklustrlabs:
I took some pointers from AVR Arduino pulseInLong() and merged with the current STM32 code:

```
unsigned long pulseInLong(uint8_t pin, uint8_t state, unsigned long timeout)
{
        // cache the port and bit of the pin in order to speed up the
        // pulse width measuring loop and achieve finer resolution.  calling
        // digitalRead() instead yields much coarser resolution.
        uint8_t bit = digitalPinToBitMask(pin);
        uint8_t port = digitalPinToPort(pin);
        uint8_t stateMask = (state ? bit : 0);

        unsigned long startMicros = micros();

        // wait for any previous pulse to end
        while ((*portInputRegister(port) & bit) == stateMask) {
                if (micros() - startMicros > timeout)
                        return 0;
        }

        // wait for the pulse to start
        while ((*portInputRegister(port) & bit) != stateMask) {
                if (micros() - startMicros > timeout)
                        return 0;
        }

        unsigned long start = micros();
        // wait for the pulse to stop
        while ((*portInputRegister(port) & bit) == stateMask) {
                if (micros() - startMicros > timeout)
                        return 0;
        }
        return micros() - start;
}
```

### DIFF
--- a/STM32F1/cores/maple/stm32f1/wiring_pulse_f1.cpp
+++ b/STM32F1/cores/maple/stm32f1/wiring_pulse_f1.cpp
@@ -1,5 +1,6 @@
 #include <wiring_pulse.h>
 #include "boards.h"
+#include "variant.h"
 /* Measures the length (in microseconds) of a pulse on the pin; state is HIGH
  * or LOW, the type of pulse to measure.  Works on pulses from 2-3 microseconds
  * to 3 minutes in length, but must be called at least a few dozen microseconds
@@ -28,13 +29,13 @@
   */
 uint32_t pulseIn( uint32_t pin, uint32_t state, uint32_t timeout )
 {
-   // cache the port and bit of the pin in order to speed up the
+   // cache the reg map and bit of the pin in order to speed up the
    // pulse width measuring loop and achieve finer resolution.  calling
    // digitalRead() instead yields much coarser resolution.
  
-	gpio_dev *dev=PIN_MAP[pin].gpio_device;
-	uint32_t bit = (1U << PIN_MAP[pin].gpio_bit);
-   
+   const gpio_reg_map * const regs = digitalPinToPort(pin)->regs;
+   const uint32_t bit = digitalPinToBitMask(pin);
+   const uint32_t stateMask = (state ? bit:0);
 
    uint32_t width = 0; // keep initialization out of time critical area
    
@@ -45,23 +46,23 @@ uint32_t pulseIn( uint32_t pin, uint32_t state, uint32_t timeout )
    volatile uint32_t dummyWidth=0;
    
    // wait for any previous pulse to end
-   while ( (dev->regs->IDR & bit)  == bit)   {
+   while ((regs->IDR & bit) == stateMask)   {
       if (numloops++ == maxloops)  {
          return 0;
       }
-	   dummyWidth++;
+      dummyWidth++;
    }
    
    // wait for the pulse to start
-   while ((dev->regs->IDR & bit) != bit)   {
+   while ((regs->IDR & bit) != stateMask)   {
       if (numloops++ == maxloops) {
          return 0;
       }
-	  dummyWidth++;
+      dummyWidth++;
    }
    
    // wait for the pulse to stop
-   while ((dev->regs->IDR & bit) == bit) {
+   while ((regs->IDR & bit) == stateMask) {
       if (numloops++ == maxloops)  {
          return 0;
       }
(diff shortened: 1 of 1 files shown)


lacklustrlabs:
Arduino pulseIn has this signature:
```
unsigned long pulseIn(uint8_t pin, uint8_t state, unsigned long timeout)
```
But i think that's worse than what we got now
```
pulseIn(1,0xf000,10000) -> unexpected results
```
My vote is that we keep the signature as is until they change to something reasonable like:

```
unsigned long pulseIn(uint8_t pin, bool state, unsigned long timeout)
```

lacklustrlabs:
I've tested these changes with some example [HC-SR04 code](https://randomnerdtutorials.com/complete-guide-for-ultrasonic-sensor-hc-sr04/). It works equally poor for new and old version of pulseIn().

I'm just getting 1-4 cm output. But i don't think its because of pulseIn() itself.

rogerclarkmelbourne:
There are better ways to do this.

The @RickKimball posted some code to www.stm32duino.com to use the CPU's own clock counter, or an alternative method is to use the hardware timers; but using the hardware timer would potentially conflict if the user also wanted to run PWM at the same time.

Ricks thread about the counter timer is here 

http://www.stm32duino.com/viewtopic.php?f=18&t=326

RickKimball:
And Pito's nice short defines ... http://www.stm32duino.com/viewtopic.php?f=18&t=2004

rogerclarkmelbourne:
@RickKimball 

Thanks.

lacklustrlabs:
Ok, if you look at the diff you'll notice that i didn't modify _how_ the microseconds are measured. 
The `width++` stuff is just like it was before. 
To correct or improve the time measuring was never my intent, just tried to complete/fulfill the API.

lacklustrlabs:
Ok, if you look at the diff you'll notice that i didn't modify _how_ the microseconds are measured/estimated. 
The `width++` stuff is just like it was before (Though I might have skewed the results a bit by shaving off 2-3 clockcycles from the inner loops)
To correct or improve the time measuring was never my intent, just tried to complete/fulfill the API.

Modifying system control bits like:
```
#define DWTEn()         (*((uint32_t*)0xE000EDFC)) |= (1<<24)
#define CpuTicksEn()    (*((uint32_t*)0xE0001000)) = 0x40000001
```
Can not be the responsibility of pulseIn(), it has to be done globally once for all. That should be done in  a system wide PR that touches all of the time measuring functions `micros()`, `pulseIn()` and maybe `millis().`

lacklustrlabs:
I made some tests with this code:
```
// Arduino Mega 2560 'time master'
const int outputPin = 22;
const int DELAY = 200;
const bool mode = LOW;

void setup() {
  pinMode(outputPin,OUTPUT);
  digitalWrite(outputPin, !mode);
}

void loop() {
  delay(DELAY);
  digitalWrite(outputPin, mode);
  delayMicroseconds(200);
  digitalWrite(outputPin, !mode);
  
  delay(DELAY);
  digitalWrite(outputPin, mode);
  delayMicroseconds(500);
  digitalWrite(outputPin, !mode);

  delay(DELAY);
  digitalWrite(outputPin, mode);
  delayMicroseconds(1000);
  digitalWrite(outputPin, !mode);
}
```

```
// Maple Mini 72MHz 'time measure'
const int pin_in = PB12;
void setup() {
  pinMode(pin_in, INPUT);
  Serial.begin(115200);
}
void loop() {
  uint32 length  = pulseIn(pin_in, LOW, 3000000);
  Serial.print("Pulse length=");Serial.print(length);Serial.println("us");
}
```

Result:
```
Pulse length=1007us
Pulse length=203us
Pulse length=507us
Pulse length=1007us
Pulse length=203us
Pulse length=501us
Pulse length=1007us
Pulse length=203us
...
```
And that's pretty much consistent with the periods measured by my SeleaeLogic@24Mhz sample rate.

These numbers are generated with the unmodified pulseIn():
 
```
Pulse length=1074us
Pulse length=216us
Pulse length=535us
Pulse length=1074us
Pulse length=217us
Pulse length=542us
Pulse length=1074us
Pulse length=216us
Pulse length=543us
```
Seems like replacing `dev->regs->IDR` with `*IDR` made the measurements more accurate. 
That's probably a g++ optimization issue though.

victorpv:
I think this should be merged as is, since it's an improvement alreadt.
About using the debug counters for micros, I agree that's a different problem and should be in a separate PR.
But about that, there have been several discussions in the forum, and even Cedric decided not to use it for the STM official core.

rogerclarkmelbourne:
Did a test compile and didnt get any errors. Sorry I don't have time to test this with hardware but I don't have a sig gen handy

PR #386, opened 2017-11-21
Update HardwareTimer.h documentation

topin89:
`attachInterrupt` and `detachInterrupt` functions can attach ISRs to channel 0,
which is overflow interrupt (update interrupt in RMs).

### DIFF
--- a/STM32F1/cores/maple/HardwareTimer.h
+++ b/STM32F1/cores/maple/HardwareTimer.h
@@ -177,7 +177,8 @@ public:
      * This interrupt handler will be called when the timer's counter
      * reaches the given channel compare value.
      *
-     * @param channel the channel to attach the ISR to, from 1 to 4.
+     * @param channel the channel to attach the ISR to, from 0 to 4.
+     *   Channel 0 is for overflow interrupt (update interrupt).
      * @param handler The ISR to attach to the given channel.
      * @see voidFuncPtr
      */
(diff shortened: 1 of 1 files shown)


rogerclarkmelbourne:
Is this just a change to the comments / documentation?

topin89:
Yes, just documentation for two functions. I checked code and both RMs for STM32F1, it should work. Definitely works for STM32F103C8T6.
Maybe I should add that `refresh` generates update interrupt to documentation. Or rewrite ISR handler so it would clear timer flags even when no handler attached.

stevstrong:
Updating documentation/comments is OK, but why rewrite the ISR handler? I didn't get the point.

topin89:
Just yesterday I had this situation:

```
Timer2.attachInterrupt(0, timeout_to_send);
Timer2.setPeriod(1000);
Timer2.pause();      //At least with Timer1 everything starts after setPeriod.
Timer2.refresh();    //Pitfall I jumped right in.
//Timer2.attachInterrupt(0, timeout_to_send); // At this point I checked ISR code
delay(2000);
Timer2.start();
```

Took me about 50 minutes to realize where was the problem.

stevstrong:
The correct order to call the functions would be:

```
Timer2.pause();      // stop the timer
Timer2.setPeriod(1000);
Timer2.refresh();    // generate update event, refresh registers
Timer2.attachInterrupt(0, timeout_to_send);
Timer2.resume(); // resume
```

I had the same problem as you. That is why we had a commit (https://github.com/rogerclarkmelbourne/Arduino_STM32/commit/dfb5bd4c30f42fa9cf2a0aef6b7d2a50a594fa97) which clears the respective interrupt flag before you attach an ISR, so it should work fine in your case, too, if you respect this order of calling the functions.

topin89:
Oh. Looks like I use outdated version. Sorry. And your solution is better anyway.
So let's change docstrings.

victorpv:
The comments look good to me too. I haven't had the issue neither one of you had so far, but I also like the commit on clearing the flag before attaching an interrupt, good catch.

rogerclarkmelbourne:
Thanks

PR #388, opened 2017-11-23
Replace hackish EEPROM size/location determination with something smarter

arpruss2:
The EEPROM size/location determination only worked on F1 boards with 128K flash. On boards with less flash, the EEPROM would be outside the address space, and on boards with more flash the EEPROM could overwrite code. (Well, that can still happen if the code size is big enough.)

The PR uses the upload size to figure out the flash size and page size.

### DIFF
--- a/STM32F1/libraries/EEPROM/EEPROM.cpp
+++ b/STM32F1/libraries/EEPROM/EEPROM.cpp
@@ -1,6 +1,22 @@
 #include "EEPROM.h"
 // See http://www.st.com/web/en/resource/technical/document/application_note/CD00165693.pdf
 
+static const uint32_t flashSize =
+    ARDUINO_UPLOAD_MAXIMUM_SIZE > 768*1024 ? 1024*1024 : 
+    ARDUINO_UPLOAD_MAXIMUM_SIZE > 512*1024 ? 768*1024 : 
+    ARDUINO_UPLOAD_MAXIMUM_SIZE > 384*1024 ? 512*1024 : 
+    ARDUINO_UPLOAD_MAXIMUM_SIZE > 256*1024 ? 384*1024 : 
+    ARDUINO_UPLOAD_MAXIMUM_SIZE > 128*1024 ? 256*1024 : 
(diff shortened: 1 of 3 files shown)


victorpv:
Personally I have a similar correction, but rather than add new options in platform.txt, made it all depend in the CPU type used, which already has many macros in the core selection things like how many timers are available or how many SPI ports are available, to be consistent with the how the core selects those other options.

victorpv:
I think I'd like more merging what I did with some of the changes here:

```
static const uint32_t eepromPageSize = EEPROM_PAGE_SIZE;
 +static const uint32_t eepromStartAddress =EEPROM_START_ADDRESS;
 +static const uint32_t eepromPage0Base = eepromStartAddress;
 +static const uint32_t eepromPage1Base = eepromStartAddress + eepromPageSize;
```

So the defines above, plus the static const usage, and add the conditions for every MCU version (I'm missing a few, like RF, RG...)

arpruss:
Only a few entries in boards.txt currently define the density, so even if one uses the MCU_* flag, one has to modify boards.txt. It seems a lot simpler to just use the upload size to figure out the flash size.

victorpv:
You shouldn't have to add anything, the Density is defined in this file for each MCU:
https://github.com/rogerclarkmelbourne/Arduino_STM32/blob/4021c32fdb809f16a61c073d04e1d8c2c487a767/STM32F1/system/libmaple/stm32f1/include/series/stm32.h#L71

arpruss:
I didn't notice that. That's nice!

However, I think that getting the flash size from the MCU definition is apt to be error-prone: there are lots of datasheets to look through, and easy to make a mistake, as opposed to a single table of flash size values.

rogerclarkmelbourne:
Victor

We could add
`
# define STM32_FLASH_END               ((void*)0x8020000)
`

But we'd need to make some slight modifications, because some of the MCU versions are grouped together because they have the same RAM etc

e.g.

`#elif defined(MCU_STM32F103RD) || defined(MCU_STM32F103RE) 
`
 block would need to be split

But I don't think this is a problem because its not that much extra text and its all  defines and not variables

Issue #390, opened 2017-11-25
HardwareTimer example uses "deprecated" methods


gnbl:
https://github.com/rogerclarkmelbourne/Arduino_STM32/blob/master/STM32F1/libraries/A_STM32_Examples/examples/Maple/TimerInterrupts/TimerInterrupts.ino uses methods declared deprecated, e.g. in https://github.com/rogerclarkmelbourne/Arduino_STM32/blob/master/STM32F1/cores/maple/HardwareTimer.h#L268

rogerclarkmelbourne:
OK

Please send a PR to fix this

PR #391, opened 2017-11-26
Update TimerInterrupts.ino

stevstrong:
- replace deprecated functions, see https://github.com/rogerclarkmelbourne/Arduino_STM32/issues/390


rogerclarkmelbourne:
Thanks

PR #392, opened 2017-11-26
Correct a bug in FreeRTOS

victorpv:
As listed in https://www.freertos.org/a00104.html#mailing_list:
"The definition of the StaticTask_t structure is incorrect if INCLUDE_xTaskAbortDelay is set to 1, or the MPU is used. The corrected structure definition can be obtained from here."
https://sourceforge.net/p/freertos/code/HEAD/tree/trunk/FreeRTOS/Source/include/FreeRTOS.h#l978


rogerclarkmelbourne:
Thanks

PR #323, opened 2017-08-02
Ili9341 develop

stevstrong:
- cleanup non-STM32 relevant parts
- add possibility to pass an SPI object to begin
- added new function: void pushColors(void* buf, uint16_t nr_pixels);
- optimize DC and CS pin toggling
- updated graphics example

PR #394, opened 2017-12-01
Corrected: TIMER_CCMR1_CC2S_INPUT_TI*

lacklustrlabs:
According to STM32F103 RM0008: 
```
Bits 9:8 CC2S[1:0]: Capture/Compare 2 selection
This bit-field defines the direction of the channel (input/output) as well as the used input.
00: CC2 channel is configured as output
01: CC2 channel is configured as input, IC2 is mapped on TI2
10: CC2 channel is configured as input, IC2 is mapped on TI1
11: CC2 channel is configured as input, IC2 is mapped on TRC.
```

I had problems with this in #273


stevstrong:
Overlapping with #323 which has already suggested a fix, but it was not yed merged.

lacklustrlabs:
Nice catch, I'll change this so there won't be any merge conflicts with #323

rogerclarkmelbourne:
@stevstrong 

I think I had some issues when I last tried https://github.com/rogerclarkmelbourne/Arduino_STM32/pull/323 

I'll try pulling it locally and testing, but to fully test it, I'll need to setup to read back from the display and upload to the PC and I'll have to dig out the last code I had which kinda's does that, and modify it to make a usable test

rogerclarkmelbourne:
Thanks.

I've tested it with the a "tone" example and the tones play the same as before the change.

PR #399, opened 2017-12-06
Wire is now compliant with the arduino core

aster94:
discussion: http://stm32duino.com/viewtopic.php?f=3&t=2932

i tested it with a mpu9250 and it works great

running a search for "hardWire" i found:

immagine

so really we are going to break only 2 codes, both examples


stevstrong:
Why don't you change those files, too?

aster94:
To be honest i was thinking to do:
-Correct library, examples, readme of adafruit_ssd1306 and oled_i2c
-Remove the hardware i2c scanner, it is useless
-i don t have any minimum idea of what the hell is rules.mk so this is up to you 😅

aster94:
@stevstrong if you agree i will commit these changes

stevstrong:
This is Roger's repo, so he shall agree with the changes.
The `rules.mk` is a make file, but I think it is not relevant, actually it could be deleted.
Regarding the examples, I would leave only one of them, the generic one, in which is explained what to do (which class to use) if software or hardware I2C is wanted.

aster94:
I did it now hardwire doesn't exist anymore in the core
i also corrected the calling of SoftWire.h (before you should call it with TwoWire)


resume:

before:
hardWire -> hardware wire
softWire -> didn't exist
twoWire -> software wire

now:
hardWire -> doesn't exist anymore
softWire ->  software wire
twoWire -> hardware wire


Note: i found some odd code inside oled_i2c/hardware/arm/HW_STM32.h if someone have time please check it

lacklustrlabs:
\<off topic>

> Note: i found some odd code inside oled_i2c/hardware/arm/HW_STM32.h if someone have time please check it

I think we need to make a group effort and see if all these legacy I2C drivers really are needed _inside_ the repository. The maintained, upstream version seems to work just fine for many of those libs
</off topic>

Testato:
+1
This renaming is very userfriendly end in line with the standard Arduino core

rogerclarkmelbourne:
It was my mistake not to rename the class when I switched the repo to using hardware I2C

I didnt realise that people instantiate the TwoWire class even though its not part of the published Arduino API

But I'd prefer if we were as compatibe as possible and the original code (before I changed it) was compatible

aster94:
news:
-moved low level stuff (WireBase.h and .cpp) inside the folder utility
-added keyword.txt (i love colors)
-added example in "i2c_scanner_wire.ino" to use the i2c2
-removed "i2c_scanner_hardwire.ino"

i tested in a maple mini softWire in pin 11 and 10
Wire both on IIC1 and IIC2
the test sketch i used is the i2c_scanner

all is working great

lacklustrlabs:
Just tested ssd1306_128x32_i2c.ino from [upstream Adafruit_SSD1306](https://github.com/adafruit/Adafruit_SSD1306) and it works with these changes.

With a little bit of hacking in Adafruit_SSD1306.cpp it works for SoftWire.h as well:
```
#include <SoftWire.h>
#define Wire SWire
SoftWire SWire(PB6, PB7, SOFT_FAST);
```

aster94:
thank for the test @lacklustrlabs 

anyway i added some keywords that could be usefull (like INPUT_PULLDOWN and PWM), i thought it was better to don't put these inside this pull request
if roger agrees to use them, these could go here:
https://github.com/rogerclarkmelbourne/Arduino_STM32/tree/master/STM32F1/libraries/A_STM32_Examples
[keywords.txt](https://github.com/rogerclarkmelbourne/Arduino_STM32/files/1542895/keywords.txt)

rogerclarkmelbourne:
Thanks

I tested with a VL53L0X, with uses I2C and everything still seems to work fine

PR #400, opened 2017-12-07
New library: Timer diagnostics

lacklustrlabs:
Added a library that can display and alter timer register values in a human readable form.
Useful when debugging timers without a proper debugger.
This code is, right now, only compatible with this core.

Note that this library uses `String` and other inefficient constructs, so it is strictly for non-critical applications.

Reviews and comments are welcome


victorpv:
Can you try to use a buffer and a char array instead of string?
I think the library is a good idea. I normally just use an stlink or a jlink for debugging, but doesn't harm to get more options for debugging.

lacklustrlabs:
I could print the data directly without going via the toString() method. But the library would still, at its core, have a large memory footprint. Each bit, or group of bits, are represented as an object. Each one of those objects have a position, size and a pointer (int8,int8,int32) and there are lots of bits in the timer registers. (Only the pointers are in ram, though)

If you guys think it's a good idea to remove toString() then I'll do it, but I don't think it will make much difference for the purpose of the library.

In hindsight I think I cheated a little bit. I created the objects for advanced timer, then removed entire registers to fit general and basic timers. If a basic timer and an advanced timer interprets the 'same' register differently the printouts will be incorrect. E.g if advanced_timer:TIMxCR1 is different from general_timer:TIMxCR1

lacklustrlabs:
Moving to [stm32_timer](https://gitlab.com/LacklustrLabs/lacklustrlabs_stm32_timer) as a multi-core library

rogerclarkmelbourne:
Is https://gitlab.com/LacklustrLabs/lacklustrlabs_stm32_timer a library, or replacement files for this core ?

lacklustrlabs:
It _will_ be a library for this core and Arduino_Core_STM32 (maybe STM32GENERIC if I can find the time), right now it's just some old workfiles.

PR #401, opened 2017-12-08
usb powerdown

arpruss3:
USB should be powered down on Serial.end().

### DIFF
--- a/STM32F1/cores/maple/libmaple/usb/stm32f1/usb.c
+++ b/STM32F1/cores/maple/libmaple/usb/stm32f1/usb.c
@@ -120,6 +120,12 @@ static void usb_suspend(void) {
     USBLIB->state = USB_SUSPENDED;
 }
 
+void usb_power_off(void) {
+    USB_BASE->CNTR = USB_CNTR_FRES;
+    USB_BASE->ISTR = 0;
+    USB_BASE->CNTR = USB_CNTR_FRES + USB_CNTR_PDWN;
+}
+
 static void usb_resume_init(void) {
     uint16 cntr;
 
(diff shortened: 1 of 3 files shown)


rogerclarkmelbourne:
Does USB power need to be enabled in Serial.begin ?

arpruss3:
I think it must already be, because (a) the bootloader powers down USB and obviously Serial is working, and (b) I have the powerdown code in my HID library, and I can shut down HID and start up Serial and it just works.

(One other thing that might be a good idea would be to move the USB interface reset code from the boards init file to Serial.begin(). But it doesn't seem necessary: I can switch between HID and Serial without it.)

rogerclarkmelbourne:
You seem to have pushed changes to loads of other unrelated files.

arpruss:
Oops. Sorry!

Each time I pull things from git, I have to change the attributes in tools/win to make the files executable on Windows, and I think all the extra changed files were files whose attributes got changed.

I really don't know how to work with git. :-( Is there a way you can just accept the changes to the USB files?

Of course, it would make things more convenient to you if you made the tools/win files executable as well, but that really should be a separate PR.

rogerclarkmelbourne:
It looks like your system is changing the permissions on file

tools/win/bmp_upload.bat 100644 → 100755 

Are you running Linux ?

These are files in the tools/win folder, so I can't quite see how a Linux system would arbitrarily change them.

The only way to partially accept a PR is to pull your PR to a local branch (I do this already), then pull all the individual files from my master, to overwrite your changes, then commit those changes then push the changes 

However as you changed 13 files of which 9 should not have been changed, I'd have to manually do that for 9 files 
(I'm not sure if I can pull via wildcard, it sometimes works but not always)

Its probably easier for me to manually edit the files, but the whole point of a PR is so I don't need to spend time and effort and potentially add errors by manually updating files

rogerclarkmelbourne:
I've tried to use your PR and overwrite the unrelated changes to file permissions, but it makes things messy, because Git still sees the changes to permissions and the changes to all those binary files get made twice; which is not a good idea.

I'll need to manually copy the 2 files you changed it my Master and commit them as if they were my change, as I cant see a better way to do it

rogerclarkmelbourne:
I've manually copied your files and committed them myself.
See https://github.com/rogerclarkmelbourne/Arduino_STM32/commit/9cd2f3a74366835930c4f3ba94193fc9db42ad10

Please confirm that the commit included all the necessary files (because your PR included changes to 18 unrelated files)

PR #404, opened 2017-12-09
Update BluePill-RTClock-test.ino

stevstrong:
- use build time for RTC
- reworked serial monitor output to use sprintf
- some fixes + cosmetics

This is a development of https://github.com/rogerclarkmelbourne/Arduino_STM32/pull/334.
The main change is to set the RTC to the build time being automatically parsed by the sketch.
Hope you find it useful.


stevstrong:
- globAlmCount is only read on main level, cleared only when the respective interrupt is detached.
- tt read/write is atomic secure, and write only happens when set over serial, in the rest of the code is only read.
- interrpt typos corrected, thanks for pointing to them.

Additionally I have reworked a bit the epoch time input over serial, it was not always correctly received in my tests, now seems to be ok.

rogerclarkmelbourne:
I can't see any instructions on how to use this example

Looking in the code it seems to be expecting 10 chars via serial, which I presume are the Data / Time, but its not clear what format this needs to be in.

I think it would be good if it had some instructions in the code or perhaps used __DATE__ __TIME__ if the date / time needs to be initially set for the example to work .

stevstrong:
OK, I will add some comments on usage.

rogerclarkmelbourne:
BTW. I just noticed it does use __DATE__ __TIME__ somewhere in the code

stevstrong:
Yes, it is the date and time taken from the GCC builtin defines, it reflects the date and time of the used platform (PC).

lacklustrlabs:
> globAlmCount is only read on main level, cleared only when the respective interrupt is detached.

> tt read/write is atomic secure, and write only happens when set over serial, in the rest of the code is only read.

They are both altered inside interrupt functions and then read in the loop(). If the compiler is set to heavy optimization it can conclude that 'these values have not change since i wrote or read them the last time (in loop()), so i'll just use that value and ignore what's actually at the address'. Volatile is just a way to force the compiler to always read and write the memory address.

From what i can tell, all read and write operations are atomic secure on stm32 (except data types over 32 bits)

csnol:
Nice and effic... codes!   Sorry for my English.
"Dummy" -> haha!

BTW:  
1. _DATE_ and _TIME_ is difficult reading for beginner. (like me).
2.  It is confused me 25 seconds.   ( These 25 seconds is time from start compile  to running ! )
``` C++
tt = rtclock.makeTime(mtt) + 25; // additional seconds to compensate build and upload delay
```
Suggest a faster and simple way under linux or macos. almost no delay.
``` C++
date +%s  > /dev/ttyS1     // Change '/dev/ttyS1' to your SerialPort.
```
3. Sorry for my English again!  Drive me crazy more than coding.

csnol:
One more question.
>God job finding those 'if & |' typos, such bugs can take me ages to track down.
@stevstrong 
What is difference of them?  Is it really a serious bug?    I can't understand it.  Thanks.
``` C++ 
if (x != m || y != n )    //  bool || bool 
if (x != m | y != n )     //  bool | bool
```

lacklustrlabs:
@csnol In this case it does not matter, that's why i said typos.

csnol:
@lacklustrlabs  Got it, Thanks.
About the 25s, I means it's unnecessary.So many uncertain factors there are. (Computers, OS, Upload method, Toolchains. )  
It's not good enough even for a second-level Clock. Of course It is more better than my last version.

You know, clock needed time correction. 
My suggestion is update it from SerialPort after MCU running.

rogerclarkmelbourne:
The `__DATE__ __TIME__` is only useful if you want to have a record of when the code was compiled

However for testing, its handy to be able to set the date and time, in examples 

As @lacklustrlabs has said, the 25 secs is the time between the code being compiled and it running on your MCU

If you reset the MCU, it will still use the same date time that has been compiled, so will be totally wrong if you reset the MCU the next day.

I like the idea of synchronising via serial, you could probably write a python script to do that

lacklustrlabs:
> I like the idea of synchronising via serial, you could probably write a python script to do that

Or copy from the UdpNTPClient example, it's all just a matter of how much effort you want to put into one example.

stevstrong:
I just added some features:
- update RTC only if current RTC clock is less than the timestamp built from `__DATE__` and `__TIME__`
- removed "Dummy" month
- bugfix: wrong month printed to serial monitor
- Processing sketch to update RTC over serial add to bottom section of the file

stevstrong:
Sorry, it seems that I again messed up some things with my latest patch, so I will close this PR.

rogerclarkmelbourne:
No worries

Perhaps you can backup your local copy of the github fork of my repo.

Then delete your fork on github, and make a new (clean) fork.
Then clone this clean fork to your local machine

Then copy files from your "dirty" forked clone e.g. just the RTC files to begin with

BTW. Sometimes I have to rename / move my local clone, as it gets very messed up and the best way is just to re-clone from github and copy file just a few files from my messed-up "dirty" version to the new clean clone.

Issue #403, opened 2017-12-08
util/atomic.h is missing


tfry-git:
The header file <util/atomic.h> is part of avr-libc (http://www.nongnu.org/avr-libc/user-manual/group__util__atomic.html) and thus may be assumed by arduino libraries. One example library using atomic.h is Mozzi.

Essentially it allows to block interrupts during certain code sections, and re-enabling them afterwards. While the need for such functionality may be more pronounced on 8-bit CPUs and their need for multi-step calculations, there are valid usecases for stm32 chips as well.

User "pico" has submitted a port of atomic.h at http://www.stm32duino.com/viewtopic.php?f=3&t=258&start=10 (see also the surrounding discussion in that thread), but it seems to have been forgotten.

I am using his code successfully, locally on an STM32F103C8T6. I can confirm that
- it compiles
- code using it works without obvious defects (the Mozzi library; and yes, I made sure to specifically test some functions that make use of "ATOMIC_BLOCK")
- I did not test, whether it actually protects from intervening interrupts (I have no reason to believe, otherwise).
- I did test to make sure that interrupts continue to work as expected outside the protected code (i.e. after ATOMIC_RESTORESTATE was called)

tfry-git:
Testcase for the patch. Works as expected (LED permanently on) on STM32F103C8T6 "blue pill".

```
#include <HardwareTimer.h>
#include <util/atomic.h>

HardwareTimer timer(1);

volatile uint32_t count;
volatile uint32_t dummy;

uint32_t countInterrupts () {
  count = 0;
  // This loop should take more than just a few microseconds, but not too long for us to wait
  for (uint32_t i = 0; i < (1 << 24); ++i) {
    dummy = i; // So compiler won't try to outsmart us
  }
  return count;
}

void timerInt () {
  ++count;
}

void setup() {
  pinMode(PC13, OUTPUT);

  timer.pause ();
  timer.setPeriod (10);
  timer.setChannel1Mode (TIMER_OUTPUT_COMPARE);
  timer.setCompare (TIMER_CH1, 1);
  timer.attachCompare1Interrupt(timerInt);
  timer.resume ();
}

void loop() {
  uint32_t a, b;
  ATOMIC_BLOCK(ATOMIC_RESTORESTATE) {
    a = countInterrupts ();
  }
  digitalWrite (PC13, a > 0);
  b = countInterrupts ();
  digitalWrite (PC13, b < 1000);
}
```

[...]

PR #405, opened 2017-12-10
Create new file "atomic.h"

stevstrong:
as discussed here: https://github.com/rogerclarkmelbourne/Arduino_STM32/issues/403, taken from http://www.stm32duino.com/viewtopic.php?f=3&t=258&start=10


rogerclarkmelbourne:
Thanks

I tested with several sketches, e.g. Tone() and I2C test and both didnt have any issues with this

Issue #402, opened 2017-12-08
error: 'constexpr' needed for in-class initialization of static data member -- while compiling STM32ADC example


tfry-git:
When including the header file Arduino_STM32/STM32F1/libraries/STM32ADC/src/STM32ADC.h - e.g. while trying to build one of the SingeConversionInterrupt example - I get the following error:

```
In file included from /bla/hardware/Arduino_STM32/STM32F1/libraries/STM32ADC/examples/SingleConversionInterrupt/SingleConversionInterrupt.ino:1:0:
/bla/hardware/Arduino_STM32/STM32F1/libraries/STM32ADC/src/STM32ADC.h:155:40: error: 'constexpr' needed for in-class initialization of static data member 'const float STM32ADC::_AverageSlope' of non-integral type [-fpermissive]
     static const float _AverageSlope = 4.3; // mV/oC   //4.0 to 4.6
```

Also a instance of the same error for the header line immediately after that. Compiler:
`arm-none-eabi-g++ (GNU Tools for ARM Embedded Processors (Arduino build)) 4.8.3 20140228 (release) [ARM/embedded-4_8-branch revision 208322]`

See also http://www.stm32duino.com/viewtopic.php?f=14&t=2207 . Changing "const" to "constexpr" in this and the following line allows compilation to proceed.

lacklustrlabs:
Confirmed. Using `-std=c++98` makes the error go away, so it was probably a side effect to the `-std=gnu++11` transition.

g++ 7.2 and 6.3 reports the same error (gnu++11).

tfry-git:
Fixed by https://github.com/rogerclarkmelbourne/Arduino_STM32/pull/407.

PR #406, opened 2017-12-11
Fix compilation with -std=gnu++11

tfry-git:
See https://github.com/rogerclarkmelbourne/Arduino_STM32/issues/402

### DIFF
--- a/STM32F1/libraries/STM32ADC/src/STM32ADC.h
+++ b/STM32F1/libraries/STM32ADC/src/STM32ADC.h
@@ -152,7 +152,11 @@ private:
     voidFuncPtr _DMA_int;
     voidFuncPtr _ADC_int;
     voidFuncPtr _AWD_int;
+#if __cplusplus >= 201103L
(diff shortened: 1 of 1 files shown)


lacklustrlabs:
I wonder if we really need the `#if __cplusplus >= 201103L` tests.
http://www.stm32duino.com/viewtopic.php?f=3&t=1241&p=38520#p38520

tfry-git:
I suppose it depends which compilers we may still want to support. If we are all set to stick with C++11(or later) for good, a fix without standards version check is even more trivial. I don't mind if this request is rejected, I'd just like to get the issue fixed one way or another.

stevstrong:
gnu11 was introduced without any fallback scenario.
So we don't need any ifdefs.

tfry-git:
Next attempt: https://github.com/rogerclarkmelbourne/Arduino_STM32/pull/407

(Seems I can change the target branch to merge to, but not the source branch to merge form in a PR)

PR #407, opened 2017-12-12
Fix compilation with -std=gnu++11

tfry-git:
See https://github.com/rogerclarkmelbourne/Arduino_STM32/issues/402

### DIFF
--- a/STM32F1/libraries/STM32ADC/src/STM32ADC.h
+++ b/STM32F1/libraries/STM32ADC/src/STM32ADC.h
@@ -152,7 +152,7 @@ private:
     voidFuncPtr _DMA_int;
     voidFuncPtr _ADC_int;
     voidFuncPtr _AWD_int;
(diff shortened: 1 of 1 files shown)


victorpv:
@rogerclarkmelbourne I have this same solution working in my local copy for a while now successfully.
Unless we add this, the library can't be compiled with the c11 flags, so everyone is having to modify their local copy.
I sent a duplicated PR for this but Steve pointed out to this one already waiting. My experience is that this is safe to do, and works fine. Tested with 4 ADC inputs in DMA.

victorpv:
@rogerclarkmelbourne this PR is safe and needed since we now use the gnu++11 flags.
Without it, the ADC library fails to compile. The PR only corrects the ADC library and doesn't touch anything else, and I have tested it successfully with Marlin, that reads multiple ADCs with DMA using that library.

PR #408, opened 2017-12-12
Necromancing libmaple I2C slave implementation [Experimental/Don't merge yet]

lacklustrlabs:
I've noticed that there once were an I2C slave implementation in libmaple.
They merged this code in 2013 and [removed it 2014](https://github.com/lacklustrlabs/libmaple/tree/I2C_slave_still_present), but I can't find why.

I've brought the i2c slave implementation back from the dead and merged it with the current code.
It seems to work with a limited example. It even works with master mode Wire.h ( Adafruit_SSD1306 )

I've tested `master_writer.ino` talking to `i2c_libmaple_slave.ino`. 

Obviously this needs extensive testing, and I can only test this on Maple Mini.

It seems like a large job to put slave mode into Wire.h, at least if done properly (individual buffers for both I2C channels, adhere to Stream API, etc etc) so that will have to be a separate PR.

All credit for this code should go to Barry Carter. barry.carter@gmail.com, headfuzz.co.uk who wrote this back in 2012.

#207 #324

PR #409, opened 2017-12-15
Temporary WireSlave library [Experimental]

lacklustrlabs:
This is a temporary library that makes it possible to test both the normal, non slave, Wire implementation and a version with added slave support (#408). All in the same repository.

~~The slave support implemented in this library is still only on the libmaple layer, so~~ it's a work in progress. The libmaple code is copied from #408 into this PR quick&dirty style, so the proper implementation is still #408.

Once this is fully implemented, tested, working properly and sanctioned by Roger,  - it can, hopefully, be merged into the #408 PR. This library should then be deleted.

To use this implementation instead of Wire you must include Wire_slave.h before Wire.h is included. That will make sure that any further references to Wire.h will be re-directed to Wire_slave.h. You will also have to make sure that you are compiling everything from scratch (change to different board and back)

I've successfully tested these examples with this i2c driver:
* master_writer talking to slave_receiver
* master_reader talking to slave_sender
* i2c_scanner_wire
* github.com/sleemanj/DS3231_Simple
* An AT24C32 lib (eeprom)
* ssd1306_128x32_i2c (lcd)


stevstrong:
I would push only the working examples.

victorpv:
I think it's ok to push the non working ones too since this is a test release, so the code is shared and everyone can contribute to get them working.

lacklustrlabs:
FYI: There is something wrong with master_reader.ino (running on stm32) talking to slave_sender.ino running on an AVR. I get the ASSERT throbbing led.

lacklustrlabs:
Ok, so I tried to add the slave functionality to WireBase, I think I started over three times only accomplishing a big mess and breaking SoftWire in the process. So instead I made and attempt basing the code on the Wire implementation from stm32duino/Arduino_Core_STM32. 

That code is more up to date with the latest Arduino API and it has lazy-allocation, expandable buffers.
It already had slave functionality, so all I had to do was to trick it into talking to libmaple instead of HAL.

From what I can tell it seems to work Ok. More tests are needed though.

There is one test that fails. - I tried to make Wire talking to Wire1 on the same machine, it crashes when the second I2C bus is started and it looks like an ISR is trying to call 0x00000000,  but I can't figure out why..  `selftest1.ino`

I intend to clean up the code a bit more, there are a few duplications of code that needs some attention  (e.g rxbuffer/txbuffer code)

lacklustrlabs:
Here is a crazy idea.. What if we always put the required libmaple code inside the Arduino library?
E.g libmaple I2C code inside Wire library and libmaple SPI code inside the SPI library.

That way we could reduce the flash size of applications that never uses those things.
The downside would be that libmaple would be coupled with the Arduino 'layer'.
But then again, does anybody use this repository only for the libmaple API?

The default i2c ISR vectors would have to be set to 'dummy' and overridden once Wire is included.

lacklustrlabs:
Here is a crazy idea.. What if we always put the required libmaple code inside the Arduino library?
E.g libmaple I2C code inside Wire library and libmaple SPI code inside the SPI library.

That way we could reduce the flash size of applications that never uses those things.
The downside would be that libmaple would be coupled with the corresponding Arduino 'layer'.
But then again, does anybody use this repository only for the libmaple API?

The default i2c ISR vectors would have to be set to 'dummy' and overridden once Wire is included.

rogerclarkmelbourne:
@lacklustrlabs 

In terms of putting the core funcs inside the lib, there are several more reasons why I don't think this is done

1. The linker should not link functions in the core which are not called. So there will be no advantage in terms of binary file size
2. People may want to write their own special I2C libs which directly use the LibMaple API for I2C and SPI. 
I've seen this before. If I remove an internal LibMaple API call or modify it to change its functionality, I bugs where people say their code no longer works
3. I think this would be on-standard,  I have a feeling that Arduino don't do it that way for their boards (and I'm not sure anyone does it that way)

rogerclarkmelbourne:
As this PR just add the WireSlave functionality, its low risk, and I think I should be able to merge it.

However, I will download and see if it compiles first ;-)

lacklustrlabs:
Note that this PR creates a Wire/library.properties file. Most notably it sets `include=Wire.h,SoftWire.h`
I did that to ensure that the Wire library and WireSlave isn't used at the same time. 
Once "WireSlave.h" is included it prevents/redirects further Wire.h inclusions.

These changes to Wire library are quite harmless, they do not break Wire library if WireSlave is removed.

rogerclarkmelbourne:
Thanks

I've merged this as its low risk, and it will allow more people to try it, if its in the Master.

fpistm:
Just saw that, for the issue with selftest1.ino.
https://github.com/rogerclarkmelbourne/Arduino_STM32/blob/42a5e9f9073e1f3b2a479abc4145a32b8e8c9339/STM32F1/libraries/WireSlave/src/Wire_slave.cpp#L402
Here it's possible the `rxBuffer` is not allocated.
```
  allocateRxBuffer(num_bytes);
  // error if no memory block available to allocate the buffer
  if (rxBuffer == nullptr) {
    return; // or assert or ..
}
```

PR #411, opened 2017-12-16
Adding PB2 as in the Generic R and Z versions.

victorpv:


rogerclarkmelbourne:
Looks OK.

I"m in the middle of looking at another PR at the moment, but this one should be easy and safe to merge, once I've processed the other one

lacklustrlabs:
Maybe keep the 'PB2 == boot1' comment

victorpv:
@lacklustrlabs after the enum?

lacklustrlabs:
@victorpv I was thinking in the margin of the PIN_MAP (board.cpp), all kinds of useful pin info is written there.

victorpv:
Ok, just added the comment there.

PR #415, opened 2017-12-20
use Adafruit_GFX instead Adafruit_GFX_AS

hyperion11:
Hello. Why not use Adafruit_GFX instead Adafruit_GFX_AS?
i check basic functions as fillscreen, line, rectangle, fillrectangle, fonts - they work fine.
image
https://youtu.be/g7dh1J98CCc

### DIFF
--- a/STM32F1/libraries/Adafruit_ILI9341_STM/Adafruit_ILI9341_STM.h
+++ b/STM32F1/libraries/Adafruit_ILI9341_STM/Adafruit_ILI9341_STM.h
@@ -7,7 +7,7 @@ This library has been modified for the Maple Mini
 #define _ADAFRUIT_ILI9341H_
 
 #include "Arduino.h"
-#include <Adafruit_GFX_AS.h>
+#include <Adafruit_GFX.h>
 #include <SPI.h>
 
 #ifndef swap
--- a/STM32F1/libraries/RTClock/src/rtc_util.c
+++ b/STM32F1/libraries/RTClock/src/rtc_util.c
@@ -1,0 +1,253 @@
+/******************************************************************************
+ * The MIT License
+ *
+ * Copyright (c) 2012 LeafLabs, LLC.
+ *
+ * Permission is hereby granted, free of charge, to any person
+ * obtaining a copy of this software and associated documentation
+ * files (the "Software"), to deal in the Software without
+ * restriction, including without limitation the rights to use, copy,
+ * modify, merge, publish, distribute, sublicense, and/or sell copies
+ * of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be
+ * included in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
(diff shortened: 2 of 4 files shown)


stevstrong:
Because many users complained that Adafruit_GFX.h is not found at build time.
As the _AS version is core internal, it is always found, there are no more complains. :)

hyperion11:
strange..
i install Adafruit GFX via library manager
image
change in Adafruit_ILI9341_STM.h include Adafruit_GFX instead Adafruit_GFX_AS and everythink work fine :)

stevstrong:
Yes, sure, if you install it additionally, it works. But not if don't install it.
Anyway, what is the problem with using the _GS version?

hyperion11:
Adafruit_GFX is newer. Support any custom fonts, generated by fontconvert utility. This is main reason :)
https://learn.adafruit.com/adafruit-gfx-graphics-library/using-fonts

stevstrong:
You can anytime change the lib as you suggested in this PR but I don't think it is worth to commit in main repo.
The basic functions you tested are overwritten anyway by the ILI9341 lib, so you actually did not test any "new" function.
Did you test any new Adafruit GFX fonts with the ILI9341 lib?

Btw, as far as can see, the new Adafruit lib also has an SPI wrapper module for TFTs: https://github.com/adafruit/Adafruit-GFX-Library. This could maybe imported by us, but I see here a lot of work, with small or no benefit at all.

hyperion11:
yes, i convert several fonts from Google Fonts site and use in my project.
ticking-timebomb40pt7b.h
ticking-timebomb20pt7b.h
CPMonoP9pt7b.h
ubuntu.condensed9pt7b.h
they works fine

tfry-git:
Not to be spamming, but to tie together the various strings: I have also submitted a "third option" PR, to base Adafruit_GFX_AS on Adafruit_GFX, so you can have the whole set of new Adafruit_GFX functions while keeping the "AS" add-ons for backwards compatibility.

Yes, this will mean you have to install Adafruit_GFX, separately, though that is pretty much common procedure _anywhere_ where pixel graphics are used in the Arduino world. You would not even have to include STM32 in your google search for missing "Adafruit_GFX", in order to be pointed to a valid solution. In contrast, any problems related to Adafruit_GFX_AS are ours entirely, as pretty much any use of Adafruit_GFX_AS outside of stm32duino seems to have ceased in favor of Adafruit_GFX.

@stevstrong : The Adafruit_GFX-class is actually doing two things: 1) Providing virtual functions for a handful of low-level drawing routines, which allows you "hook in" any hardware driver. 2) Provide a bunch of non-virtual high-level API functions on top of this, including, among other things, much more flexible font handling (basic buttons are another handy addition). So yes of course, basing ILI9341_STM on on Adafruit_GFX (directly, as suggested here, or indirectly as in my PR) means you automatically get to profit from any additions and fixes to the high-level API in Adafruit_GFX.

Anyway, perhaps the best place to discuss the best way going forwards is in the forum: http://stm32duino.com/viewtopic.php?f=9&t=3352

stevstrong:
I think your last update in rcc.h is wrong, the RTC_Enable bit is 15, so correctly it should be:
`#define RCC_BDCR_RTCEN                  (0x1 << RCC_BDCR_RTCEN_BIT)`

tfry-git:
Also, I don't think those commits were meant to end up in this pull request...

Issue #309, opened 2017-07-09
Failed to use ADC


wodiga:
When I try to use a STM32ADC example, I got a problem.
Error log like this:
Arduino:1.8.2 (Windows 7), 开发板："Generic STM32F103C series, STM32F103C8 (20k RAM. 64k Flash), Serial, 72Mhz (Normal)"

I:\arduino-1.6.12\arduino-builder -dump-prefs -logger=machine -hardware I:\arduino-1.6.12\hardware -hardware D:\Users\tracis\AppData\Local\Arduino15\packages -tools I:\arduino-1.6.12\tools-builder -tools I:\arduino-1.6.12\hardware\tools\avr -tools D:\Users\tracis\AppData\Local\Arduino15\packages -built-in-libraries I:\arduino-1.6.12\libraries -libraries D:\Users\tracis\Documents\Arduino\libraries -fqbn=Arduino_STM32-master:STM32F1:genericSTM32F103C:device_variant=STM32F103C8,upload_method=serialMethod,cpu_speed=speed_72mhz -ide-version=10802 -build-path D:\Users\tracis\AppData\Local\Temp\arduino_build_773450 -warnings=default -build-cache D:\Users\tracis\AppData\Local\Temp\arduino_cache_36075 -prefs=build.warn_data_percentage=75 -verbose I:\arduino-1.6.12\hardware\Arduino_STM32-master\STM32F1\libraries\STM32ADC\examples\MultiChannelContinuousConversion\MultiChannelContinuousConversion.ino
[...]

PR #416, opened 2017-12-20
Change floats const to constexpr in STM32ADC

victorpv:
Since we are using the C11 flags, we can't use static const with a float, results in error:
/STM32F1/libraries/STM32ADC/src/STM32ADC.h:155:40: error: 'constexpr' needed for in-class initialization of static data member 'const float STM32ADC::_AverageSlope' of non-integral type [-fpermissive]

### DIFF
--- a/STM32F1/libraries/STM32ADC/src/STM32ADC.h
+++ b/STM32F1/libraries/STM32ADC/src/STM32ADC.h
@@ -152,7 +152,7 @@ private:
     voidFuncPtr _DMA_int;
     voidFuncPtr _ADC_int;
     voidFuncPtr _AWD_int;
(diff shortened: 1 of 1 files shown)


victorpv:
This was referenced in https://github.com/rogerclarkmelbourne/Arduino_STM32/issues/309 and related forum post, but no one submitted a PR. I have it working fine in my local copy with constexpr. Tested with Marlin reading 4 ADC inputs with DMA.

stevstrong:
I think this is duplicate of https://github.com/rogerclarkmelbourne/Arduino_STM32/pull/407.

victorpv:
You are right, closing this one.

Issue #328, opened 2017-08-19
No DFU capable USB device available


drnickyoung:
Hi All,

Love the project and the idea, but bit of an issue, might be me missing something. I have got the bootloader on the STM32 and working. Uploaded the blink script and it works fine. Did several times to test out different blink delays, just testing the upload. Had to hit reset a few times, unplug and plug back in a couple but nothing major. Then went to do it again and all I get now is:

> Arduino: 1.8.3 (Mac OS X), Board: "Generic STM32F103C series, STM32F103C8 (20k RAM. 64k Flash), STM32duino bootloader, 72Mhz (Normal), Debug (-g)"
> 
> Sketch uses 17168 bytes (26%) of program storage space. Maximum is 65536 bytes.
> Global variables use 2824 bytes (13%) of dynamic memory, leaving 17656 bytes for local variables. Maximum is 20480 bytes.
> dfu-util 0.9
> 
> dfu-util: Invalid DFU suffix signature
> dfu-util: A valid DFU suffix will be required in a future dfu-util release!!!
> Copyright 2005-2009 Weston Schmidt, Harald Welte and OpenMoko Inc.
> Copyright 2010-2016 Tormod Volden and Stefan Schmidt
> This program is Free Software and has ABSOLUTELY NO WARRANTY
> Please report bugs to http://sourceforge.net/p/dfu-util/tickets/
> 
> dfu-util: No DFU capable USB device available
> An error occurred while uploading the sketch
> 
> This report would have more information with
> "Show verbose output during compilation"
> option enabled in File -> Preferences.

So the question is what is going on here?

Would appreciate any pointers.
Regards,
Nick

csnol:
Yes, I got a same error message.
``` c++
dfu-util 0.8

Copyright 2005-2009 Weston Schmidt, Harald Welte and OpenMoko Inc.
Copyright 2010-2014 Tormod Volden and Stefan Schmidt
This program is Free Software and has ABSOLUTELY NO WARRANTY
Please report bugs to http://sourceforge.net/p/dfu-util/tickets/

dfu-util: Invalid DFU suffix signature
dfu-util: A valid DFU suffix will be required in a future dfu-util release!!!
Deducing device DFU version from functional descriptor length
dfu-util: No DFU capable USB device available
```
BTW,  my dfu-util version is 0.8, How to update to 0.9?

rogerclarkmelbourne:
[...]

PR #417, opened 2017-12-24
wait for DFU device to appear

danielkucera:


lacklustrlabs:
I have not tested your modification yet, but it might be a good idea to update tools/linux64/maple_upload too

tfry-git:
Works great for me on Linux. Without this I get one successful upload in a dozen attempts or so (symptoms as in https://github.com/rogerclarkmelbourne/Arduino_STM32/issues/328), with this, I get reliable uploads.

In addition to linux and linux64, same thing on Mac, I presume.

(But actually, when is linux64 used? I _am_ running this on a 64 bit linux system, and for me, the "linux" code is being used.)

lacklustrlabs:
I'm not sure there is a preinstalled `lsusb` under mac, I think i installed mine via homebrew. 
(hint: `ioreg -p IOUSB -l -w 0`)

You will have to make custom changes to platform.txt (or use platform.local.txt) to make use of the linux64 tools. It's not really needed with you have IA32 installed, but some of us actually use the linux64 tools.

hanyazou:
The dfu_util itself have that function. Please see PR #297 and patch https://github.com/rogerclarkmelbourne/Arduino_STM32/pull/297/commits/19028676b46d2d6fd7089227e04b358e8a4b98b2

tools/linux/maple_upload
```diff
@@ -28,6 +28,12 @@ DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 
 "${DIR}/upload-reset" ${dummy_port_fullpath} 750
 
+echo "Waiting for device in DFU mode"
+while ! lsusb | grep -i ${usbID}
+do
+  sleep 1
+  echo "."
```
kolyshkin: I'd suggest doing `echo -n .` instead, i.e. remove the newline. Also, add an empty echo after `done`, so it will look like this:
```
Waiting for device in DFU mode .........
```

tools/linux/maple_upload
```diff
@@ -28,6 +28,12 @@ DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 
 "${DIR}/upload-reset" ${dummy_port_fullpath} 750
 
+echo "Waiting for device in DFU mode"
```
kolyshkin: add `-n` flag to `echo` (see below)

tools/linux/maple_upload
```diff
@@ -28,6 +28,12 @@ DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
...
+while ! lsusb | grep -i ${usbID}
+do
+  sleep 1
+  echo "."
+done
```
kolyshkin: add an empty `echo` (will print a newline) after done

danielkucera:
closing in favour of #297

PR #419, opened 2017-12-29
Update wirish_debug.cpp - reserve SWD lines only for debug

stevstrong:
reserve SWD lines only for debug

### DIFF
--- a/STM32F1/cores/maple/stm32f1/wirish_debug.cpp
+++ b/STM32F1/cores/maple/stm32f1/wirish_debug.cpp
@@ -37,5 +37,5 @@ void disableDebugPorts(void) {
 }
 
 void enableDebugPorts(void) {
-    afio_cfg_debug_ports(AFIO_DEBUG_FULL_SWJ);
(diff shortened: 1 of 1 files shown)


rogerclarkmelbourne:
Steve

Whats the difference between these 2 settings ?

I think it may be possible to do a sort of PrintF via SWD but I've never had time to try this.
I thought that debug print via SWD may be part of the ARM core of the chip, but I'm not sure how it could be viewed.

I have used Segger RTT on the nRF51822 MCU, which is basically the same thing, but using a JLink (and seggers code and their viewer)

stevstrong:
I forgot to mention the forum topic: http://www.stm32duino.com/viewtopic.php?f=35&t=3029
This will just release PB3 and PB4 for being reserved for debug interface and use them as GPIO, and make possible to upload via STLink simultaneously.
Many users had problems with not working PB3/4 when uploading with STLink.

rogerclarkmelbourne:
Ah OK

Its to downgrade to SWD only instead of full JTag, which no one uses.

I'll merge the PR as its low risk !

PR #423, opened 2018-01-09
Replace {build.path}/{archive_file} by {archive_file_path}

fpistm:
See https://github.com/arduino/Arduino/issues/6757#issuecomment-331861565
From @facchinm:

>   Replacing {build.path}/{archive_file} with archive_file_path
>   should remove the arduino_build_xxxxxx/.. part and make the path
>   a bit shorter.

Issue also raised here using BP:
http://stm32duino.com/viewtopic.php?f=3&t=3070


rogerclarkmelbourne:
Thanks

I'll need to test to see if it has an effect on build times etc

PR #424, opened 2018-01-09
option '#define XTAL16M' for boards with 16MHz-crystal

BlackBrix-Editor:
tested on STM32F103VE only but surely can be used on other STM32F103xC, STM32F103xD, STM32F103xE with the same clock PLL structure.
Therefore it could be used in other "../variant/.." -folders as well, if needed.
 No change to core-files needed.

### DIFF
--- a/STM32F1/variants/generic_stm32f103v/wirish/boards.cpp
+++ b/STM32F1/variants/generic_stm32f103v/wirish/boards.cpp
@@ -130,6 +130,12 @@ static void setup_clocks(void) {
     wirish::priv::board_setup_clock_prescalers();
     rcc_configure_pll(&wirish::priv::w_board_pll_cfg);
 
+#ifdef XTAL16M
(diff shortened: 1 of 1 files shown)


rogerclarkmelbourne:
Seems OK, but how do you intend to define XTAL16M (as an option in your boards.txt ?)

BlackBrix-Editor:
I put it into " ../ board / board.h " because there are all the things that are special to a board like LEDs, Buttons and so on ...
(so staying a variant-folder not touching the core files or the boards.txt)

--> http://stm32duino.com/viewtopic.php?p=39437#p39437

"...the #define XTAL16M -statement can be put into the referring "..\STM32F1\variants\YOUR_BOARD\board\board.h" then (if needed) ..."

rogerclarkmelbourne:
OK. That symbol is not used elsewhere, as far as I can tell, so I will merge it.

Thanks

PR #428, opened 2018-01-21
Update RTC bluepill

stevstrong:
Another try to continue #404 .
Relevant are only the first 3 files, the rest of 6 are actually unchanged, I don't know why github thinks they are changed.
Sorry, I cannot do it better.


rogerclarkmelbourne:
Steve

This PR now just seems to change some of the RTClock files.

But it also replaces files like syscalls.c with identical versions.

I can manually replace the RTC files with the ones in your PR and then commit this myself, then I won't have the confusion of unrelated files being replaced with identical copies of themselves.

stevstrong:
Yes, those syscalls.c files are those I mentioned as unchanged, but github considers them though as "changed".
Manually replacing and committing the 3 RTC-related files is the best idea.

rogerclarkmelbourne:
Hi Steve

OK.

I'll pull your PR to a separate local branch, then copy the RTC lib, outside of the git maintained folder, then switch back to my Master branch and copy the RTC lib files back into the git maintained folder, and hopefully Git will notice that just those 3 RTC files have changed

Thanks

rogerclarkmelbourne:
Steve.
Thanks...
I've manually copied the RTC files across to my Master and committed the change in

https://github.com/rogerclarkmelbourne/Arduino_STM32/commit/6f1429175cb5723e0bf3ec09c89950ae1f5b9cff

So I'll close this PR.

But if what I did, does not work, please re-open the PR.

Thanks

PR #429, opened 2018-01-23
Added platform.local.txt and boards.local.txt to .gitignore

lacklustrlabs:
See [core_stm32](https://github.com/stm32duino/Arduino_Core_STM32/pull/192) for description


rogerclarkmelbourne:
OK.  I don't use in Local mode, but I can see how this would be a problem for anyone attempting to clone etc and then commit and push to github

Issue #422, opened 2018-01-08
volatile qualifier


aster94:
hello to all!

I was trying to use this library: https://github.com/bolderflight/uNavAHRS
which needs this other lib: https://github.com/bolderflight/Eigen

but a lot of errors from the maple core appear, really it seems that the error is only one ` 'volatile' qualifiers cannot be applied to 'std::ios_base&`

any idea why this happen?

[i am using the last version of stm32duino from the board manager and a bluepill]

here a complete list of the errors:
`In file included from /home/vincenzo/.arduino15/packages/stm32duino/hardware/STM32F1/2018.1.3/cores/maple/boards.h:37:0,
                 from /home/vincenzo/.arduino15/packages/stm32duino/hardware/STM32F1/2018.1.3/cores/maple/wirish.h:54,
                 from /home/vincenzo/.arduino15/packages/stm32duino/hardware/STM32F1/2018.1.3/cores/maple/Arduino.h:30,
                 from /tmp/arduino_build_821086/sketch/uNavAHRS_MPU9250.ino.cpp:1:
/home/vincenzo/.arduino15/packages/stm32duino/hardware/STM32F1/2018.1.3/system/libmaple/include/libmaple/libmaple_types.h:55:14: error: 'volatile' qualifiers cannot be applied to 'std::ios_base&'
 #define __io volatile
              ^
/home/vincenzo/.arduino15/packages/stm32duino/hardware/STM32F1/2018.1.3/system/libmaple/include/libmaple/libmaple_types.h:55:14: error: 'volatile' qualifiers cannot be applied to 'const std::ios_base&'
 #define __io volatile
              ^
/home/vincenzo/.arduino15/packages/stm32duino/hardware/STM32F1/2018.1.3/system/libmaple/include/libmaple/libmaple_types.h:55:14: error: 'volatile' qualifiers cannot be applied to 'std::ios_base&'
 #define __io volatile
              ^
/home/vincenzo/.arduino15/packages/stm32duino/hardware/STM32F1/2018.1.3/system/libmaple/include/libmaple/libmaple_types.h:55:14: error: 'volatile' qualifiers cannot be applied to 'std::ios_base&'
 #define __io volatile
              ^
/home/vincenzo/.arduino15/packages/stm32duino/hardware/STM32F1/2018.1.3/system/libmaple/include/libmaple/libmaple_types.h:55:14: error: 'volatile' qualifiers cannot be applied to 'std::ios_base&'
[...]

PR #431, opened 2018-01-24
__io is replaced by the more correct __IO

aster94:
please see:
https://github.com/rogerclarkmelbourne/Arduino_STM32/issues/422
http://www.stm32duino.com/viewtopic.php?f=2&p=39999#p39999

[i also added some color highlight in keywords.txt]


lacklustrlabs:
Why are there .exe and .dll files in there?

rogerclarkmelbourne:
@lacklustrlabs 

I agree the exe's etc should not be in the PR.

I had another PR which had some exe's which seemed to have been something to do with the unix permissions

I could not merge that PR either, and had to manually merge the files that really needed to be changed

aster94:
I must have done some error with git, i am not very proficient with it
Tomorrow i will edit the PR and fix them

rogerclarkmelbourne:
No worries.  Git does strange things all the time.

Sometimes I have to rename the whole folder and clone again from the server, and then copy the files which have actually changed, however in this case as you have already pushed back to GitHub, you'd somehow need to get the exe's again and then push again and somehow update the PR (perhaps make a new PR)

You are not the only one who has problems like this ;-)

BTW. If all you did was a global search and replace on all c,cpp and h files for _IO and replace with __IO, I can easily do that in Notepad++ on Windows and perhaps make my own branch for this ??

aster94:
@rogerclarkmelbourne @lacklustrlabs I am reading only now after removing my guilty files, anyway now this PR doesn't change any .dd and .exe files

P.S: im sure that i didn't modified those files, is it possible that uploading a few sketches with the arduino IDE get them modified?

stevstrong:
This does not change F4 files, right?

aster94:
this does change F1,2,3,4
shouldn't i put them?

stevstrong:
I cannot see any F4 file in the changed file list.

stevstrong:
Ups, sorry, my browser seems to play nuts. I see now all files.

aster94:
never mind steve! 
if someone is wondering why there are 4 lines of difference between the add and delete, this is for "keywords.txt"
+1,256 −1,252  Files changed 118

aster94:
thanks @victorpv, i know that there are many reason to don't do this change, on the other hand i am sure it will solve problems in the future
it is like for the wire library, imho it is better to be compliant with the arduino world and in this case also with the compiler

rogerclarkmelbourne:
@victorpv 
Did you get around to trying this ?

I'm actually in favor of this, but I know it comes with a lot of risk

victorpv:
Just tested with the mp3 software in the bluepill RET board. It uses the SDIO port and SPI2/I2S, both with DMA, so does a fair amount of register access. Works fine.
If I get a chance in the next few days will test with the WAV player that uses 2 SPI ports and a timer with DMA.

rogerclarkmelbourne:
Thanks
Tested with a fairly complex application using 2 SPI devices and touch screen and also One Pulse timing and it compiled and ran fine

PR #434, opened 2018-01-28
Update board.h

hhdang:
Pins Update


hhdang:
STM32F103ZE's Pins is A to G，111 Pins(except BOOT1)

rogerclarkmelbourne:
Can you give more details on what this PR does and why ?

I see you removed "PD2".

Whats wrong with that pin ?

hhdang:
No,I removed "PB2", it reference to STM32F103VC's board.h,that file removed "PB2" because of BOOT1

rogerclarkmelbourne:
There is no reason to do that.

You can use PB2 for normal GPIO operation.


You just cant use it while boot0 is Low.

The Maple mini board uses it just fine as a normal pin 

see

https://github.com/leaflabs/maplemini/blob/master/maplemini.pdf

hhdang:
OK, I forked your old version,  your newest STM32F103V's board.h has added "PB2"。

rogerclarkmelbourne:
No sure why you'd want to have a fork where its removed.

If you don't want to use PB2, simply don't use it in your code

hhdang:
haha, i'm a junior github user, i forget to fetch the newest code before pull.

PR #435, opened 2018-01-31
Update BluePill-RTClock-test.ino

KenWillmott:
Clarify English documentation by changing all occurrences of "10 bits" to "10 digits", referring to Epoch time.


stevstrong:
It contains changed comments only, so I consider it safe.

victorpv:
And the changes make total sense. I also think this should be merged.

rogerclarkmelbourne:
OK. These are really low risk.

I'll merge them now

PR #438, opened 2018-02-03
Fixing now() function in RTClock.h

JiriBilek:
It is my first post here so let me first thank you for the library. I like it very much and it helps me with fast development of STM32 apps.

**Now the issue:**

The now() function was defined in RTClock.h using the #define directive. That is
wrong as it prevents using the keyword "now" in other situations.

The minimal error case is:

```
int now;

#include "RTClock.h"

void setup() {
  now = 1;
}

void loop() {}
```

Crashes with:
```
In file included from D:\sketch_feb03a\sketch_feb03a.ino:3:0:

D:\sketch_feb03a\sketch_feb03a.ino: In function 'void setup()':

D:\arduino-1.8.5\hardware\Arduino_STM32\STM32F1\libraries\RTClock\src/RTClock.h:55:25: error: 'getTime' was not declared in this scope

             #define now getTime

                         ^

D:\sketch_feb03a\sketch_feb03a.ino:6:3: note: in expansion of macro 'now'

   now = 1;

   ^

exit status 1
Error compiling for board Generic STM32F103C series.
```

### DIFF
--- a/STM32F1/libraries/RTClock/src/RTClock.h
+++ b/STM32F1/libraries/RTClock/src/RTClock.h
@@ -53,7 +53,7 @@ class RTClock {
 	time_t getTime();
 	void getTime(tm_t & tmm );
 
-    time_t now() { return getTime(); }
+	time_t now() { return getTime(); }
 	void now(tm_t & tmm ) { getTime(tmm); }  // non-standard use of now() function, added for compatibility with previous versions of the library
 
 	uint8_t year(void)    { getTime(tmm); return tmm.year; }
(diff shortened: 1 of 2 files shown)


stevstrong:
It would be even better if you declare the new functions as "inline".

lacklustrlabs:
..and fix-up indentation of your new method a bit, [STM32F1/libraries/RTClock/src/RTClock.h looks crooked](https://github.com/rogerclarkmelbourne/Arduino_STM32/pull/438/files)

But the idea is good, macros should never be named like frequently used variable names. 👍

JiriBilek:
I believe that member functions that are defined in class definition are inline by default.
See [https://www.cprogramming.com/tips/tip/member-functions-are-inline-by-default-cincrement](https://www.cprogramming.com/tips/tip/member-functions-are-inline-by-default-cincrement)

stevstrong:
It looks good for me.

PR #440, opened 2018-02-10
Update timer.h

stevstrong:
corrected mask for OCxM bits in CCMR1/2 registers (see RM0008 rev.16 p. 413)


rogerclarkmelbourne:
Thanks

I'll double check the manual and merge if it looks OK

stevstrong:
Reference to RM0008: chap. 14.4.7 ad 15.4.7:
Bits 14:12 OC2M[2:0]: Output compare 2 mode
Bits **6:4** OC1M: Output compare 1 mode
Both OCxM parameters are 3 bits wide, hence the mask should be 0x7.
These defines are, however, not used in the SW, so the risk something to fail is zero.

rogerclarkmelbourne:
OK.

Tested with my One Pulse + 2 x SPI etc test rig and it worked fine

PR #441, opened 2018-02-11
Update rtc_util.c

KenWillmott:
Modify rtc_get_count() and rtc_get_divider() to fix non-atomic accesses to RTC counter registers

### DIFF
--- a/STM32F1/libraries/RTClock/src/utility/rtc_util.c
+++ b/STM32F1/libraries/RTClock/src/utility/rtc_util.c
@@ -182,8 +182,10 @@ void rtc_set_count(uint32 value) {
 	rtc_wait_sync();
 	rtc_wait_finished();
 	rtc_enter_config_mode();
-	RTC->regs->CNTH = (value >> 16) & 0xffff;
-	RTC->regs->CNTL = value & 0xffff;
+	do {
+		h = RTC->regs->CNTH & 0xffff;
+		l = RTC->regs->CNTL & 0xffff;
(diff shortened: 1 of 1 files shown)


rogerclarkmelbourne:
Do you have some test code to demonstrate the problem in the existing code ?

KenWillmott:
It's a bug with a predicable probability of one occurrence approximately every 2.2 years at most. So one can write test code, but then one might have to wait that, or even longer, for conclusive results.

rogerclarkmelbourne:
Do you meant its not possible to set the criteria such that the fault occurs in less time?

rogerclarkmelbourne:
I did a test build, but have no way to fully test it.

So I've merged this in an attempt to clear the backlog of PR's

stevstrong:
I don't think that this atomic access would solve the posed problem, because what this PR is trying to solve is the consecutive read of registers whose value can change between the accesses, independent on the interrupts, only due to the working manner of the hardware interface.

edogaldo:
Ok, got the point, thanks!

victorpv:
I see the problem, I missed it when I first review that the first change is done in the wrong function, should have been in rtc_get_count(), not rtc_set_count() as Steve pointed out.
https://github.com/rogerclarkmelbourne/Arduino_STM32/blob/c6ef07affed27af470504f5cf85f248f3c781ef1/STM32F1/libraries/RTClock/src/utility/rtc_util.c#L171-L172

@KenWillmott can you correct both functions?

edogaldo:
Fix available in PR #488

PR #443, opened 2018-02-14
Add getTextHeight() and getTextWidth()

ImpulseAdventure:
- Extract out common text sizing logic into standalone functions
- This API addition also supports the integration of `Arduino_STM32/Adafruit-GFX-AS` into the `GUIslice` library (and any other libraries that offer more advanced text alignment functionality)
- An inconsistency was noted in the original code between the horizontal positioning gap defined in `drawUnicode()`, `drawCentreString()` and `drawRightString()` for `LOAD_FONT4` mode. This PR commit updates `drawRightString()` to use the same gap (`-3`) as the other calls for consistency. However, if there was a reason for the deviation, that particular change could be reverted.

### DIFF
--- a/STM32F1/libraries/Adafruit_GFX_AS/Adafruit_GFX_AS.cpp
+++ b/STM32F1/libraries/Adafruit_GFX_AS/Adafruit_GFX_AS.cpp
@@ -6,7 +6,7 @@ paired with a hardware-specific library for each display device we carry
 
 Adafruit invests time and resources providing this open source code, please
 support Adafruit & open-source hardware by purchasing products from Adafruit!
- 
+
 Copyright (c) 2013 Adafruit Industries.  All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
@@ -94,7 +94,7 @@ void Adafruit_GFX::drawCircle(int16_t x0, int16_t y0, int16_t r,
     x++;
     ddF_x += 2;
     f += ddF_x;
-  
+
     drawPixel(x0 + x, y0 + y, color);
     drawPixel(x0 - x, y0 + y, color);
     drawPixel(x0 + x, y0 - y, color);
@@ -126,7 +126,7 @@ void Adafruit_GFX::drawCircleHelper( int16_t x0, int16_t y0,
     if (cornername & 0x4) {
       drawPixel(x0 + x, y0 + y, color);
       drawPixel(x0 + y, y0 + x, color);
-    } 
+    }
     if (cornername & 0x2) {
       drawPixel(x0 + x, y0 - y, color);
       drawPixel(x0 + y, y0 - x, color);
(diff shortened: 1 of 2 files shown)


tfry-git:
Hi!

I guess I should have checked existing PRs, earlier. Anyway, I just submitted a different PR about Adafruit_GFX_AS: https://github.com/rogerclarkmelbourne/Arduino_STM32/pull/461 . The idea in mine is to base Adafruit_GFX_AS on Adafruit_GFX (which is compatible with STM32 out of the box, these days), while keeping the Adafruit_GFX_AS as a compatibility-layer, only.

My idea is that this will mostly obsolete Adafruit_GFX_AS, while providing a smooth migration path to Adafruit_GFX (which has getTextBounds()). However, it would certainly also be possible to continue developing Adafruit_GFX_AS as an add-on to Adafruit_GFX. In this case it would be easily possible to merge your additions into my split.

Would be great to hear your thoughts on this!

rogerclarkmelbourne:
Personally I think if the _AS version has been phased out, we should move away from it as the default

I'm not sure we even need to keep it, as people can always go back to an old commit to get an older version if things won't compile

ImpulseAdventure:
Hi Thomas -- that would be great! IMHO, reverting to the upstream GFX for the updated base functionality and then just offering GFX_AS as a compatibility-layer seems like a nice solution.

Roger -- it would be good to phase it out, but note that some of the examples (eg. in ILI9341_STM) depend on the additional functionality outside of the upstream GFX. Thomas' approach with a compatibility layer would appear to still keep those examples intact.

So, perhaps the default across the board could be changed to GFX, with the exception for the few examples (that leverage the GFX_AS string functionality) -- these would continue to import the GFX_AS compatibility wrapper?

tfry-git:
@ImpulseAdventure -- To clarify, the choice between "GFX" and "GFX_AS" will still be at the driver level, it cannot simply be limited to a specific example with the solution I proposed (it's not hard at all to switch the driver from one base to the other, but it still needs a modification to the lib).

Now, "keeping" the driver based on the "new" GFX_AS will mean it gets all updates to "GFX", _and_ the compatibility layer. Also the compatibility layer will be "for free" until a sketch actually uses its functions (at which point the additional fonts will be linked in, taking up a substantial bit of flash).

ImpulseAdventure:
@tfry-git -- got it.

I have now tested the GUI library ([GUIslice](https://github.com/ImpulseAdventure/GUIslice) branch "arduino_stm32") with your pull request #461 and it works nicely! Tested with both the default font (`LOAD_GLCD`) and the additional GFX FreeFonts. As expected, your PR greatly simplifies at least one aspect of porting libraries to Arduino_STM32. Thanks!

victorpv:
@ImpulseAdventure if all the functionality added here is included in #461 can this PR be closed?
The #461 seems like a great way of adding functionality to the Adafruit GFX library without having to actually modify it, should make it really easy to maintain both now.

ImpulseAdventure:
Hi @victorpv - Yes, it would be great if #461 were merged. I was intending to close this PR if/once the other PR becomes accepted (it certainly is a more desirable solution).

PR #445, opened 2018-02-16
Added Variant STM32F103VB

MackPI:
Added Board Varient STM32F103B.
Replaced #ifdef STM32_HIGH_DENSITY
with #if STM32_NR_PORTS > 4
Where it makes sense for High Pin count medium density parts like the
STM32F103VB.


victorpv:
Looks good other than the _c8 linker scripts are not needed. Those may be some left over from copying files from c8 boards.
Please remove them from your PR (flash_c8.ld and the like)

rogerclarkmelbourne:
My only slight concern is the change to

 defined(STM32_HIGH_DENSITY) || defined(STM32_XL_DENSITY)

to the number of GPIO ports.

Why was this necessary, as I'm concerned it affects other devices

MackPI:
The part I added (STM32F103VB) is a high pin count medium density part.  It didn't make sense to me to define it as a high density part because it doesn't have the high density peripherals. It does however have 5 ports. In gpio1_f1.h It seemed wrong to define the nigher ports as part of the high density peripherals.  In rcc_f1.h I moved the high density and xl density down to affect the peripherals associated with high/xl density parts. 
I needed port E defined for my project to compile.  
I only have 2 STM32F103 boards to work with, the VB variant and the CB (Blue pill) variant.
I have tested with the VB variant. I can test with the CB variant.
I am not an expert with the STM32 devices, I just made changes that made sense to the board I needed to add and hoped it would be useful to others.

I can remove the linker files @victorpv mentioned and re-commit if that makes sense.
Thank you for your consideration of my code.

victorpv:
I agree the pin count should not be dependent on the MCU being XL or High density for the same reason as @MackPI pointed out. That said, how was it resolved for the RB MCU that's already one of the variants?

MackPI:
The RB Variant doesn't have a PORT E so it doesn't need any code that was in the High density sections.

MackPI:
I have removed the ld files that @victorpv requested.

rogerclarkmelbourne:
The risk with this PR is the changes to use the number of ports rather than the "Density" constant

The only way I'm going to be able to tell if this is an issue is just to merge then wait for people with F103R or V or Z boards to complain that their project no longer works.

I guess its worth the risk

MackPI:
I tried to make the changes in a way to limit how it would affect existing projects.  I don't expect any adverse effects. Of course I can't see all the possible ways someone might have used this code.
Thank you for your consideration on this.

rogerclarkmelbourne:
No problem

I'll pull locally and do a couple of test builds, selecting the R,V,and Z boards and see if anything fails

rogerclarkmelbourne:
Did some test builds.  Seems to compile OK

Thanks....

PR #450, opened 2018-02-25
Use useradd instead of adduser

maikwoehl:
`adduser` isn't available on all linux systems. 

`usermod` is available.


rogerclarkmelbourne:
I'm not a linux expert. Can anyone confirm this works for them?

rogerclarkmelbourne:
I checked with @ahull on www.stm32duino.com and he thought the PR was OK and should be merged. 
Hence thats what I've done

PR #459, opened 2018-03-13
Some updates to the XPT2046 touchscreen library:

tfry-git:
- Add TS_Point-class and getter for better compatibility with various other touchscreen libraries out there
- Add oversampling for better accuracy. Default is seven samples per point
- Use SPI.beginTransaction() / endTransaction() to set proper SPI speed (as supported by XPT2046)

### DIFF
--- a/STM32F1/libraries/Serasidis_XPT2046_touch/src/XPT2046_touch.cpp
+++ b/STM32F1/libraries/Serasidis_XPT2046_touch/src/XPT2046_touch.cpp
@@ -8,81 +8,97 @@
 #include "Arduino.h"
 #include "XPT2046_touch.h"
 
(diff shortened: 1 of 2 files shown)


rogerclarkmelbourne:
I don't think I have this touch screen hardware so I can test this

Can you post a link to a display that uses this ?

tfry-git:
Seems to a rather popular choice among current cheap SPI ILI9341's, and that's where I got mine. E.g.: https://www.ebay.com/itm/2-4-SPI-TFT-LCD-Display-2-4-Inch-Touch-Panel-LCD-ILI9341-240x320-5V-3-3V/201950756194?hash=item2f0533fd62:g:viIAAOSw1BlZOmKM . (It's not mentionend in the specs, but you can see it in the pictures from the backside).

rogerclarkmelbourne:
OK
I think I have one of those displays.

I'll double check

rogerclarkmelbourne:
OK.

I have that display, so I'd need to test this, before I can merge

Unless anyone else wants to test it ?

PR #462, opened 2018-03-15
Update i2c.c

stevstrong:
fix for unrecognized devices having ID 0x7..

I tested in my case (ID=0x78) and it works as expected.
Very possible that will fix other unrecognized devices as well, see http://www.stm32duino.com/viewtopic.php?f=14&t=3259&start=10#p41749.


rogerclarkmelbourne:
Thanks

PR #465, opened 2018-03-20
Increase tone resolution

stevstrong:
Small change to increase tone resolution by multiplying before dividing.


victorpv:
Good catch.
Do you think the compiler optmizes away the 2/1000 to just make /500? if not, perhaps is a good idea to replace it.
Also thinking if it doesn't currently optimizes it away, perhaps by taking the 2 out of the parenthesis will allow it to see it and remove it.

stevstrong:
Maybe this will force the compiler to optimize the "*2/1000" to a single "/500".

rogerclarkmelbourne:
AFIK order of operation will be division first

I'd normally make sure that all the multiplications are done first by putting a bracket around them, otherwise you can get a loss of precision,but in this case as both 2 and 1000 are constants, making the compiler multiple by 2 and divide by 1000 is not going to cause loss of precision

Whether the compiler optimises 2/1000 to 1/500 probably depends on the optimisation settings, but my guess would be that it always optimises this to 1/500.

PR #467, opened 2018-03-23
Enable tamper output to RTC calibration

hyperion11:
add two functions.
void rtc_enable_calibration(void);
void rtc_disable_calibration(void);

useful rtc calibration. After enabling on PC13 will be found LSE/64 freq.
For check i disable internal led blinking on PC13. I do not know what will happen if enabled TAMPER output to PC13 and digitalwrite to PC13. And dont't want try to check it :)
details about calibration
http://www.st.com/resource/en/application_note/cd00167326.pdf
in Russian
http://uc.org.ru/node/27
check with logic analyser freq
work fine

image


hyperion11:
sorry for large numbers of corrections :) Still newbie in github and programming :)

rogerclarkmelbourne:
The one potential problem is this line 

RCC_BASE->BDCR |= RCC_BDCR_RTCSEL;

It will apply all the time, not just when using the calibration functions you added


What does this change do ?

hyperion11:
I remove this. Adding this was mistake, but with this string work fine :) BTW better without it.

rogerclarkmelbourne:
OK. (LOL)  

Do you mean it works better without this ?

hyperion11:
I think it works equally :)
I upload test code with this string. It work normal. But when i google this string, after your comment, i found that i not correct understand what it doing.

rogerclarkmelbourne:
OK.

Less risk if we don't have that line

BTW. Not sure why, but gitub now tells me there is merge conflict with 

STM32F1/libraries/RTClock/examples/BluePill-RTClock-test/BluePill-RTClock-test.ino

Did you change something else ?

rogerclarkmelbourne:
This now has conflicts which I don't know how to resolve

rogerclarkmelbourne:
I'll close this, but please re-open when the conflicts have been resolved

PR #468, opened 2018-03-23
Update platform.txt - Reduces line length of build output message

stevstrong:
cosmetics: remove unnecessary blank space

Reduces line length of build output message.

### DIFF
--- a/STM32F1/platform.txt
+++ b/STM32F1/platform.txt
@@ -57,7 +57,7 @@ compiler.ar.extra_flags=
 compiler.elf2hex.extra_flags=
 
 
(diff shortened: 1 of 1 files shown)


rogerclarkmelbourne:
Does this work on all platforms ?

e.g. Linux and Mac as well as PC?

stevstrong:
I only removed some pointless spaces from the line, it should work for Linux and Mac, but I haven't tried.

rogerclarkmelbourne:
OK. 
I thought perhaps you had split the line into 2 lines

victorpv:
Looked at my local copy, and I see the bunch of spaces in Windows. Not sure what character code they are, and if it will show the same in linux, but definitely showing in Windows as just spaces, this change looks good to me.

PR #474, opened 2018-03-29
Sdio patch

victorpv:
Multiple improvements and corrections to the F1 SDIO library, tested with 4bit mode now in a bluepillF4 with F103RET6 MCU.
Greatly improved sequential read and write speeds even when using 512bytes blocks, similar to improvements in the EX class in SdFat when using SPI mode, but now using the hardware SDIO interface.

### DIFF
--- a/STM32F1/cores/maple/HardwareSerial.cpp
+++ b/STM32F1/cores/maple/HardwareSerial.cpp
@@ -187,7 +187,7 @@ int HardwareSerial::availableForWrite(void)
   return 1;
 }
 
-size_t HardwareSerial::write(uint8_t ch) {
+size_t HardwareSerial::write(unsigned char ch) {
 
     usart_putc(this->usart_device, ch);
 	return 1;
--- a/STM32F1/cores/maple/HardwareSerial.cpp
+++ b/STM32F1/cores/maple/HardwareSerial.cpp
@@ -187,7 +187,7 @@ int HardwareSerial::availableForWrite(void)
   return 1;
 }
 
-size_t HardwareSerial::write(unsigned char ch) {
+size_t HardwareSerial::write(uint8_t ch) {
 
     usart_putc(this->usart_device, ch);
 	return 1;
--- a/STM32F1/cores/maple/HardwareSerial.h
+++ b/STM32F1/cores/maple/HardwareSerial.h
@@ -138,7 +138,7 @@ public:
     virtual int read(void);
     int availableForWrite(void);
     virtual void flush(void);
-    virtual size_t write(uint8_t);
+    virtual size_t write(uint8_t ch);
     inline size_t write(unsigned long n) { return write((uint8_t)n); }
     inline size_t write(long n) { return write((uint8_t)n); }
     inline size_t write(unsigned int n) { return write((uint8_t)n); }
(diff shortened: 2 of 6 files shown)


stevstrong:
HI Victor, can you please shortly explain how did you manage to improve speed?
Is it possible to port the changes to F4? Or doesn't have the F1 SDIO the speed limitation to 24MHz?

victorpv:
The F1 doesn't have the limitation, but I couldn't go faster than 24Mhz due to missing the pull up resistors. 
But the speed improvements are achieved by using the multiblock read and write commands and not sending the command to finish them after an individual read of write. Instead it keeps track of the previous LBA, and if the next read of write is sequential, it reads or writes more data without sending a new read or write command. When reading, the card will have the sector ready in its buffer and there is no latency. When writting, it will write to its buffer and the next write will go to the buffer again, and only write to the card after the sd card buffer fills up. It's similar to the sdFatEX class from Greiman, only that he does it in spi mode, and I do it in SDIO mode with DMA.
The changes should be portable to the F4, compare the code with some code compare tool and you can see the changes. It's mostly to readBlocks, writeBlocks and the sync function.
I also separate the SDIO registers setup and the DMA setup in 2 different functions, but that may not be stricly needed, and moved the order in which those are set.
I plan whenever I have more time to work on it  a bit further. I disable optimizations in one of the functions because when debugging it, looked like the compiler was removing chunks that were needed, but need to double check on that and may add the optimizations away.
Currently works great and can get about 2MB reads and writes in a class2 card with 512bytes transfers.

victorpv:
See this posts for results with a class 10 card with your test sketch (and only at 12Mhz due to lack of pullups):
http://www.stm32duino.com/viewtopic.php?f=19&t=2633&start=810#p42392

stevstrong:
OK, thanks, I'll put the port to F4 on the TODO list.

victorpv:
I may give a shot to backporting that once I get a bluepillF4 assemble with an F4 mcu, haven't had time for that yet.
If the same techniques work with the F4 SDIO and DMA, it will kick ass ;)

rogerclarkmelbourne:
@victorpv 

I've merged this, but I had to make a manual correction because of a conflict with the _io changes in another PR

Please can you test whether the SDIO functionality still works

PS. I notice in sdio.h there are both upper can lowercase __io (and __IO) in the sdio_reg_map typedef struct

I think perhaps the upper case version should be changed to lower case

rogerclarkmelbourne:
Thanks

I can do it if you want.... but you'd still then need to resync and there have been loads of other changes as I processed several PRs yesterday and 1 today

victorpv:
I'll work further in the SDIO code today or tomorrow, will pull your master first and then if I have to make changes I'll submit a new PR. I also discussed with Greiman in his SdFat repo and need to make some more changes to make compatible with both the normal sdFat and the SdFat EX class.
@stevstrong don't work on the F4 SDIO based on this yet, I have more changes coming. so no point trying to modify yours based on this yet.

stevstrong:
OK, I am currently trying to port your SPI extended DMA function to F4, no work on SDIO.

stevstrong:
Very old thread, but I just want to ask @victorpv whether the SDIO lib in this repo is functional or it still needs some tweaks.
I am currently upgrading to SdFat V2 and I have to rework the SDIO for F4. I thought it would be a good occasion to merge these improvements from F1, in case it is stable.

stevstrong:
@victorpv, it would be nice if you could let me know if you are still involved in this issue.

PR #476, opened 2018-03-29
16 mhz support

victorpv:
Adds a way to use 16Mhz crystals thru a menu option in boards.txt and small changes to the source files.
Only 1 file changed in the core, to add the register setting for HSE/2, which was not present in the core.
Then a change to the variant files to apply that or not depending on the macros passed at compilation time, thanks to the boards.txt option.
I can't see where the change to .gitignore comes from, so I can't see how to revert it.


victorpv:
Cancelling this PR. will try to resubmit cleaner.

BlackBrix:
is it a conflict with this PR: https://github.com/rogerclarkmelbourne/Arduino_STM32/pull/424  ?

victorpv:
No conflict because they do it in a slightly different way, so both pieces of code can be there, but after seeing your code to do the same, I think we may be able to find even a better way modifying 1 single file in the core, and the way I used to define the macro in the boards.txt file, so we don't have to modify each variant file, which both your way and my way currently did. Will have a look at it later.

victorpv:
@BlackBrix 
I just tested if I add your changes to boards.cpp, and then add this in boards.txt I can get the 16Mhz XTAL option selected thru menus with the minimal changes:
genericSTM32F103R.menu.cpu_speed.speed_16m_crystal=72Mhz with 16Mhz HSE crystal
genericSTM32F103R.menu.cpu_speed.speed_16m_crystal.build.f_cpu=72000000L
genericSTM32F103R.menu.cpu_speed.speed_16m_crystal.build.hs_flag=-DXTAL16M

If you want to give it a shot and confirm all is good with the your Vx board, we can submit a new PR to add it to the menus.

victorpv:
I think at least all F1 mcus support that
I tested with an 103R. I'm thinking it there can be a common place to add that line other than in variant, so we modify that only once in the core, and then if someone needs to use all he need to do is add the menu lines to the board he wants to use it too.
If we keep it in the boards.cpp file, we would need to modify each variant.

BlackBrix:
OK, that sounds good !

victorpv:
Just tested we can make your change in the core in rcc_F1.c file, and then we don't need to modify each variant file.
If we insert this in line 137:
```
#ifdef XTAL16M
    // 16MHz crystal (HSE)
    // in this case we additionally set the Bit 17 (PLLXTPRE=1)  =>  then HSE clock is divided by 2 before PLL entry
    cfgr |= RCC_CFGR_PLLXTPRE;
#endif
```

Then for any board to have that option, we just need the menu lines I mentioned:
```
genericSTM32F103R.menu.cpu_speed.speed_16m_crystal=72Mhz with 16Mhz HSE crystal
genericSTM32F103R.menu.cpu_speed.speed_16m_crystal.build.f_cpu=72000000L
genericSTM32F103R.menu.cpu_speed.speed_16m_crystal.build.hs_flag=-DXTAL16M
```

Could even be a separate option, so one could chose 48Mhz speed, and 16Mhz base crystal. I'll send a new PR modifying this.
Could you test in your board to comment out the lines you added in the variant/wirish/board.cpp file, and add the lines above in the rcc_F1.c, and finally add a menu item for your board?
It works fine in my RET board.

rogerclarkmelbourne:
Victor

Just by coincidence I was looking through the old PRs and I just added the PLL multiplier switching code to all variants to select the correct multiplier if F_CPU is 16MHz

And I'm about to look at the option for 8Mhz by routing HSE direct rather than via the PLL
(though I need to change ever variant for this as well)

I'm not going to add boards.txt options for these new speeds as Boards.txt is  complex enough already and if anyone really wants to use these speeds they can modify their own local files

rogerclarkmelbourne:
@victorpv 

I've used your suggestion and updated rcc_f1.c

See commit
 
https://github.com/rogerclarkmelbourne/Arduino_STM32/commit/4f691b9eebfcd63fa6f918cc02171e127ac9f246

I think this PR can no be considered to have been actioned

BlackBrix:
sorry, I lost the overview somehow ... ;-)
did I get it right? -> so we WANT to use the modified rcc_f1.c now, but NOT the menu options

If yes,
do we have to remove/revert the Pull https://github.com/rogerclarkmelbourne/Arduino_STM32/pull/424/files
which modified the "boards.cpp" ?

rogerclarkmelbourne:
@BlackBrix 
No need to do anything.. ;-)

I've implemented @victorpv 's suggestion

Anyone who has a 16Mhz crystal on their board would need to modify their own board.txt to add -DXTAL16M to the list of defines.

There are so few boards that use this, it would be very confusing for the majority of users to add this as an option on their board e.g. BluePill as its not an option they could select (and it still work)

victorpv:
#424 should be reverted, since they both acomplish the same, but #424 only did in the variant for a single board and the change in rcc_F1.c allows it to work for every variant including any future one.

About the menu options, I agree with Roger it's best to leave them out, and I'll add an entry in the wiki explaining how to add it, so users can add it in their own local copies if they need.

PR #481, opened 2018-04-06
Update to fix FSMC inclusion in Rx devices

victorpv:
The FSMC is only available in high density and XL density devices with more than 4 ports.
Basically excluded in the RC/D/E/F/G models, since they dont have enough pins.
Page 11 in the Datasheet references this. The table shows what is present, and shows the FSMC is NOT present in Rx MCUs:
http://www.st.com/content/ccc/resource/technical/document/datasheet/59/f6/fa/84/20/4e/4c/59/CD00191185.pdf/files/CD00191185.pdf/jcr:content/translations/en.CD00191185.pdf

Recent changes to other files highlight this issue during compilation time. The issue was already there, but GPIO E and GPIO F were declared and defined, while they don't actually exist in Rx MCUs, so FSMC was being included for those MCU even though they don't actually have an FSMC device, or at least no external ports to use it. 
With the recent changes GPIO E and F and correctly excluded from the build, but since FSMC was still included and needs GPIO E and F, it was failing to compile. I.e.:
C:/Users/Victor/Documents/Arduino/Hardware/Arduino_STM32/STM32F1/cores/maple/libmaple/fsmc_f1.c:51:19: error: 'GPIOE' undeclared (first use in this function)

Since the FSMC is not available (I suspect the silicon is there, but there is no IO pins to use it), the correct solution is to include FSMC only when using an MCU with more than 4 GPIO ports, which is what matches the datasheet.

### DIFF
--- a/STM32F1/system/libmaple/stm32f1/include/series/stm32.h
+++ b/STM32F1/system/libmaple/stm32f1/include/series/stm32.h
@@ -187,6 +187,8 @@ extern "C" {
 #       define STM32_TIMER_MASK         0x1FE /* TIMER1--TIMER8 */
 #       if STM32_NR_GPIO_PORTS > 4
 #           define STM32_HAVE_FSMC      1
+#       else
+#           define STM32_HAVE_FSMC      0
 #       endif
 #       define STM32_HAVE_DAC           1
 #    elif defined(STM32_XL_DENSITY)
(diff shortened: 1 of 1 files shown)


victorpv:
This is related to https://github.com/rogerclarkmelbourne/Arduino_STM32/pull/445
It was not caused by that PR, but rather that PR highlighted something we had wrong, we were included GPIOE, GPIOF and FSMC in all high density and XL density devices, while in reality the R versions (64 pins MCUs) do not have GPIOE/F or FSMC, as per the datasheet linked above.
I think this correction is urgent otherwise compilation to RC/D/E MCUs will fail since FSMC is included and GPIOE/F are not. 
PR 445 removed the GPIO E/F inclusion, but not the FSMC inclusion. Now this PR removes FSMC from the MCUs that do not have it (High/XL Density but with less than 5 GPIO Ports).
I'm already testing this locally since I use an RCT6 MCU in some bluepillF4 boards.
Will add if I need to make any other change, but with this chance compilation is successful again.

rogerclarkmelbourne:
Thanks

Issue #487, opened 2018-04-10
Bug in PR #441


edogaldo:
As per topic http://stm32duino.com/viewtopic.php?f=2&t=3480&sid=4ce86bcf8df7ba9701ae6e4f269cbb58, PR #441 introduces a bug: it modifies function "rtc_set_count()" instead of function "rtc_get_count()" as declared.

KenWillmott:
Why "instead of"? It seems to me like the get method also must use a non-atomic access.

edogaldo:
In your first comment of #441 you said "_Modify **rtc_get_count()** and rtc_get_divider() to fix non-atomic accesses to RTC counter registers_" but then in your code you modified **rtc_set_count()**.
Furthermore in **rtc_set_count()** you are not updating the registers anymore.
This was not intended behaviour, correct?

edogaldo:
Fixed with #488 merge

PR #488, opened 2018-04-10
fix PR #487

KenWillmott:
Implements the needed correction in PR #487, modifies get method with no change to set method. Exor comparison changed to equality comparison for readability.

### DIFF
--- a/STM32F1/libraries/RTClock/src/utility/rtc_util.c
+++ b/STM32F1/libraries/RTClock/src/utility/rtc_util.c
@@ -169,8 +169,8 @@ uint32 rtc_get_count() {
 	rtc_wait_sync();
 	rtc_wait_finished();
 	do {
-	h = RTC->regs->CNTH & 0xffff;
-	l = RTC->regs->CNTL & 0xffff;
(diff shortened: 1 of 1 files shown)


edogaldo:
Seems ok to me now, thanks!!

victorpv:
why this change?
https://github.com/rogerclarkmelbourne/Arduino_STM32/pull/488/files#diff-2f78cb1efe5eaf4b0b9f469372a9a044L219

rogerclarkmelbourne:
I guess this is still pending a response ??

edogaldo:
a XOR b 
and 
a != b 
are logically equivalent so I can't see any issue in this change

rogerclarkmelbourne:
Thanks...

I did my usual basic test, compile with my current project and my project still works.
(My project does not use RTC, so this is really just to test it doesn't break mainstream functionality)

victorpv:
@edogaldo I thought they should be equivalent, that's why I asked what's that change is for.

edogaldo:
He said it in the first post ;)
> [...] Exor comparison changed to equality comparison for readability.

PR #489, opened 2018-04-11
With these modify the core compile all stm32F103 variants

bigjohnson:
Added a define and copied a ld script

### DIFF
--- a/STM32F1/system/libmaple/include/libmaple/libmaple_types.h
+++ b/STM32F1/system/libmaple/include/libmaple/libmaple_types.h
@@ -53,6 +53,7 @@ typedef void (*voidFuncPtr)(void);
 typedef void (*voidArgumentFuncPtr)(void *);
 
 #define __IO volatile
+#define __io volatile
 #define __attr_flash __attribute__((section (".USER_FLASH")))
 #define __packed __attribute__((__packed__))
 #define __deprecated __attribute__((__deprecated__))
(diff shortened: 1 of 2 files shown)


rogerclarkmelbourne:
I'm not seeing any problems with the problem of some things being declared as __io and some as __IO

So I checked and everything in the core seems to be defined as __IO    Upper case

I'll manually merge the change to add the missing linker file, as I'd noticed that the F103V is no longer linking, but I need clarification about the __io change and whether its still needed ??

rogerclarkmelbourne:
I've pushed a commit to address the missing linker file issue, using the file from your PR

https://github.com/rogerclarkmelbourne/Arduino_STM32/commit/df96076900320869328302aacff8bb79a35a8198

Please let me know if the other change is still needed and if so how I can reproduce the error that it fixes

Thanks

edogaldo:
The problem of "__io" errors was due to the change in "STM32F1/system/libmaple/include/libmaple/libmaple_types.h" of 
`#define __io volatile`
in
`#define __IO volatile`
so now there's no more definition for "__io".

The remaining "__io" references should anyway have been fixed in PR #493 which has just been merged.

victorpv:
And the whole purpose of changing __io to __IO is to remove __io from the core because it was used by some code somewhere, and Arduino and CMSIS used __IO.
So if we added a new definition for __io in the core all the work changing to __IO would be in vain.

PR #492, opened 2018-04-13
usbcomposite library

arpruss:
Add now stable USBComposite (HID/Midi/Mass Storage) library.


rogerclarkmelbourne:
I've merged this, but the USB Mass storage example doesn't compile on the maple mini

The error message seems to suggest it does not have enough memory, probably because I'm using an original Maple mini bootloader, which means the sketch has less memory.

But as it works fine on the BluePill , I think its fine.

edogaldo:
Hi all, in the library there are files "usb_serial.h" and "usb_serial.c" which could conflict with the corresponding Core versions, what about changing their names?

arpruss:
Is that a problem since they are in a different directory? I've never had trouble from this in compiling.

edogaldo:
Well, I got a conflic problem while trying to transform the usb_serial into a library in my personal core branch.
Of course I know this is not a supported scenario for the core usage but it's an indicator of possible conflics which could happen also in other scenarios..
Having different file names would prevent any risk of conflicts at all.

edogaldo:
Consider you reference both with 
`#include <usb_serial.h>`

arpruss:
That's a slip-up in the library's usb_setup.cpp. I actually think the include isn't needed at all in usb_setup.cpp. Elsewhere in the library, it's:
`#include "usb_serial.h"`

edogaldo:
I mean, anybody which for any reason, would like to
`#include <usb_serial.h>`
in their sketch, would run into the conflict..

arpruss:
They probably shouldn't do that, since usb_serial doesn't expose functionality that a sketch-writer should be using directly. But the point is taken and I am renaming in the upstream library.

edogaldo:
This code tries to access USBHID_CDCACM_SET_LINE_CODING defined in your library, can it make sense?
If yes, it won't compile..
```
#include <usb_serial.h>

int a=USBHID_CDCACM_SET_LINE_CODING;

void setup() {

}

void loop() {

}
```

edogaldo:
Thanks!!!

edogaldo:
@roger: would it be fine for you to accept the change?

PR #493, opened 2018-04-13
Multiple updates to SDIO library

victorpv:
Changes __io for __IO
Increases the time for some delays
Adds compatibility with SdSdioEX class while using DMA
Implement a retry mechanism (needs further testing)
Implement command ACMD42 to disconnect the sd internal pullup in D3 in 4bit mode
Some other small changes to increased reliability in bluepillF4 with F1 mcu

### DIFF
--- a/STM32F1/libraries/SDIO/SdioF1.cpp
+++ b/STM32F1/libraries/SDIO/SdioF1.cpp
@@ -30,7 +30,7 @@
 //==============================================================================
 //#define SDHC_PROCTL_DTW_4BIT 0x01
 #define CMD8_RETRIES 10
-#define BUSY_TIMEOUT_MILLIS 1500
+#define BUSY_TIMEOUT_MILLIS 1000
 //==============================================================================
 #define CMD_RESP_NONE SDIO_CMD_WAIT_NO_RESP
 #define CMD_RESP_R1   SDIO_CMD_WAIT_SHORT_RESP // normal response
@@ -61,6 +61,12 @@
 #define CMD38_XFERTYP   (uint16_t)( CMD38  | CMD_RESP_R1b  )
 #define ACMD41_XFERTYP  (uint16_t)( ACMD41 | CMD_RESP_R3   )
 
+/*
+ * ACMD42 to enable disable CD/D3 pull up. Needed for 4bit mode.
+ */
+const uint8_t ACMD42 = 0X2A;
+#define ACMD42_XFERTYP  (uint16_t)( ACMD41 | CMD_RESP_R1   )
+
 #define CMD55_XFERTYP   (uint16_t)( CMD55  | CMD_RESP_R1   )
 
 //=============================================================================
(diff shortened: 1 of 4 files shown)


rogerclarkmelbourne:
Thanks

Seems to have fixed some compile issues on F103Z as well !

stevstrong:
@victorpv , I am porting your differences to F4 and I have some doubts.
Specially about dmaTrxPrepare(), in which my DMA setup differs a lot from yours, it seems that you are not using the FIFO and peripheral control?
Also, the number to transfer in you version is only half of that what I use.
Could you eventually comment these?

STM32F1/libraries/SDIO/SdioF1.cpp
```diff
@@ -573,52 +701,157 @@ uint32_t SdioCard::kHzSdClk() {
...
+#if USE_DEBUG_MODE
+      Serial.print("New lba, syncing :");
+      Serial.println(lba);
+#endif
+      _state = syncBlocks();
+      DBG_PRINT();
+      if (!_state) {
+        return false;
+      }
+      m_limitLba = (lba + 1024); //arbitrary limit, tested with 32KB before and worked fine.
```
stevstrong: Wouldn't this cause some buffer overflow if the buffer is only 512 bytes large?

victorpv: Sorry Steve, I have been away from Github for quite a while.
The modifications I made to the library use the commands to read and write multipleblocks, and keep accessing blocks sequentially, until it finds a block is requested that's not sequential to the previous one. The limit part is from the teensy. One of the teensy MCUs actually has a bug in which it can not manage more an IO or more than NNN blocks. It's not our case, so I changed the number and bumped it to 1024 blocks.
What that flag does is just ensure that if you are doing sequential IO, it sends the command to complete the IO and start a new one every 1024 blocks at max. We could actually take it out, as I mentioned I tried up to 32k blocks and still worked fine.
Those are blocks (512 bytes each). The buffer size doesn't have an impact on that, is just about how many sequential blocks are accessed before the MCU tells the SD card to complete the previous IO and start a new one.

stevstrong:
Victor, thanks for taking time to answer.
Any comment on the defined but not used variables?

victorpv:
The not used variables are most likely left overs from modifying the code and eventually removing parts, and forgetting to clean up those variables.
About the DMA setup, the F1 is very different. It doesn't use a FIFO, and the way it calculates the number of bytes to transfer is different, so results in having the load the counter with a value that's half of what you use in the F4.
You shouldn't need to change much to your DMA code. Once I backported your F4 library to F1 is when I made the DMA changes to F1, but adding the multiblock and SDIOEX class support required little change or none to the DMA, it was mostly around how it manages the commands to start and stop reading.

stevstrong:
Thanks for clarification, I will give the porting another try.

PR #491, opened 2018-04-12
Update timer.h

stevstrong:
fix for https://github.com/rogerclarkmelbourne/Arduino_STM32/commit/dfb5bd4c30f42fa9cf2a0aef6b7d2a50a594fa97, see #412

PR #497, opened 2018-04-13
Update timer.h

stevstrong:
fix compile error for #491


stevstrong:
I have tested now and it seems to work.

rogerclarkmelbourne:
This looks like its superseded

edogaldo:
Sorry @rogerclarkmelbourne and @stevstrong, I'm missing why this should be superseded, clould you please explain?

Thanks in advance, E.

stevstrong:
Roger, it seems that the fix is still not merged in the core.
Edogaldo has merged its own version in his own repo, which version is similar to mine.
But none of them was merged here.

rogerclarkmelbourne:
OK

If this PR is still valid I will merge it

edogaldo:
Well, this PR is more than 1 month old so day +/day -... :)
Anyway I think this is a required fix..

Issue #486, opened 2018-04-09
Optimize Print::write(const char *str)


vshymanskyy:
https://github.com/rogerclarkmelbourne/Arduino_STM32/blob/596dc0d78e27fc00e9a48f37e583f29d612b67f7/STM32F1/cores/maple/Print.cpp#L49
This should really call `strlen()` and then call `Print::write(const void *buffer, uint32 size)` implementation.
Why so? Because the `write(buffer, size)` is commonly overridden and implemented in an optimized manner (sending data to the wifi/gsm module in chunks, not byte by byte...).

vshymanskyy:
Well, virtually every library out there expects Arduino Core to behave this way ;)
I am not asking for bounds checks, I'm asking to call a different overloaded/overridden function.
Check out this: https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/cores/arduino/Print.h#L52

stevstrong:
I think that this would make sense, even if at the end some of the overloaded functions end up with the same single writes.
I would go further and move these functions in a header file as Arduino core does.

rogerclarkmelbourne:
Ah,

I thought you were proposing bounds checking.. My mistake...

stevstrong:
Victor, take a look at the very first link of OP.

vshymanskyy:
Thanks! BTW is it only related to F1 or tp other cores as well? Didn't check them

rogerclarkmelbourne:
Good point, I should update the F4 and perhaps the F3

Note. The F3 is really not supported, there arent any F3 boards in the boards.txt, its just there as an archive, but I should probably remove it, because git will keep a record of it in the older commits

rogerclarkmelbourne:
Fixed

PR #501, opened 2018-04-17
F4: update Print.cpp

stevstrong:
port from #486.

Tested.


rogerclarkmelbourne:
Thanks Steve

I had intended to do that one myself, but I got side tracked :-(

PR #502, opened 2018-04-17
Allow IRQ overriding

edogaldo:
As per post http://stm32duino.com/viewtopic.php?f=28&t=3467&start=10#p44025


stevstrong:
Hm, if we really want to do it in a consequent way, then the .S file should be actually removed, and all IRQs defined as weak in the appropriate .c files. Or am I missing something?

edogaldo:
I think the .S file makes sense to map all the undefined exceptions and IRQs to the default handler.

rogerclarkmelbourne:
I will try to get to this today

PR #503, opened 2018-04-20
Add simple SPI slave example

VictorLamoine:
A very simple SPI slave example, I think the project lacks simple examples like this; it helps a lot getting things running!

### DIFF
--- a/STM32F1/libraries/SPI/examples/spi_slave/spi_slave.ino
+++ b/STM32F1/libraries/SPI/examples/spi_slave/spi_slave.ino
@@ -9,11 +9,9 @@
 
 void setupSPI(void)
 {
-  // MOSI, MISO, SCK PINs are set by the library
-  pinMode(BOARD_SPI_DEFAULT_SS, INPUT); // SS
-
   // The clock value is not used
   // SPI1 is selected by default
+  // MOSI, MISO, SCK and NSS PINs are set by the library
   SPI.beginTransactionSlave(SPISettings(18000000, MSBFIRST, SPI_MODE0, DATA_SIZE_8BIT));
 }
 
--- a/STM32F1/libraries/SPI/examples/spi_slave/spi_slave.ino
+++ b/STM32F1/libraries/SPI/examples/spi_slave/spi_slave.ino
@@ -29,7 +29,7 @@ void loop()
 {
   // Blocking call to read SPI message
   uint8_t msg = SPI.transfer(++count);
-  Serial.print("a Received = 0b");
+  Serial.print("Received = 0b");
   Serial.print(msg, BIN);
   Serial.print(", 0x");
   Serial.print(msg, HEX);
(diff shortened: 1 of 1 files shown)


stevstrong:
Two comments:
- the SCLK, MISO and MOSI pin modes are already setup in the core.
- loop: wouldn't be more appropriate to use
`uint8_t msg = SPI.transfer(data);`
This replaces pushing the cnt value into DR and the read.

victorpv:
Also as mentioned in the forum, setModule(1) is not needed since the SPI object uses SPI1 by default. 
Having setModule there may induce people to think it's needed in any case.

VictorLamoine:
I have updated the code with your suggestions

stevstrong:
After a more closer look into the core code, it seems that the NSS pin is also setup correctly by the core for slave mode, see: https://github.com/rogerclarkmelbourne/Arduino_STM32/blob/1175f37c379513fbfc17bab75af4297d6f89e09f/STM32F1/cores/maple/libmaple/spi_f1.c#L71.
So the corresponding line in setup example can be also removed.

VictorLamoine:
Ah cool, pull request updated!

rogerclarkmelbourne:
Awaiting response to @stevstrong question about DMA enabling before things have been fully setup

stevstrong:
I have to setup a master-slave configuration to test this, hope I can do it today.

VictorLamoine:
> Awaiting response to @stevstrong question about DMA enabling before things have been fully setup
This example is without DMA?! I don't understand what you mean

I'll make an other pull request for the full duplex SPI with DMA slave.

stevstrong:
Successfully tested the attached salve with the attached master.
I added to the slave a dummy SPI.read() to pre-empty the Rx pipe.

[STM32_SPI_master.ino.txt](https://github.com/rogerclarkmelbourne/Arduino_STM32/files/1958618/STM32_SPI_master.ino.txt)
[STM32_SPI_slave.ino.txt](https://github.com/rogerclarkmelbourne/Arduino_STM32/files/1958619/STM32_SPI_slave.ino.txt)

Would it make sense to add the master code to the slave as comment block?
I will also make a master and slave example for SPI with DMA.

kvaale:
I get this error when trying to compile:
`'class SPIClass' has no member named 'beginTransactionSlave'; did you mean 'beginTransaction'?
`

along with:
`'DATA_SIZE_8BIT' was not declared in this scope; did you mean 'SPI_DATASIZE_8BIT'?`
...from the argument passed to the `SPI.beginTransactionSlave`-method.

I checked the commits on [SPI.h](https://github.com/stm32duino/Arduino_Core_STM32/blob/main/libraries/SPI/src/SPI.h) but didn't find any changes regarding this `SPI.beginTransactionSlave`-method.
Am I missing something obvious? If so: sorry, I am noob.

PR #508, opened 2018-04-25
F1 : Update Wire.cpp

stevstrong:
bugfix hw I2C unrecognized addresses 0x76, 0x77.

As discussed in http://www.stm32duino.com/viewtopic.php?f=14&t=3259&p=44458#p44392.
Tested with i2c scanner sketch.


rogerclarkmelbourne:
Thanks.

PR #509, opened 2018-04-27
Fixed missing return value.

GitMoDu:
Overloaded method was not returning the result.


rogerclarkmelbourne:
Thanks.

Looks like a mistake

However I'll need to test this to confirm there are no side effects of fixing it.

rogerclarkmelbourne:
Thanks

Tested with a BMP280 using the I2C scanner and also the Adafruit BMP280 lib

Both worked fine.

GitMoDu:
I'll test some other I2C devices in the meanwhile.

rogerclarkmelbourne:
I think its fine.

Its now merged into the core.

I'm sure people will let me know if they have problems with it ;-)

PR #515, opened 2018-05-12
add void transfer(void *buf, size_t count)

naughtyStark:
the arduino SPI library contains a function transfer(void *buf, size_t count). Just increasing portability.

### DIFF
--- a/STM32F1/libraries/SPI/src/SPI.cpp
+++ b/STM32F1/libraries/SPI/src/SPI.cpp
@@ -385,13 +385,13 @@ uint16_t SPIClass::transfer16(uint16_t data) const
     return ret;
 }
 
-void SPIClass::transfer(uint8_t *buf, size_t count) 
+void SPIClass::transfer(uint8_t *buf, size_t count) //implementation written by stevstrong. 
  {
     if (count == 0) return;
     spi_rx_reg(_currentSetting->spi_d);      // clear the RX buffer in case a byte is waiting on it.
     spi_reg_map * regs = _currentSetting->spi_d->regs;
     // start sequence: write byte 0
-    regs->DR = *buf;  
+    regs->DR = *buf;  //transmit first byte
     while (--count > 0) 
     {
         while( !(regs->SR & SPI_SR_TXE) );   // wait for TXE flag
@@ -403,7 +403,7 @@ void SPIClass::transfer(uint8_t *buf, size_t count)
     }
     // read remaining last byte
     while ( !(regs->SR & SPI_SR_RXNE) );     // wait till data is available in the Rx register
-    *buf++ = (uint8)(regs->DR); 
+    *buf++ = (uint8)(regs->DR);  //read and store the last byte. 
 }
 
 /*  Roger Clark and Victor Perez, 2015
--- a/STM32F1/libraries/SPI/src/SPI.cpp
+++ b/STM32F1/libraries/SPI/src/SPI.cpp
@@ -385,6 +385,26 @@ uint16_t SPIClass::transfer16(uint16_t data) const
     return ret;
 }
 
+void SPIClass::transfer(void *buf, size_t count) 
+ {
+    if (count == 0) return;
+    spi_dev * spi_d = _currentSetting->spi_d;
+    uint8 *p = (uint8_t *)buf;
+    spi_tx_reg(spi_d, *p); //transmitting first byte
+    while (--count > 0) 
+    {//if you wanted to transfer 4 things, it will go 3,2,1 and when it hits 9 the loop is exited , which is why the first byte is sent outside the loop
+        uint8 out = *(p + 1); //setting the thing to send(is generally 0)
+        while (spi_is_tx_empty(spi_d) == 0); // wait until TXE=1
+        while (spi_is_busy(spi_d) != 0);     // wait until BSY=0 .waiting for clearance(from any last transmission or reception)
+        uint8 in = spi_rx_reg(spi_d); // reading SPI data register 
+        spi_tx_reg(spi_d, out); // transmitting stuff
+        *p++ = in; // placing the data in the buffer and incrementing address by 1.
+    }
+    while (spi_is_tx_empty(spi_d) == 0); // wait until TXE=1
+    while (spi_is_busy(spi_d) != 0);     // wait until BSY=0
+    *p = spi_rx_reg(spi_d);//get the last byte in 
+}
+
 /*  Roger Clark and Victor Perez, 2015
 *	Performs a DMA SPI transfer with at least a receive buffer.
 *	If a TX buffer is not provided, FF is sent over and over for the lenght of the transfer. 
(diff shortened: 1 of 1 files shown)


victorpv:
The send() function already reads and writes to the buffer:
https://github.com/naughtyStark/Arduino_STM32/blob/a5ed45017f3eec50ffbd1ed7ed0c0f99dfca905c/STM32F1/libraries/SPI/src/SPI.cpp#L312

Instead of adding a new function, can't we just make it an alias of send()?

stevstrong:
I implemented [here](https://github.com/stevstrong/Arduino_STM32/commit/9e543c993ecf73ae47b6632d6f627efd0d1869ea#diff-48b5f6ab32a24d9cd0835c829b41c629R523) some transfer function variants for both 8 bit and 16 bit mode.

victorpv:
@stevstrong the Arduino API has a transfer function that sends and receive from the same buffer. As @naughtyStark suggested adding that will improve compatibility with existing libraries. Our read() function does the same and takes the same parameters, and you optimized it a while back. 
What do you think on making transfer() and alias, or otherwise making it an inline call to read()?

naughtyStark:
Might be a little un-related to the pull request but is definitely related to SPI. I used an oscilloscope to debug my code (using both the read and the transfer function) and i found that for some reason the MOSI pin wasn't doing it's job sometimes(sometimes it would work and then sometimes it wouldn't which was really weird). Is there something I can do in the code to make sure that the pins are actually doing their job?(The code will be used for a multirotor so i can't be using un-reliable code :P)

stevstrong:
Please use the forum for issues.
This is the place where bugs or new features are discussed.

Now regarding the read() function, I totally removed from my repo, because the new transfer functions superseded it. Main reason for that is that the read() will anyway send in a way or other some data, so now in the transfer() function you can specify the data to be sent.

STM32F1/libraries/SPI/src/SPI.cpp
```diff
@@ -385,6 +385,26 @@ uint16_t SPIClass::transfer16(uint16_t data) const
     return ret;
 }
 
+void SPIClass::transfer(void *buf, size_t count) 
+ {
+    if (count == 0) return;
+    spi_dev * spi_d = _currentSetting->spi_d;
+    uint8 *p = (uint8_t *)buf;
```
stevstrong: This new pointer makes no sense, it is better to specify the buf parameter as (uint8_t *), as it can only transfer 8 bit data anyway.

naughtyStark: 1)You're right it doesn't make any sense. It is like that because when i was trying to make my ADNS3080 work i was in a bit of a hurry and i basically copied the code from the spi.cpp of the original arduino. 
2)I thought you guys were gonna just alias the read() function. I tried the read() function and it works just fine(maybe better but my oscilloscope didn't show any difference). Plus this code is inefficient anyway and you guys pointed that out already so i m not really sure why you included my code into the main stm32 branch.

naughtyStark:
so I conducted a test today, I used the read function and the transfer(mine, I took your suggestion and specified the buf as uint8_t pointer) function and it turned out that i had commented out some if conditions(in the code that read the sensor) that would check if the data was even valid(i was more concerned with just getting anything from the sensors in the first place then). 
So when i uncommented those if conditions and used the "read" function, it didn't work anymore(no data because no data validity. it checks data validity by seeing whether a particular(first) byte in the received data is a x01 or a x00) and when i used the transfer function it worked just fine. 
So I looked into the difference between the 2 definitions and it turns out that "read" just transmits 0xFF while the transfer function transmits whatever is passed to the function. It appears that the sensor cares about what you send to it.
So what should I do here? create another pull request with the new definition or can you guys do something?

stevstrong:
You can change your file in pull request.
I suggest you implement it as I did here: https://github.com/stevstrong/Arduino_STM32/blob/master/STM32F1/libraries/SPI/src/SPI.cpp#L533-L552

naughtyStark:
@stevstrong I saw your implementation. and it appears to be doing the following: 
transmit first byte outside the loop
enter while loop
      1)wait for txe flag
      2)transmit another byte //here is my problem. you already sent one byte, why send another? 
      3)wait for rxe flag
      4)read the received byte.
wait for rxe flag //these 2 lines appear redundant because you didn't transmit any byte corresponding
read the data on  //to the bytes that you expect to receive. 

shouldn't you read the received byte first and then transmit the next byte?(as shown-)
transmit the first byte
enter while loop
      1)wait for the txe flag
      2)wait for rxe flag
      3)read received data
      4)transmit data 
wait for txe flag
wait for rxe flag 
get the last byte. 

Also, i checked. The definition for the "read" function which I tested is pretty much the same as your definition for transfer(same order of doing things) and it gave me errors (while my implementation of transfer worked with no errors). 

So right now I am simply changing the way in which i do things(using data registers instead of using functions and toggling interrupts as you have done) but keeping the order the same.

naughtyStark:
I see. Shouldn't all the bytes be transmitted from within the while loop then? You implementation is optimized, true, but the last 2 lines are kind of redundant because you're not transferring anything but expecting to receive something. This wouldn't work if the sensor expects a value to be transmitted to it in order for it to send something back.

stevstrong:
If you carefully follow the figure, you will notice that nothing is redundant there, everything is needed as implemented. It was tested ad it works as expected.

naughtyStark:
I tried using your implementation but it gave me errors. I will check again though(next weekend). Would the following implementation work?  

void SPIClass::transfer(uint8_t *buf, size_t count) 
 {
    if (count == 0) return;
    spi_rx_reg(_currentSetting->spi_d);      // clear the RX buffer in case a byte is waiting on it.
    spi_reg_map * regs = _currentSetting->spi_d->regs;

    while ((count--)) 
    {
        while( !(regs->SR & SPI_SR_TXE) );  // wait for TXE flag
        noInterrupts();                     //go atomic level - avoid any interrupts while receiving data.
        regs->DR = *buf;                    //write next data item into the DR register. This clears the TXE flag. 
        while ( !(regs->SR & SPI_SR_RXNE) );//wait till data is available on the DR register.
        *buf = (uint8_t)(regs->DR);       //read and store the received byte. This clears the RXNE flag                    
        buf++;
        interrupts();                       //let systick do it's job
    }
    while( !(regs->SR & SPI_SR_TXE) ); // wait until TXE=1
    noInterrupts();                     //go atomic level - avoid any interrupts while receiving data.
    while ( !(regs->SR & SPI_SR_RXNE) );//wait till data is available on the DR register.
    interrupts();                       //re-enable interrupts.
    *buf = spi_rx_reg(spi_d);           //get the last byte in 
}

stevstrong:
You don't need `noInterrupts()` and `interrupts()` outside the loop. Just stick to my version **as it is** and it should work.

naughtyStark:
I have done as you said. I will test this by next weekend.

stevstrong:
Resolved by https://github.com/rogerclarkmelbourne/Arduino_STM32/commit/73e264474ed3a3d4f4ee4c9497d394510509ab1f.

PR #516, opened 2018-05-12
adding transfer(void *buf,size_t count) as member

naughtyStark:
adding the name of the member function corresponding to the function i added in the SPI.cpp


victorpv:
See comment on https://github.com/rogerclarkmelbourne/Arduino_STM32/pull/515

stevstrong:
Overruled by #586 .

PR #517, opened 2018-05-14
Add frequency parameter to hardware i2c

jjbubka:
Tested with Naze32 (STM32F103C8T6) @72Mhz & BMP280 working at 1.2Mhz
It doesn't break existing code

Example :
..
TwoWire i2c(I2C_PORT, 0, 800000); //frequency in Hz
..
Naze32 @120Mhz works up to 1.4Mhz
Naze32 @72Mhz works up to 1.2Mhz
Naze32 @24Mhz works up to 800Khz


victorpv:
@jjbubka I also tested operation up to 1.2Mhz fine at 72Mhz. I never got around to submit a PR for my version, glad to see you did.
I only have one concern, why did you remove this section?
https://github.com/rogerclarkmelbourne/Arduino_STM32/pull/517/files#diff-45915b50dee325ec7a022d2c1bef070fL90

I haven't looked at the i2c peripheral in a while, but if I remember right that's a flag for the hardware CR, needed when going faster than 100Khz. Shouldn't we still set or clear that flag depending on the speed selected?

jjbubka:
I have been reading a lot in this github and in the forum for a long time. And i think i was in debt to do something useful. :)

Well, regarding to the flag i2c_fast_mode, is used only in libmaple ambit (instead of using new frequency variable). The flag that is needed in the register is ccr |= I2C_CCR_FS, that is applied when frequency > 100khz in function set_ccr_trise. So in this case is no longer needed, but is taked into account only for existing code compatibility.
Regards,
Jose

victorpv:
Thanks for the explanation. I'll try to test this version with some fram as soon as a I have a chance.

rogerclarkmelbourne:
@victorpv 

Did you every get time to test this ?

stevstrong:
I could eventually commit this, but the two SPI files should not be changed in my opinion.
Can you please remove the changes (a11af7a) to those two files form the PR?

stevstrong:
Overruled by #675 .

jjbubka:
Closed by #675

PR #519, opened 2018-05-17
Linux and MacOSX for HID bootloader

Serasidis:


rogerclarkmelbourne:
There seem to be unrelated changes to files not associated with the HID bootloader

e.g.
tools/macosx/maple_upload

tools/linux/maple_upload

and what looks like formatting changes to

STM32F1/cores/maple/usb_serial.cpp


This means I can't automatically merge it, and have to merged the files manually. As I don't think those 3 files should have changed

Serasidis:
The first two files are for fixing the DFU bootloader in Linux and MacOSX.
I should have sent a second PR for those changes.
The most changes in usb_serial.cpp are text formatting and a fix for finding the magic word "1EAF" in Linux and MacOSX

rogerclarkmelbourne:
I've made a new branch so that this can be tested

https://github.com/rogerclarkmelbourne/Arduino_STM32/tree/add-usb-hid-bootloader

I've merged your PR. But I reverted some of the files which were unrelated to the HID bootloader

and also

STM32F1/cores/maple/usb_serial.cpp

Because this has formatting changes, not related to the code change.

As far as I can tell the only code change in usb_serial.cpp is

```
+        int len = usb_cdcacm_data_available();
-        if (usb_cdcacm_data_available() >= 4) 
+        if (len >= 4) 
-		{
+        {
-            uint8 chkBuf[4];
+            uint8 chkBuf[64];
 
 
             // Peek at the waiting bytes, looking for reset sequence,
             // Peek at the waiting bytes, looking for reset sequence,
             // bailing on mismatch.
             // bailing on mismatch.
-            usb_cdcacm_peek_ex(chkBuf, usb_cdcacm_data_available() - 4, 4);
+            usb_cdcacm_peek(chkBuf, len);
+
             for (unsigned i = 0; i < sizeof(magic); i++) {
             for (unsigned i = 0; i < sizeof(magic); i++) {
-                if (chkBuf[i] != magic[i]) 
+                if (chkBuf[len + i - 4] != magic[i]) 
-				{
+                {
-			        reset_state = DTR_LOW;
+                  reset_state = DTR_LOW;
-                    return;
+                  return;
                 }
                 }
```

Why did you increase the buffer size to 64 bytes ?

What is wrong with the existing code ?
Note this code is used by when the upload type is the maple / dfu bootloader, so changes to this need to be tested for that bootloader as well

Also. You have new binaries for the hid upload tool. Did you add these files in your PR, or are they updates ?

Serasidis:
well, the changes in the usb_serial.cpp file have benn done after a lot of debugging. The problem was the following.
When the software (upload-reset) sends the the DTR + RTS + 1EAF sequence, at the usb_serial.cpp side the rxHook function is executed by receiving the following bytes:
1st execution
`some-scrap-bytes and ASCII 1` Data length > 4 . The last byte is the ASCII "1" (or 0x31)

2nd execution
`only the ASCII character "E"` Data length = 1 that is less than 4 that needs the usb_cdcacm_peek_ex() function

3nd execution
`only the ASCII character "A"` Data length = 1 that is less than 4 that needs the usb_cdcacm_peek_ex() function

4nd execution
`only the ASCII character "F"` Data length = 1 that is less than 4 that needs the usb_cdcacm_peek_ex() function

Normally, the usb_cdcacm_peek_ex() should have read the whole data. Not only a part of them. Perhaps, in Linux and Mac the usb_serial data send process is a little bit slower than windows.

The usb_cdcacm_peek() function works properly. It reads the incoming data buffer without removing the data from it (peek). 
What I did was to create a local 64bytes buffer (chkBuf) that every time gets a copy of the incoming data without removing them from the main buffer. After that, it compares the chkBuf[] with the  magic[] to see if they match and does the  MCU reboot.

  

> Also. You have new binaries for the hid upload tool. Did you add these files in your PR, or are they updates ?

In Ubuntu and MacOS, the binary files of the repo have not been working. So, I forced to recompile the files from the source folder. I changed nothing in the source code.

Serasidis:
**Steve**,  the **usb_cdcacm_peek_ex()** didn't work properly neither on Linux nor on MacOSX. The **usb_cdcacm_peek()** works ok on Windows, Linux and MacOSX. The disadvantage is that the **usb_cdcacm_peek()** doesn't have the offset attribute like the **usb_cdcacm_peek_ex()**. That means, you cannot get the latest bytes by using the **peek()**. peek() gets always the first available bytes of the buffer. The "1EAF" is usually at the end of the buffer.

That is why I used a local (temporary buffer to check the incoming data.

stevstrong:
You're right.
I think a new function `usb_cdcacm_peek_char_ex(uint16 offset)` would make sense which would return the received char with the specified offset counted from the first received one. The `offset` goes then from 0 up to the number of available received chars. This way no extra buffer would be needed.

rogerclarkmelbourne:
I think this has been superseded, so I'm going to close it

PR #522, opened 2018-05-21
Changed prescaler variable type to uint32. Fixes #520.

### LINKED ISSUE #520: Make RTC prescaler uint32
The RTC prescaler register allows 20 bits wide values to be written, whereas RTClock library uses `uint16` for prescaler argument:

https://github.com/rogerclarkmelbourne/Arduino_STM32/blob/8fa6bb45960097be75b6df6fc08705c85d4e4a09/STM32F1/libraries/RTClock/src/RTClock.h#L44

My device has 12 MHz crystal. The prescaler should be 93749 (12000000/128) in my case, but it won't fit into `uint16` argument.
The underlying `rtc_set_prescaler_load` function, which is used internally, already uses `uint32` argument.
Please update the library to use `uint32` instead.

me21:
Signed-off-by: Alexandr Zarubkin <me21@yandex.ru>

### DIFF
--- a/STM32F1/libraries/RTClock/src/RTClock.cpp
+++ b/STM32F1/libraries/RTClock/src/RTClock.cpp
@@ -39,7 +39,7 @@
 	
 	}//end RTC
 
-	RTClock::RTClock(rtc_clk_src src, uint16 prescaler ) {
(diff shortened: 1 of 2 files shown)


me21:
Of course.

PR #523, opened 2018-05-21
Added preprocessor symbol USE_HSI_CLOCK for running with HSI clocking

me21:
This change is done for generic variants only. By default, HSE clocking scheme is used, as before this change.

Signed-off-by: Alexandr Zarubkin <me21@yandex.ru>


rogerclarkmelbourne:
Looks OK, but I'd need to test it before I can merge

Issue #531, opened 2018-06-19
delayMicroseconds(0); call freezes the MCU


gioblu:
Ciao @rogerclarkmelbourne, thank you very much for your work. 

As described by the title if I call:
```cpp
delayMicroseconds(0);
```
the device ends up frozen. 
This issue has been discovered by @osman-aktepe and it has been discussed here: https://github.com/gioblu/PJON/issues/212

stevstrong:
I think this can be closed now, fixed with  #532 .

PR #532, opened 2018-06-20
Fix to delayMicroseconds(0) bug

gioblu:
Ciao @rogerclarkmelbourne here is the fix as you requested https://github.com/rogerclarkmelbourne/Arduino_STM32/issues/531

### DIFF
--- a/STM32F1/system/libmaple/include/libmaple/delay.h
+++ b/STM32F1/system/libmaple/include/libmaple/delay.h
@@ -46,6 +46,7 @@ extern "C" {
  * @param us Number of microseconds to delay.
  */
 static inline void delay_us(uint32 us) {
+    if(!us) return;
     us *= STM32_DELAY_US_MULT;
 
     /* fudge for function call overhead  */


rogerclarkmelbourne:
Thanks

I'll need to pull locally and test before I can merge.
I do this even for seemingly simple PR's as I had problems in the past with some PR's not compiling because of typos etc

gioblu:
Sure, no problem with have temporarily patched where needed.
I will wait for this pr to be closed to delete the patch :)
Thank you

rogerclarkmelbourne:
I'd need to check what happens on the AVR Arduino.

We can potentially compensate for the number of clock cycles taken for the if statement, but that code should probably be written in assembler to guarantee that we know how many cycles that would be.

It looks like there is already a hack where the value is decremented to account for the function call overhead

```
    /* fudge for function call overhead  */
us--;
```

It looks like the official Arduino checks for zero

https://github.com/arduino/ArduinoCore-samd/blob/395695b0008765e18fee32e0214c9db53fb6c097/cores/arduino/delay.h#L66-L69

And does not seem to compensate for it.


Looking in more detail, the code also has this

` us *= STM32_DELAY_US_MULT;
`
Which is a bit alarming, because multiplication is slow unless the value is power of 2

gioblu:
Ciao @rogerclarkmelbourne I agree with you about the multiplication issue.

rogerclarkmelbourne:
I pushed my own fix, which is semantically different but functionally the same

NOTE. I know this adds a small additional delay, but this function is not highly accurate.

If someone wants to write a more accurate version please feel free to send a PR

rogerclarkmelbourne:
BTW.
If you notice an issue with the timing we could possibly change the line that does 

`us--;`

to 

`us -=2;`

stevstrong:
I am not sure if that `us -= 2;` would bring improvement, because what happens if us = 1?

rogerclarkmelbourne:
Steve

Thats a good point.

I think we'll need to leave it as it is, as I don't think its ever been that accurate.

stevstrong:
I agree.

PR #534, opened 2018-07-02
F1: added function dma_get_count

stevstrong:
Utility function to check the number of transfer yet to be done by DMA.


rogerclarkmelbourne:
Looks very low risk.

I think I can merge.

When is it used ?

stevstrong:
It is useful to check if the DMA has finished the transfer, and if not it gives how many transfers are yet to be done.

rogerclarkmelbourne:
ok.


thanks

VictorLamoine:
@rogerclarkmelbourne for reference of an example use case: http://www.stm32duino.com/viewtopic.php?p=46799#p46799

Thanks @stevstrong

PR #537, opened 2018-07-06
F1: fix wrong delay timing

stevstrong:
Fixes the improper sub-millisecond delay, as discussed here: https://github.com/rogerclarkmelbourne/Arduino_STM32/commit/85eb51aacb3b6ceaf1575112cd363eb9d3a4649b#commitcomment-29584548
Use working (tested) stock AVR code instead.

### DIFF
--- a/STM32F1/cores/maple/wirish_time.cpp
+++ b/STM32F1/cores/maple/wirish_time.cpp
@@ -34,13 +34,18 @@
 #include <libmaple/delay.h>
 #include "Arduino.h"
 
-void delay(unsigned long ms) {
-    uint32 start = millis();
-	do
-	{
-		yield();
-	}
-    while (millis() - start < ms);
+void delay(unsigned long ms)
+{
+    uint32 start = micros();
+    while (ms > 0)
+    {
+        yield();
+        while ( (ms > 0) && ((micros() - start) >= 1000) )
+        {
+            ms--;
+            start += 1000;
+        }
+    }
 }
 
 void delayMicroseconds(uint32 us) {


rogerclarkmelbourne:
Are you sure this is accurate for long millisecond delays

It looks like your change relies on micros() but I don't think micro's is that accurate

the existing delay relies indirectly on the systick interrupt which is going to be a lot more accurate than calling micros() thousands of times

stevstrong:
micros() is based partially on millis(): https://github.com/rogerclarkmelbourne/Arduino_STM32/blob/master/STM32F1/cores/maple/wirish_time.h#L54-L71.
So I am confident that it works also for long millis delay.
However, of course, now the timing range will be more limited (up to 65535 milliseconds?). But I think if someone wants such a long delay (or more) he/she should use other timing methods.

justinschoeman:
There is no change to the range of the delay (the loop in 1000us increments will prevent micros() overflows from affecting the millis() range).

The function could potentially be made more efficient (in exchange for large size) - but why bother - the function is there to waste time anyway ;) .

I might suggest one tweak though:

`while ( ms > 0 && (micros() - start) >= 1000)`

becomes

`while ( ms > 0 && (uint32_t)(micros() - start) >= 1000)`

simply because I am paranoid about the ways in which the compiler might promote types...

stevstrong:
I don't think that tweak is necessary as both `micros()` and `start` are 32 bit variables.

justinschoeman:
Correct - it is not necessary. It is just paranoia. Integer promotion is frequently broken in compilers. As an example, it is actually broken in the current Arduino ARM compiler - but broken in such a way that the above code works irrespective of int width (it is not supposed to work correctly if shorter int types are used, but it does).

Adding an explicit type cast will ensure that it will work irrespective of int width, and irrespective of how well the promotion rules are implemented by the compiler.

rogerclarkmelbourne:
Sounds like this change is not actually necessary, so I will close it

justinschoeman:
The actual pull is necessary - it returns delay() to the standard Arduino API definition (delay *at least* x ms), and fixes compatibility issues with at least 2 libraries.

My comments were purely on the coding style of the fix.  It is perfectly functional within this build environment.

Thanks,
Justin

rogerclarkmelbourne:
Re-opened

rogerclarkmelbourne:
Tested and merged...

Thanks

justinschoeman:
Thanks!

PR #542, opened 2018-07-21
fix not working availableForWrite for STM32F1

smarq8:
I already tested it and it work fine.

### DIFF
--- a/STM32F1/cores/maple/HardwareSerial.cpp
+++ b/STM32F1/cores/maple/HardwareSerial.cpp
@@ -126,12 +126,7 @@ int HardwareSerial::peek(void)
 
 int HardwareSerial::availableForWrite(void)
 {
-/* Roger Clark. 
- * Currently there isn't an output ring buffer, chars are sent straight to the hardware. 
- * so just return 1, meaning that 1 char can be written
- * This will be slower than a ring buffer implementation, but it should at least work !
- */
-  return 1;
+    return this->usart_device->wb->size-rb_full_count(this->usart_device->wb);
 }
 
 size_t HardwareSerial::write(unsigned char ch) {


smarq8:
I can not post any simple test but I can show my resoult from my flight controller connected to PC via ftdi over Serial3 (in my script it is Serial but I just use "#define Serial Serial3" for some backward compatibility)

Here SerialUSBdebug() print to my serial monitor on live available space in RX buffer (orange) and available space in TX buffer after my modyfication (green)
Over Serial3 i stream 4 floats (16bytes+2byte header) in MPU_Fs frequency (see in the right window) and also my PC send request for 38 paramaters on every 10ms, in almost any case the answer is float number (4bytes+2byte header).
I also increase USART_RX_BUF_SIZE and USART_TX_BUF_SIZE to 256
screenshot 1532285191

and here is resoult after increace MPU_Fs and now my uc is overloaded and available space in buffers drop.
screenshot 1532285199

and here is transition before fully overload
screenshot 1532285368
screenshot 1532285263

stevstrong:
As long as nobody else can reproduce what you are doing, it doesn't help.

smarq8:
as you can see during filling available space is dropping and then its increasing as it don't put any new data to buffer.

By the way I catch another issue that until first loop iteration end then there is no data on my PC from USB serial port.
```
void setup() {
  Serial.begin(115200);
  Serial3.begin(115200);
  for(int i=0;i<20;i++){ // fill TX buffer
    Serial.println("this will be not visible");
  }
}

int l =0;
void loop() {
if (l>2)
  while(1){}
  Serial.print("start: ");
  Serial.println(l);
  for(int i=0;i<250;i++){ // fill TX buffer
    Serial3.write(255);
    Serial.println(Serial3.availableForWrite());
    delayMicroseconds(45); // control filling rate/slope
  }
  for(int i=0;i<250;i++){ // watch tx buffer during sending bytes
    Serial.println(Serial3.availableForWrite());
  }
  
  Serial.print("end: ");
  Serial.println(l);
  delay(2000);

  l++;
}
```
screenshot 1532351932

stevstrong:
> By the way I catch another issue that until first loop iteration end then there is no data on my PC from USB serial port.
This is no issue, it happens because the software does not wait until the serial port is opened on the PC so that the first bytes are lost.
If you insert in your code
`while (!Serial);`
this will wait till the PC opens the USB serial connections and will start sending data only then.

rogerclarkmelbourne:
Close as its not actually an issue. Its a known side effect of the code not waiting for the host system to be read (Thanks Steve....)

stevstrong:
Roger, sorry but my last comment above was on a side issue, the main issue was to get the free space available to write to the ring buffer which issue is corrected by this pr.
So i actually vote for this pr to be merged.

rogerclarkmelbourne:
OK.

I tested and merged

PR #543, opened 2018-07-26
Define defaults build.flags

fpistm:
Useful when using arduino-builder

### DIFF
--- a/STM32F1/boards.txt
+++ b/STM32F1/boards.txt
@@ -49,8 +49,6 @@ mapleMini.menu.cpu_speed.speed_128mhz.build.f_cpu=128000000L
 
 #-- Optimizations
 mapleMini.menu.opt.osstd=Smallest (default)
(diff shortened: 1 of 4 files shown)


stevstrong:
Aren't the optimization flags specified over an Arduino IDE option (see boards.txt)?
If so, why to specify it here and then to overrule it?

fpistm:
Yes, thanks the Arduino IDE but for example when using arduino-builder (or arduinon CLI), this require to fully specified this. while it is the defaults value.
You're right that it's overrule. So the boards.txt could be clean a little, I've updated my commit.

stevstrong:
+1

rogerclarkmelbourne:
I tested this and it seems OK on the Maple mini.

Note. I had to manually merge F1 platform.txt because of a conflict with the HID bootloader.
(I'm not sure why git thought there was a conflict, possibly just the CRLF line ending on the bootloader "pattern"

PR #545, opened 2018-07-30
Multiple devices per I2C bus

pamribeirox:
I've found that when more than one device is using the I2C bus (i've found this on I2C1), when the second library calls the Wire.begin() it triggers a low level ASSERT() hanging the MapleMini (Baite Clone) with the blue led pulsing.
I don't know any reason why the bus (Wire/HWire) should be "initialized" more than once.
So, to avoid messing with some low level functions of the maple library, I've added an attribute that limits the call the i2c_master_enable() to once per "Wire" instance.
Please review the changes and see if this is good for the base ArduinoSTM32
Keep the good work on this resource!
To reproduce the "problem", just call Wire.begin() twice in the setup/main function.

### DIFF
--- a/STM32F1/cores/maple/libmaple/i2c.c
+++ b/STM32F1/cores/maple/libmaple/i2c.c
@@ -175,8 +175,8 @@ void i2c_init(i2c_dev *dev) {
  *                         SDA/PB9.
  */
 void i2c_master_enable(i2c_dev *dev, uint32 flags) {
-    /* PE must be disabled to configure the device */
-    ASSERT(!(dev->regs->CR1 & I2C_CR1_PE));
+    /* If the device is already enabled, don't do it again */
+    if(dev->regs->CR1 & I2C_CR1_PE) return;
 
     /* Ugh */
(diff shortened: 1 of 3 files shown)


rogerclarkmelbourne:
I think this fix potentially has a problem if you are using both I2C busses on the STM32, because it would not call the enable on the 2nd bus.

Probably a better fix is to change i2c_master_enable 

https://github.com/rogerclarkmelbourne/Arduino_STM32/blob/20168b13591d1228d8adcde2d4d1dadf6fba72cf/STM32F4/cores/maple/libmaple/i2c.c#L401

And rather than ASSERT, just return if the bus is already enabled,

pamribeirox:
Thanks for the advice.
I wasn't too confident about doing the changes at the libmaple
Please confirm this is the right solution.

rogerclarkmelbourne:
Thanks for the update.

I'll need to test before I can merge it

rogerclarkmelbourne:
Thanks

Tested with a simple I2C device and seems to have no problems.

PR #548, opened 2018-08-02
Add pgm_read_ptr and pgm_read_ptr_far for better compatibility with libs written for AVR

tfry-git:
Ran into missing pgm_read_ptr while trying to compile NeoGPS for a blue pill. This is easily fixed by adding it to the existing pgmspace.h compatibility header. Also added pgm_read_ptr_far for consistency / completeness.

This change seems to be all that is needed to make NeoGPS work on STM32F1.


rogerclarkmelbourne:
Thanks

Looks OK, but I'll need to double check and do a test build with it to confirm everything is OK

rogerclarkmelbourne:
Thanks

Tested, and does not seem to cause any problems...

PR #554, opened 2018-08-20
Fix clearing of TIMx_SR register

jorisvr:
To clear specific bits in TIMx_SR registers, libmaple uses

```
  TIMx_SR &= ~(events_handled);  // wrong
```

This is wrong. The `&=` operator expands to a read-modify-write sequence consisting of multiple instructions. If a new event occurs between the read from TIMx_SR and the write to TIMx_SR, that new event will be cleared immediately without ever being noticed or handled.

A better way to clear specific bits is

```
  TIMx_SR = ~(events_handled);  // good
```

This writes '0' to the bits to be cleared, and '1' to all other bits. The hardware will not allow software to change any SR bits from 0 to 1; only hardware events can do that. So bits written as '1' are effectively ignored and bits written as '0' are cleared. Exactly what we need.

I tested this in my program which captures input on TIM4_CH1 while simultaneously tracking TIM4 overflow events. Without this fix, timer overflow events are sometimes lost if a capture event occurs just before the timer overflows. This fix solves it.

### DIFF
--- a/STM32F1/system/libmaple/timer_private.h
+++ b/STM32F1/system/libmaple/timer_private.h
@@ -129,7 +129,7 @@ static inline __always_inline void dispatch_single_irq(timer_dev *dev,
         void (*handler)(void) = dev->handlers[iid];
         if (handler) {
             handler();
-            regs->SR &= ~irq_mask;
+            regs->SR = ~irq_mask;
         }
     }
 }
@@ -165,7 +165,7 @@ static inline __always_inline void dispatch_adv_trg_com(timer_dev *dev) {
     handle_irq(dsr, TIMER_SR_TIF,   hs, TIMER_TRG_INTERRUPT, handled);
     handle_irq(dsr, TIMER_SR_COMIF, hs, TIMER_COM_INTERRUPT, handled);
 
-    regs->SR &= ~handled;
+    regs->SR = ~handled;
 }
 
 static inline __always_inline void dispatch_adv_cc(timer_dev *dev) {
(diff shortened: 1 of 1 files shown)


rogerclarkmelbourne:
@RickKimball 
Can you advise on this one as well 

Thanks

stevstrong:
To extend some extra info, the SR bits are of type "rc_w0", which means that they are cleared by writing 0.
That is why i think this PR is ok.

rogerclarkmelbourne:
@stevstrong 
Thanks

justinschoeman:
I am battling to find the correct reference from the ST manuals again, but this proposal is definitely the way the manual says it should be done...

RickKimball:
You are probably correct @justinschoeman  In STM32CUBEMx they use a write as opposed to a MODIFY_READ_WRITE

__STATIC_INLINE void LL_TIM_ClearFlag_UPDATE(TIM_TypeDef *TIMx)
{
  WRITE_REG(TIMx->SR, ~(TIM_SR_UIF));
}

RickKimball:
@jorisvr do you have a test case for this change?

jorisvr:
@RickKimball I tested and confirmed the fix in a custom setup. I don't see an easy way to create an automated regression test. Is there an existing timer testsuite that I could extend?

jorisvr:
Sure. I'll attach my Arduino sketch. It is not a minimal test to demonstrate the bug or anything like that.

The central issue is that I want to capture input events (PPS pulses) with TIM4 while also detecting each overflow of the timer counter. I clearly see that I sometimes "lose" a timer overflow interrupt if an input pulse occurred very close to the moment of timer overflow (i.e. within a few microseconds).

This is due to the fact that the libmaple timer code can sometimes clear a pending event from TIM4_SR before even noticing that the hardware raised the pending event.

[example_sketch.txt](https://github.com/rogerclarkmelbourne/Arduino_STM32/files/2753354/example_sketch.txt)

rogerclarkmelbourne:
It looks like this has now been tested by several people.
See http://www.stm32duino.com/viewtopic.php?t=4685

So I'll merge it

PR #556, opened 2018-08-23
Add -fno-use-cxa-atexit compiler option to reduce unused code.

justinschoeman:
C++ requires that static object destructors are called when the
program terminates. Recent versions of GCC use __cxa_atexit to
register exit handlers which call these object destructors.

With the -fno-use-cxa-atexit command line option, GCC uses atexit
instead (same functionality, but only up to 32 exit handlers can
be registered).

Since the Arduino main process can not return, and calls to exit()
simply call an infinite loop - this is completely irrelevant. No exit 
handlers are ever called.

By removing the unused __cxa_atexit functionality we save a bit of
space, and enable further cleanups (in later commits).


stevstrong:
Tested and seems to work.
Thanks!

PR #557, opened 2018-08-24
Remove uncallable libc call

justinschoeman:
main() never exits. The subsequent call to exit() (which can never be reached) links in the libc exit symbol and all that it calls (including free(), exitproc and a few others).

Removing this saves around 2k of flash in most cases.


stevstrong:
Tested and it works.
Thanks!

PR #558, opened 2018-08-24
Remove exit() and atexit() handlers from libc.

justinschoeman:
This is the last of the basic memory use cleanups - all revolving around removing static object cleanup handlers and the associated dynamic memory management code, which is included in all built projects, but never used.

NOTE: Original Arduino cores also replace atexit with dummy handler.

exit()/atexit() handling is *only* for when main exits, which is not
possible in the Arduino environment.

We use the gcc linker to mangle these symbols and replace them with
our own library.  Symbols are stong in the default library, so simple
overloading does not work.

Total saving around 5k of flash and 2k of ram - all for functionality
which is linked in but never used.

* all patches are independant and functional, but most saving is with
all patches included.

### DIFF
--- a/STM32F1/cores/maple/libc_repl.c
+++ b/STM32F1/cores/maple/libc_repl.c
@@ -1,0 +1,7 @@
+#include <stdlib.h>
+
+int __wrap_atexit(void (*function)(void) __attribute__((unused))) { return 0; }
+
+int __wrap___cxa_atexit(void (*func) (void *) __attribute__((unused)), void * arg __attribute__((unused)), void * dso_handle __attribute__((unused))) { return 0; }
+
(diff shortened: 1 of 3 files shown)


stevstrong:
Although I am in favor of this, it looks like it need extensive testing for all kind of scenarios.

justinschoeman:
I am closing this pull request.
I must admit, the build system in this tool chain is unfamiliar to me.  I can not find where libc is linked - but it seems to be after the application objects, so application symbols take precedence over libc symbols.
Having two strong symbols (even in a defined order) should result in at least warnings (esp since --warn-common is set) - but it does not.  I have no idea why not though...
The end result is that this can be implemented as a library, and does not need to go into the core.

stevstrong:
I tried it with an empty sketch in Arduino IDE and this patch saves ~500 bytes code and 50 bytes RAM.

PR #567, opened 2018-09-18
Fix for possible break of DFU functionality

shodan8192:
Current implementation of reset to bootloader routine in `rxHook` from file `STM32F1/cores/maple/usb_serial.cpp` can cause receiving signal SIGSEGV during try to upload new sketch. The mechanism of that is described by me with details here : http://www.stm32duino.com/viewtopic.php?f=53&t=4125 and its occurence is dependant of USB data structures location in RAM.

Proof-of-concept sketch, triggering described behavior was tested on Arduino IDE 1.8.7 on Linux x64 with latest commit of Arduino_STM32 and it can be downloaded here : https://gist.github.com/shodan8192/b1b68d747f4395da3edb37e6300e39e7

First upload of this sketch will be succesfull. Next try will cause fast pulsation of LED and message : `dfu-util: No DFU capable USB device available`. After manual reset, board is in "perpetual bootloader" and sketch can be uploaded.

This fix remove `wait_reset` procedure and replace asm exception exit code with simple `nvic_sys_reset()`

### DIFF
--- a/STM32F1/cores/maple/usb_serial.cpp
+++ b/STM32F1/cores/maple/usb_serial.cpp
@@ -273,18 +273,6 @@ static void ifaceSetupHook(unsigned hook __attribute__((unused)), void *requestv
 #endif	
 }
 
-#define RESET_DELAY 100000
-#ifdef SERIAL_USB 
-static void wait_reset(void) {
-  delay_us(RESET_DELAY);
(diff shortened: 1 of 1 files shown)


stevstrong:
Tested several consecutive uploads and it seems to work.
Thanks.

PR #569, opened 2018-09-24
F4: fix HID upload with Arduino IDE

stevstrong:
See http://stm32duino.com/viewtopic.php?f=32&t=3661&start=150#p49567.
Without this the HID upload won't work with Arduino IDE.


rogerclarkmelbourne:
Thanks

PR #580, opened 2018-12-02
Explicit STM32duino.com platform origin

Squonk42:
According to http://www.stm32duino.com/viewtopic.php?f=16&t=4377&p=51233#p51233


rogerclarkmelbourne:
Perhaps we need to include the words "Unofficial" or even "UNOFFICIAL" in the title ;-)

PR #581, opened 2018-12-09
add RTC adjustment example in RTClock lib

ag88:
I've added my RTC drift adjustments as a rather complete working sketch / app in STM32F1/libraries/RTClock/examples/RTCAdj

this is tested on maple mini. i'd think it'd possibly work on stm32f4 as well but i've not tested that hence i've currently added it to the STM32F1 branch

i noted that my additions seem to have added some changes in 
 tools/linux/maple_upload and tools/macosx/maple_upload
those additions i couldn't remember when i made them but it is mainly to echo the dfu-util command so that it is visible in the console when the sketch is downloaded. 
unfortunately i'm not sure how to remove this part of changes as this PR is mainly on the RTC adjustments examples

the RTC adjustments is a day or so work and i've done some tests on them to see that they work as intended on maple mini, pretty much a  complete sketch/app on its own. But it is good to have it as 'examples' as not everyone needs this, the adjustments functions are also organised into its own header file and cpp file so that it is pretty much a mini library module with the sketch itself as the example and app


hyperion11:
image
hi! try upload latest commit.
Only #include "src\rtadjust\rtadjust.h"
Compiler error multiple definition of `delim'

ag88:
oops i missed out on an email notification on this, i've updated the codes as suggested. compiled ok no issues. it seemed the variable delim may be quite a common variable name. and commited the changes in my repository (fork)

the reason the messages appears twice is due to the way i'm handling key inputs.
i made the codes respond to a single keystrokes, hence, if you hit enter send blank lines etc, it would show the 'help' message again.

ag88:
re-organising my git, re-creating my local copy in sync with master git, closing this and would send again after i've a clean local copy

PR #592, opened 2019-01-25
Add method Print.printf().  Fixed & tested.

u48:
Added the printf() method to the Print class.

With this implementation, Print.print() works with any devices - Uart, TFT, SDCard and others.

The previous version has not been tested. I'm sorry.
This worked on other my platforms, but a bug was discovered on the Arduino STM32. 
Fixed.

### DIFF
--- a/STM32F1/cores/maple/Print.cpp
+++ b/STM32F1/cores/maple/Print.cpp
@@ -221,32 +221,32 @@ size_t Print::println(const Printable& x)
 
 /*  
 //Bug detected 
-//failed: fopencookie( 0, "w+",  { NULL,  WR_fn, NULL, NULL }) - failed on gcc stm32.
-//failed: fopencookie( 0, "rw+", { NULL, WR_fn, NULL, NULL }) - failed on gcc stm32. 
+//failed: fopencookie( 0, "w+",  { NULL,  WR_fn, NULL, NULL }) 
+//failed: fopencookie( 0, "rw+", { NULL, WR_fn, NULL, NULL }) 
 //
-//worked: fopencookie( 0, "rw+", { RD_fn, WR_fn, NULL, NULL }) - failed on gcc stm32.
-//worked: fopencookie( 0, "rw+", { NULL,  WR_fn, NULL, NULL }) - failed on gcc stm32.
+//worked: fopencookie( 0, "rw+", { RD_fn, WR_fn, NULL, NULL }) 
+//worked: fopencookie( 0, "rw+", { NULL,  WR_fn, NULL, NULL })
 //
 //Resume:  To fix - you should always specify the cookies_read function, zero is not valid. 
 */
 
 
-  // Read/Write blocks to device.
-  // Non class function for use from libc.printf,
-  // pointer to class Print passed in argumwent dev.
-  static ssize_t cookie_read_helper(void *dev __attribute__((unused)) , const char* buff __attribute__((unused)) , size_t len __attribute__((unused)) )
-  {
+// Read/Write blocks to device.
+// Non class function for use from libc.printf,
+// pointer to class Print passed in argumwent dev.
+static ssize_t cookie_read_helper(void *dev __attribute__((unused)) , const char* buff __attribute__((unused)) ,
+				    size_t len __attribute__((unused)) )
+{
     return 0; 
-  }     
+}     
 
-  static ssize_t cookie_write_helper(void *dev, const char* buff, size_t size)
-  {     
+static ssize_t cookie_write_helper(void *dev, const char* buff, size_t size)
+{     
     Print* pPrint=(Print*)dev;
     ssize_t len=0;
     for(char* c = (char*)buff; size-- >0; len++) pPrint->print(*c++);
     return len;
-
-  }
+}
    
    
 size_t Print::printf(const char *format, ...)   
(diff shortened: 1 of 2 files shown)


RickKimball:
Is there any way to add an additional format specifier for binary? I'd love to be able to use:

Serial.printf("bits=0b%08b\n", 0x33);

justinschoeman:
Not tested, but this should work if you are not using nano.specs:
https://www.gnu.org/software/libc/manual/html_node/Customizing-Printf.html

stevstrong:
you can choose one solution from here: https://stackoverflow.com/questions/111928/is-there-a-printf-converter-to-print-in-binary-format

RickKimball:
Thanks @justinschoeman! We learn something new everyday.

rogerclarkmelbourne:
Why does one of the files include this comment ?

`/*  
int Print::printf (__const char *__restrict __format, ...)
//Bug detected 
 {
//failed: fopencookie( 0, "w+",  { NULL,  WR_fn, NULL, NULL }) 
FILE *__restrict __stream;
//failed: fopencookie( 0, "rw+", { NULL, WR_fn, NULL, NULL }) 
     int ret_status = 0;
//

//worked: fopencookie( 0, "rw+", { RD_fn, WR_fn, NULL, NULL }) 

//worked: fopencookie( 0, "rw+", { NULL,  WR_fn, NULL, NULL })
     va_list args;
//
     va_start(args,__format);
//Resume:  To fix - you should always specify the cookies_read function, zero is not valid. 
     ret_status = vfprintf(__stream, __format, args);
*/`

u48:
> Is there any way to add an additional format specifier for binary? I'd love to be able to use:
> 
> Serial.printf("bits=0b%08b\n", 0x33);

As implemented in LIBC. :)
Need to check this.

u48:
> As far as I can tell this code references the function
> fopencookie
> which does not exist

Hmm..
This is a LIBS extension defined in the header file studio.h. And provided by the compiler you are using.

This is supported in the native Linux gcc, Raspberry,  gcc for Arduino DUE and ESP8286/32.
But why is it disabled in the mingw.
What compiler You use?

u48:
>>Why does one of the the change to print.cpp include this block of comment ?
It's just to avoid questions why need an empty cookie_read_helper () function.
And there was no attempt to remove it.
Read and delete it.

rogerclarkmelbourne:
I didn't test whether it compiled yet. 
I just wasnt aware that fopencookie was a function in the standard c libraries.

I'll locally merge in a separate branch and see whether I get any errors when I compile.

rogerclarkmelbourne:
OK.
It looks like it fopencookie can be linked in.

But more testing would be needed before I can merge

u48:
fopencookie () is an extension of LIBC.

The correct use of this extension is as follows:
   #define _GNU_SOURCE
   #include <stdio.h>

And not as it is now written in the source, only
#include <stdio.h>

It is possible that such a trifle is a source of problems.

rogerclarkmelbourne:
This PR does not seem to work for me.

Please post a small example sketch which works
Also let me know if you have USB serial connected and what upload method you are using (as this affects the config of the USB Serial and Serial devices e.g. "Serial" can be either USB Serial or UART Serial depending on upload method

u48:
I suspect that this is not compiled correctly or works.
I take 1-4 days for tests.

I will do a few test sketches, and describe their behavior on various new platforms assembled from scratch.
For experiments there is all the necessary and unnecessary hardware

rogerclarkmelbourne:
OK.

Thanks

BTW.
@RickKimball 

Did this PR work for you ?

STM32F1/cores/maple/Print.cpp
```diff
@@ -208,24 +208,72 @@ size_t Print::println(const Printable& x)
...
- }
- #endif
+
+/*  
+//Bug detected 
+//failed: fopencookie( 0, "w+",  { NULL,  WR_fn, NULL, NULL }) 
+//failed: fopencookie( 0, "rw+", { NULL, WR_fn, NULL, NULL }) 
+//
+//worked: fopencookie( 0, "rw+", { RD_fn, WR_fn, NULL, NULL }) 
+//worked: fopencookie( 0, "rw+", { NULL,  WR_fn, NULL, NULL })
```
stevstrong: Line 229 (worked) seems to be identical to line 226, which failed.
Btw, do we really need this comment block? I would remove it.
The solution matters, which is implemented below.

u48: Block comments should be removed. :)

The GLIBC documentation says that when fopencookie () is called, pointers to both the write () and read () functions must be passed.

That is, if reading is not required, a valid pointer to a dummy function should be passed, and not NULL. Otherwise it does not work.

stevstrong: I understand the reason, I just asked to remove unneeded comment block (which includes contradictory information).

victorpv:
Steve looks like the comments were cleaned up on February 28th, after your last comment.

fpistm:
Maybe you could evaluate this:
https://github.com/stm32duino/Arduino_Core_STM32/blob/master/cores/arduino/Print.cpp#L223

stevstrong:
I think the solution from @fpistm should be applied, especially lines
https://github.com/stm32duino/Arduino_Core_STM32/blob/42b2ab21373f6769cfc6a9b6ca03bf303fe9b4fa/cores/arduino/Print.cpp#L274-L290

PR #597, opened 2019-02-08
Protect _settings[] array from accesses beyond end

mitchmitchell:
_settings[] array is accessed beyond the end of the array if BOARD_NR_SPI is defined as 1 on for example the HyTiny board.

Running NodeManager/MySensors software crashes with a SIGSEGV due to SPI issue where _settings[] array is written to beyond the end of the array and places the address of dma1() in the getTimePtr in the Arduino Time.cpp library.  This occurs when _settings[1].spiRxDmaChannel is assigned DMA_CH4 in SPI.cpp on the HyTiny where the _settings[] array is only one element long.

### DIFF
--- a/STM32F1/libraries/SPI/src/SPI.cpp
+++ b/STM32F1/libraries/SPI/src/SPI.cpp
@@ -554,12 +554,16 @@ void SPIClass::onReceive(void(*callback)(void)) {
     _currentSetting->receiveCallback = callback;
     if (callback){
         switch (_currentSetting->spi_d->clk_id) {
+            #if BOARD_NR_SPI >= 1
         case RCC_SPI1:
             dma_attach_interrupt(_currentSetting->spiDmaDev, _currentSetting->spiRxDmaChannel, &SPIClass::_spi1EventCallback);
             break;
+            #endif
+            #if BOARD_NR_SPI >= 2
         case RCC_SPI2:
             dma_attach_interrupt(_currentSetting->spiDmaDev, _currentSetting->spiRxDmaChannel, &SPIClass::_spi2EventCallback);
             break;
+            #endif
             #if BOARD_NR_SPI >= 3
         case RCC_SPI3:
             dma_attach_interrupt(_currentSetting->spiDmaDev, _currentSetting->spiRxDmaChannel, &SPIClass::_spi3EventCallback);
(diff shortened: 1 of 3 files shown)


rogerclarkmelbourne:
Looks reasonable

BTW.

Whats "NodeManager"

mitchmitchell:
I use https://www.mysensors.org/ platform, and NodeManager is a package that sits on top of it that makes building an image for different types of sensors really really easy.  MySensors is built around the 2.4 Ghz radios like the NRF24L01 and several others that allow nodes to communicate with gateways that publish the data to things like MQTT. https://www.sparkfun.com/datasheets/Components/nRF24L01_prelim_prod_spec_1_2.pdf

BTW, I did a search for BOARD_NR_SPI and it looks like there are a few other places where it is checked for >= 3 but not 1 or 2 and probably should be.  I can fix those and submit another pull request if you like.

The Black Magic Probe is awesome for locating this sort of thing LOL!

rogerclarkmelbourne:
If there are other places in the code, we may as well fix them in the same PR if you don't mind updating it.

Thanks.

PS. I always do a test compile and often a hardware test before I merge PRs, as it saves problems later, as I've had PR's from reputable contributors which failed to compile for me ;-)

mitchmitchell:
Will do, give me a few minutes.

rogerclarkmelbourne:
Ta,

rogerclarkmelbourne:
I probably won't get chance to test this tonight (its 17:20 local). 
I'll try to merge it tomorrow

mitchmitchell:
Ok I got it built with the additional changes and it looks like github picked them up in the pull request.  I'm running the software on my HyTiny test board now, takes about 10 minutes to crash if this didn't work.

mitchmitchell:
Ok so far so good.  I will let it run the rest of the night and make sure its still alive and kicking in the morning (later today LOL)

stevstrong:
Is the check for `#if BOARD_NR_SPI >= 1` really necessary?
I mean, each board with F103 will have at least one SPI, so why to test against that?

rogerclarkmelbourne:
@stevstrong 

I noticed that as well.

But I don't think it causes any problems, and may allow the core to work with some F1xx which don't have SPI (but perhaps all F1 series have SPI )

stevstrong:
I am not aware of any F1xy chip without at least one SPI.
I just want to avoid unneeded ifdefs.

mitchmitchell:
The original developer of the SPI.cpp file thought it necessary in several cases, but just didn't provide the protection in a few cases where code could run off the ends of an array.  I'm not in a position to second guess them so I made the checks consistent throughout.

rogerclarkmelbourne:
No worries.

I'd rather leave the >= 1 #if's in place as the are preprocessor directives and have no impact on code size

mitchmitchell:
Thanks Roger!

The little HyTiny board has been running all day without a hiccup so I think this eliminates the crash I was seeing.

mitchmitchell:
Please let me know if you need me to do anything else on this.

rogerclarkmelbourne:
OK

I need to test this on some real hardware

rogerclarkmelbourne:
Thanks

I tested SPI on a BluePill, and it still compiled and ran OK.

I did a test compile on all other variants and the compile didn't have any errors.

mitchmitchell:
Thank you!  Its been running on my HyTiny all this time as well.

rogerclarkmelbourne:
Thanks.

Sorry its taken a while, but I like to at least partially test any changes using a real project.
I'm not using a HYTiny in this case, but I did test it with my current project, which is a system for Ham radio operators which parses text data, thats supposed to go to a Nextion display, and send it to a ePaper display.

(Its nothing fancy, I'm just extracting the data that I need from the Nextion commands and presenting it in a useful format on a 2.9 inchWaveshare ePaper display).

PR #607, opened 2019-03-17
Fixed float constant

profdc9:
Fixed a float constant so double precision multiply is not linked in


stevstrong:
Compiles error free. Merged.
Thanks.

Issue #608, opened 2019-03-22
Not compiling examples on STM32ADC


fa1ke5:
SingleChannelSingleConversion:8: error: 'ADC' does not name a type
 ADC myAdc(ADC1);
 ^
/home/madmax/Arduino/hardware/Arduino_STM32/STM32F1/libraries/STM32ADC/examples/SingleChannelSingleConversion/SingleChannelSingleConversion.ino: In function 'void int_func()':
SingleChannelSingleConversion:13: error: 'adc_result' was not declared in this scope
   Serial.println(adc_result);
                  ^
/home/madmax/Arduino/hardware/Arduino_STM32/STM32F1/libraries/STM32ADC/examples/SingleChannelSingleConversion/SingleChannelSingleConversion.ino: In function 'void setup()':
SingleChannelSingleConversion:18: error: 'myADC' was not declared in this scope
   myADC.setTrigger(ADC_EXT_EV_SWSTART);//start on SWStart bit
   ^
exit status 1
'ADC' does not name a type

**********************************************

what am I doing wrong?

stevstrong:
Check out #610 and let us know if it works.

fa1ke5:
now i have

/tmp/arduino_modified_sketch_718014/SingleChannelSingleConversion.ino: In function 'void int_func()':
SingleChannelSingleConversion:13: error: 'adc_result' was not declared in this scope
   Serial.println(adc_result);
                  ^
/tmp/arduino_modified_sketch_718014/SingleChannelSingleConversion.ino: In function 'void setup()':
SingleChannelSingleConversion:18: error: 'myADC' was not declared in this scope
   myADC.setTrigger(ADC_EXT_EV_SWSTART);//start on SWStart bit
   ^
exit status 1
'adc_result' was not declared in this scope

stevstrong:
OK, I fixed those errors, try it again.
You may eventually need to adapt the analog input pin.

fa1ke5:
D:\users\madmax\Documents\Arduino\sketch_mar23a\sketch_mar23a.ino: In function 'void setup()':

sketch_mar23a:22:9: error: 'class STM32ADC' has no member named 'attachADCInterrupt'

   myADC.attachADCInterrupt(int_func);

         ^

exit status 1
'class STM32ADC' has no member named 'attachADCInterrupt'

stevstrong:
It compiles now for me, check it yourself.

fa1ke5:
[...]

PR #610, opened 2019-03-22
Update SingleChannelSingleConversion.ino

stevstrong:
Corrected class definition.

Intends to solve #608 .


stevstrong:
Compiled error free and tested.

Issue #606, opened 2019-03-12
EEPROM multiple data at the same address


Dave4675:
I have a bluepill of 128KB with a bootloader loaded, I really apreciate this repo, so thank you for that.

I'm tring to store some settings data into the EEPROM but when I load the example I found an issue (or maybe not) if I press 3 for what I know I expect to write at the location at 0x10, until I press 4 of course,, but when I press 5 it takes the next location, keeping the previous stored data that also become inaccessible.

If this issue is well known I wish to understand if I have to manage with the page address directly, and how to do that, otherwise I need to save all data into the ram, format the EEprom and write my new data again, wich I hope not, 'couse the limited wear count.

Dave4675:
Okay, so I made some test, I tried to modify the EE_VerifyPageFullWriteVariable function, the one that calls the real function that writes the half word into the EEprom, or flash to be more precise. 
i made some changes that give the result that I want, but is still not possible to write to the same virtual address till two conditions, or the free side is equal to 0xFFFF wich means that the flash or the page has been erased or the data that we want to write is less than the one already written.
Also the update function is useless because this function already verify this.

this is my modified function:

uint16 EEPROMClass::EE_VerifyPageFullWriteVariable(uint16 Address, uint16 Data)
{
	FLASH_Status FlashStatus;
	uint32 idx, pageBase, pageEnd, newPage;
	uint16 count;
	boolean found = false;
	// Get valid Page for write operation
	
	pageBase = EE_FindValidPage();
	if (pageBase == 0)
		return  EEPROM_NO_VALID_PAGE;
	
	// Get the valid Page end Address
	pageEnd = pageBase + PageSize;			// Set end of page
	
	for (idx = pageEnd - 2; idx >= pageBase; idx -= 4)
	{
		if ((*(__IO uint16*)idx) == Address)		// Find last value for address
		{
			found = false;
			
			count = (*(__IO uint16*)(idx - 2));	// Read last data
			
			if (count == Data)
				return EEPROM_OK;
			
			if (count == 0xFFFF || Data < count)
			{
			
			FlashStatus = FLASH_ProgramHalfWord(idx - 2, Data);	// Set variable data
			if (FlashStatus == FLASH_COMPLETE)
				return EEPROM_OK;
			}
			
			break;
		} else {
			found = true;
			
		}
	}
	if(found){
				// Check each active page address starting from begining
				for (idx = pageBase + 4; idx < pageEnd; idx += 4){
[...]

PR #615, opened 2019-03-27
write and update have been modified

Dave4675:
details described in issue named EEPROM multiple data at the same address

### DIFF
--- a/STM32F1/libraries/EEPROM/EEPROM.cpp
+++ b/STM32F1/libraries/EEPROM/EEPROM.cpp
@@ -212,56 +212,71 @@ uint16 EEPROMClass::EE_VerifyPageFullWriteVariable(uint16 Address, uint16 Data)
 	FLASH_Status FlashStatus;
 	uint32 idx, pageBase, pageEnd, newPage;
 	uint16 count;
-
+	boolean found = false;
 	// Get valid Page for write operation
+	
 	pageBase = EE_FindValidPage();
 	if (pageBase == 0)
 		return  EEPROM_NO_VALID_PAGE;
-
+	
 	// Get the valid Page end Address
 	pageEnd = pageBase + PageSize;			// Set end of page
-
-	for (idx = pageEnd - 2; idx > pageBase; idx -= 4)
+	
+	//FLASH_Unlock();
+	
+	for (idx = pageEnd - 2; idx >= pageBase; idx -= 4)
 	{
 		if ((*(__IO uint16*)idx) == Address)		// Find last value for address
 		{
+			found = false;
+			
 			count = (*(__IO uint16*)(idx - 2));	// Read last data
+			
 			if (count == Data)
 				return EEPROM_OK;
-			if (count == 0xFFFF)
+			
+			if (count == 0xFFFF || Data < count)
 			{
-				FlashStatus = FLASH_ProgramHalfWord(idx - 2, Data);	// Set variable data
-				if (FlashStatus == FLASH_COMPLETE)
-					return EEPROM_OK;
+			
+			FlashStatus = FLASH_ProgramHalfWord(idx - 2, Data);	// Set variable data
+			if (FlashStatus == FLASH_COMPLETE)
+				return EEPROM_OK;
 			}
+			
 			break;
+		} else {
+			found = true;
+			
 		}
 	}
-
-	// Check each active page address starting from begining
-	for (idx = pageBase + 4; idx < pageEnd; idx += 4)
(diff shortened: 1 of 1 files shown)


stevstrong:
Can you please tell me a reason why is this change needed? What kind of issue is solved with this?
Any test sketch which can demonstrate the issue and the solution you provided?

Dave4675:
When I used the eeprom library i found that it didnt overwrite the data to a specific address but created a new one, so the eeprom will give fullfilled error after a bunch of time. my little modification let the data shift from one page to another, so you can overwrite the address that you written before.

STM32F1/libraries/EEPROM/EEPROM.cpp
```diff
@@ -212,56 +212,71 @@ uint16 EEPROMClass::EE_VerifyPageFullWriteVariable(uint16 Address, uint16 Data)
...
+		for (idx = pageBase + 4; idx < pageEnd; idx += 4){
+			if ((*(__IO uint32*)idx) == 0xFFFFFFFF)				// Verify if element 
+			{													//  contents are 0xFFFFFFFF
+				FlashStatus = FLASH_ProgramHalfWord(idx, Data);	// Set variable data
+				if (FlashStatus != FLASH_COMPLETE)
+					return FlashStatus;
+					FlashStatus = FLASH_ProgramHalfWord(idx + 2, Address);	// Set variable virtual address
+					if (FlashStatus != FLASH_COMPLETE)
+						return FlashStatus;
+					return EEPROM_OK;
```
stevstrong2: I think the indentation of these lines is not correct, please check and correct.

STM32F1/libraries/EEPROM/EEPROM.cpp
```diff
@@ -521,28 +537,6 @@ uint16 EEPROMClass::write(uint16 Address, uint16 Data)
...
-  *			- Flash error code: on write Flash error
-  */
-uint16 EEPROMClass::update(uint16 Address, uint16 Data)
-{
-	if (read(Address) == Data)
-		return EEPROM_SAME_VALUE;
-	else
-	    return write(Address, Data);
-}
-
```
stevstrong2: Why is this function removed?

STM32F1/libraries/EEPROM/EEPROM.cpp
```diff
@@ -212,56 +212,71 @@ uint16 EEPROMClass::EE_VerifyPageFullWriteVariable(uint16 Address, uint16 Data)
...
+	
+	
 	// Empty slot not found, need page transfer
 	// Calculate unique variables in page
 	count = EE_GetVariablesCount(pageBase, Address) + 1;
 	if (count >= (PageSize / 4 - 1))
 		return EEPROM_OUT_SIZE;
-
+	}
```
victorpv: Why is this added here? the "if" statement doesn't have the opening bracket

STM32F1/libraries/EEPROM/EEPROM.cpp
```diff
@@ -212,56 +212,71 @@ uint16 EEPROMClass::EE_VerifyPageFullWriteVariable(uint16 Address, uint16 Data)
...
 		{
+			found = false;
+			
 			count = (*(__IO uint16*)(idx - 2));	// Read last data
+			
 			if (count == Data)
 				return EEPROM_OK;
-			if (count == 0xFFFF)
+			
+			if (count == 0xFFFF || Data < count)
```
victorpv: This is incorrect, you can't reprogram 1s in a halfword of flash, only 0s, so comparing Data < count  to decide whether the word can be rewritten is invalid and will lead to corruption.
Either leave is as before (comparing to 0xFFFF) or compare to 0 to make sure the new data has all bits 0.
As per the flash programming manual:

> Standard programming
In this mode the CPU programs the main Flash memory by performing standard half-word
write operations. The PG bit in the FLASH_CR register must be set. FPEC preliminarily
reads the value at the addressed main Flash memory location and checks that it has been
erased. If not, the program operation is skipped and a warning is issued by the PGERR bit in
FLASH_SR register (the only exception to this is when 0x0000 is programmed. In this case,
the location is correctly programmed to 0x0000 and the PGERR bit is not set).

`if (count == 0xFFFF || Data = 0)`

victorpv:
I think this PR is related to this issue:
https://github.com/rogerclarkmelbourne/Arduino_STM32/issues/606

I added  some comments in line but upon a quick review of the code, it would fail to work properly since it tries to rewrite data on halfwords already written. That is invalid as per the programming manual. I added some comments in line.
I haven't checked the whole rest of the changes, since that part already wouldn't work as intended.

This needs to gets rewritten as per the programming manual requirements.

Issue #604, opened 2019-03-04
stlink_upload uses wrong arguement


dwillmore:
On linux, the stlink_upload script passes $4 to st-flash which is null.  The binary path is provided in $1.  Changing this makes IDE uploads with st-link work.

Arduino IDE 1.8.5, recent Arduino_STM32 copy.

Problem applies to the linxu64 version as well--which seems to have other problems outside the scope of this issue.

stevstrong:
Obviously this is solved by #616 
If you consider that it is not, you can still reopen this issue.

PR #616, opened 2019-03-30
Fixed incorrect arguments for stlink_upload

timsavage:
The number of arguments being passed to the stlink_upload script on
linux, linux64, macosx is one (this may have been more in the past but
the with version 1.8.8) this is now only a single argument.

The change changes the variable referenced in the script from $4 to $1.

This change already appears to have been applied to the windows version
of the script.

This appears to be open issue also #604

### DIFF
--- a/tools/linux/stlink_upload
+++ b/tools/linux/stlink_upload
@@ -1,5 +1,5 @@
 #!/bin/bash
-$(dirname $0)/stlink/st-flash write "$4" 0x8000000
+$(dirname $0)/stlink/st-flash write "$1" 0x8000000
 exit 0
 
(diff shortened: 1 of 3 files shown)


rogerclarkmelbourne:
It was caused by this change I made

https://github.com/rogerclarkmelbourne/Arduino_STM32/commit/0f868ecb7eb1a104de2bcfa45c294e15fd745bdd#diff-5ae7354fa4b9d8739cc748960851d64b

Where I forgot to update the linux files to match the changes to platform.txt (and the Windows STLink upload)

rogerclarkmelbourne:
Thanks

PR #624, opened 2019-04-08
Allow stlink_upload to work with home dir names containing spaces on Mac OS-X

Troynica:


stevstrong:
I cannot check this as I don't have any MAC.
Can someone please double check this?

rogerclarkmelbourne:
I do have a mac, but I don't use it for Arduino development.

The PR looks slightly problematic, because it appears to use a different argument number.

I think the author needs to explain their changes.

PR #630, opened 2019-05-09
Handle strchr_P alias for AVR code compat

tpruvot:
This function is used in some projects (Marlin) and require to be defined


stevstrong:
Compiles error free so I merge this.
Thanks!

PR #632, opened 2019-05-13
Phonog phonog patch 1

InnovativeIdeas:
pull request  phonog CAN


stevstrong2:
@InnovativeIdeas ,
do we really need those 3 files (usb.c, rcc.h, rcc._f1.c) under STM32F1/libraries ?
I think they are not placed correctly.

stevstrong:
OK, I think the changes do not affect the core functionality, they are only extend it.
So I will merge this.

PR #640, opened 2019-06-13
bugfix:HardwareSerial check framing or parity error

ag88:
Currently HardwareSerial uart does not check if a framing error
(parity error or invalid stop bit) has occured and simply
forward the received byte to the caller. This results in spurious
invalid data being received e.g. when the uart pin is left floating
or when a connected device is reset. This confuses apps using uart,
and cause apps to abort with errors.

this fix checks received bytes for framing error (parity error or
stop bit errors) and discards them if a framing error occured


stevstrong:
Seems good so I merge this.
Thanks.

ag-123:
thanks !

PR #653, opened 2019-07-28
SPI: fix annoying unused variables warnings

tpruvot:
- use const for ff variable to avoid unused variable on each SPI.h include
- move the spi_this refs where its used...
- and also the 3 others ones :

STM32F1\libraries\SPI\src\SPI.cpp:784:12
  warning: enumeration value 'RCC_AHB' not handled in switch [-Wswitch]

STM32F1\libraries\SPI\src\SPI.cpp:392:5:
  warning: this 'while' clause does not guard... [-Wmisleading-indentation]

Sample use : https://travis-ci.org/MarlinFirmware/Marlin/jobs/564740480


stevstrong:
Seems good, so I merged this.
Thanks.

PR #656, opened 2019-08-08
Code cleanup and removed compiler warnings about while loops

saloid:


stevstrong:
Looks good. In addition I would move the function to the beginning of the file where it is declared.

saloid:
I am trying to follow code style - so it near other same functions, which are declared on top and defined at bottom.

stevstrong:
Except that this is an inline function.
Have you checked that it is really compiled as such?

saloid:
I have compiled with and without "inline" - so file

stevstrong:
Try to move the whole inline function (without any other declarations) to the header file, then it will be for sure inlined.

stevstrong:
I merged this as it seems to make the core more readable.
Another commit of SPI.h will follow to complete this PR.

PR #663, opened 2019-08-29
Added missing extern C to STM32ADC helper files

ksuszka:
when I tried to use STM32ADC::setChannels method I got an error about missing function implementation from util_adc.c file, which was caused by missing extern "C" declaration in util_adc.h file.


stevstrong:
It looks good and compiles error free, so I merge this.

PR #672, opened 2019-09-11
Remove errant "init" of extern

thinkyhead:
Fix a compiler warning that an `extern` cannot be set in its declaration.


stevstrong:
I merged the PR as it is, and kept the variable declaration, it eventually may be used in the future.

thinkyhead:
Thanks! One less warning is nice.

PR #674, opened 2019-09-14
Increment size of emulated EEPROM

svmac:
Defines paremeter Pages for use more than one page per bank on emulated EEPROM.


stevstrong:
Do you have a simple test sketch to demonstrate the functionality and the advantages it brings?

svmac:
Yes, I have already modified the example, look at it. Option 1 uses two pages.

stevstrong:
How far is this compatible with the previous version and what problem is solved with this?

svmac:
It is 100% compatible, because Pages are set to 1 by default, so if you do not use (reference) Pages then it is like the previous version.
Solve when you need more EEPROM storage.

stevstrong:
Looks good, so I will merge this.

PR #675, opened 2019-09-14
Add frequency to hardware i2c

jjbubka:
Tested with Naze32 @72Mhz & BMP280 working at 1.2Mhz
It doesn't break existing code


jjbubka:
In the conflict
ASSERT(!(dev->regs->CR1 & I2C_CR1_PE));

must be changed by
if(dev->regs->CR1 & I2C_CR1_PE) return;

This was changed in a newer commit. It has passed more than a year...

Sorry for adding garbage to the old PR #517

stevstrong:
It compiles error free and it seems not to break anything, so I will merge this.

PR #676, opened 2019-09-15
USBComposite: fix circular dependency

tmp64:
*USBComposite.h* includes *USBMIDI.h*, *USBMassStorage.h* and *USBXBox360.h*
which include *USBComposite.h*.

Include guard in *USBComposite.h* prevents compile errors, but it
causes Qt Creator's clang code model to show errors and stop working.


stevstrong2:
I have tested the jigglemouse example, it builds error free.

PR #677, opened 2019-09-17
fix: remove warning __always_inline not always possible

stevstrong2:
replaces "static __always_inline" with "inline" directive

### DIFF
--- a/STM32F1/cores/maple/ext_interrupts.h
+++ b/STM32F1/cores/maple/ext_interrupts.h
@@ -106,7 +106,7 @@ void detachInterrupt(uint8 pin);
  *
  * @see noInterrupts()
  */
-static __always_inline void interrupts() {
+inline void interrupts() {
     nvic_globalirq_enable();
 }
 
@@ -120,7 +120,7 @@ static __always_inline void interrupts() {
  *
  * @see interrupts()
  */
-static __always_inline void noInterrupts() {
+inline void noInterrupts() {
     nvic_globalirq_disable();
 }
 
--- a/STM32F1/cores/maple/libmaple/exti.c
+++ b/STM32F1/cores/maple/libmaple/exti.c
@@ -250,7 +250,7 @@ __weak void __irq_exti15_10(void) {
  * won't actually be cleared in time and the ISR will fire again.  To
  * compensate, this function NOPs for 2 cycles after clearing the
  * pending bits to ensure it takes effect. */
-static __always_inline void clear_pending_msk(uint32 exti_msk) {
+inline void clear_pending_msk(uint32 exti_msk) {
     EXTI_BASE->PR = exti_msk;
     asm volatile("nop");
     asm volatile("nop");
@@ -258,7 +258,7 @@ static __always_inline void clear_pending_msk(uint32 exti_msk) {
 
 /* This dispatch routine is for non-multiplexed EXTI lines only; i.e.,
  * it doesn't check EXTI_PR. */
-static __always_inline void dispatch_single_exti(uint32 exti) {
+inline void dispatch_single_exti(uint32 exti) {
     voidArgumentFuncPtr handler = exti_channels[exti].handler;
 
     if (!handler) {
@@ -270,7 +270,7 @@ static __always_inline void dispatch_single_exti(uint32 exti) {
 }
 
 /* Dispatch routine for EXTIs which share an IRQ. */
-static __always_inline void dispatch_extis(uint32 start, uint32 stop) {
+inline void dispatch_extis(uint32 start, uint32 stop) {
     uint32 pr = EXTI_BASE->PR;
     uint32 handled_msk = 0;
     uint32 exti;
--- a/STM32F1/system/libmaple/dma_private.h
+++ b/STM32F1/system/libmaple/dma_private.h
@@ -37,8 +37,8 @@
 /* Wrap this in an ifdef to shut up GCC. (We provide DMA_GET_HANDLER
  * in the series support files, which need dma_irq_handler().) */
 #ifdef DMA_GET_HANDLER
-static __always_inline void dma_irq_handler(dma_dev *dev, dma_tube tube) {
-
+inline void dma_irq_handler(dma_dev *dev, dma_tube tube)
+{
     void (*handler)(void) = DMA_GET_HANDLER(dev, tube);
     if (handler) {
         handler();
--- a/STM32F1/system/libmaple/include/libmaple/libmaple_types.h
+++ b/STM32F1/system/libmaple/include/libmaple/libmaple_types.h
@@ -57,13 +57,6 @@ typedef void (*voidArgumentFuncPtr)(void *);
 #define __packed __attribute__((__packed__))
 #define __deprecated __attribute__((__deprecated__))
 #define __weak __attribute__((weak))
-#ifndef __always_inline
-#if __GNUC_PREREQ__(3, 1) || (defined(__INTEL_COMPILER) && __INTEL_COMPILER >= 800)
-#define	__always_inline	__inline__ __attribute__((__always_inline__))
-#else
-#define	__always_inline
-#endif
-#endif
 #ifndef __unused
 #define __unused __attribute__((unused))
 #endif
--- a/STM32F1/system/libmaple/include/libmaple/nvic.h
+++ b/STM32F1/system/libmaple/include/libmaple/nvic.h
@@ -109,14 +109,14 @@ void nvic_sys_reset();
 /**
  * Enables interrupts and configurable fault handlers (clear PRIMASK).
  */
-static __always_inline void nvic_globalirq_enable() {
+inline void nvic_globalirq_enable() {
     asm volatile("cpsie i");
 }
 
 /**
  * Disable interrupts and configurable fault handlers (set PRIMASK).
  */
-static __always_inline void nvic_globalirq_disable() {
+inline void nvic_globalirq_disable() {
     asm volatile("cpsid i");
 }
 
--- a/STM32F1/system/libmaple/include/libmaple/usb_cdcacm.h
+++ b/STM32F1/system/libmaple/include/libmaple/usb_cdcacm.h
@@ -172,7 +172,7 @@ int usb_cdcacm_get_n_data_bits(void); /* bDataBits */
 
 void usb_cdcacm_set_hooks(unsigned hook_flags, void (*hook)(unsigned, void*));
 
-static __always_inline void usb_cdcacm_remove_hooks(unsigned hook_flags) {
+inline void usb_cdcacm_remove_hooks(unsigned hook_flags) {
     usb_cdcacm_set_hooks(hook_flags, 0);
 }
 
--- a/STM32F1/system/libmaple/stm32f1/include/series/gpio.h
+++ b/STM32F1/system/libmaple/stm32f1/include/series/gpio.h
@@ -486,7 +486,7 @@ typedef exti_num afio_exti_num;
 /**
  * @brief Deprecated. Use exti_select(exti, port) instead.
  */
-static __always_inline void afio_exti_select(exti_num exti, exti_cfg port) {
+inline void afio_exti_select(exti_num exti, exti_cfg port) {
     exti_select(exti, port);
 }
 
--- a/STM32F1/system/libmaple/stm32f1/include/series/spi.h
+++ b/STM32F1/system/libmaple/stm32f1/include/series/spi.h
@@ -75,13 +75,13 @@ extern void spi_config_gpios(struct spi_dev*, uint8,
  * @brief Deprecated. Use spi_config_gpios() instead.
  * @see spi_config_gpios()
  */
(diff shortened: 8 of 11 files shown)


rogerclarkmelbourne:
Hi Steve
I'm not sure why you submitted this PR but didn't merge it.

Do you want me to test it ?

I don't know anything about the differences between __always_inline and inline, or why inline is better.

I presume inline is better, because its a standard C language modifier, rather than __always_inline, which is probably a compiler specific directive.

IMHO. Using standard C language modifiers would be preferable if they work OK, but the only way to know would be to compile and then test it on some real hardware.

stevstrong:
I have tested it on blue-pill with two applications, using SPI with DMA, IO interrupts and timer interrupts.
It did work as expected, so I am confident that this will not break anything.

Additional info about "static inline" vs "inline" can be found here: https://gist.github.com/htfy96/50308afc11678d2e3766a36aa60d5f75 (also check the link at the bottom of the page).

Does anything speaks against merging it?

stevstrong:
This intends to complete #551  .

rogerclarkmelbourne:
I think its OK to merge.

I probably can not test it until the weekend, work is very busy for me at the moment.

dewhisna:
This PR seems to have broken some things.  Without this PR my project compiles, but with the aforementioned warnings.  However, with it, I strangely get the following link errors:

```
.pio/build/genericSTM32F103VG/libFrameworkArduino.a(dma_f1.c.o): In function `__irq_dma1_channel1':
dma_f1.c:(.text.__irq_dma1_channel1+0x6): undefined reference to `dma_irq_handler'
.pio/build/genericSTM32F103VG/libFrameworkArduino.a(dma_f1.c.o): In function `__irq_dma1_channel2':
dma_f1.c:(.text.__irq_dma1_channel2+0x6): undefined reference to `dma_irq_handler'
.pio/build/genericSTM32F103VG/libFrameworkArduino.a(dma_f1.c.o): In function `__irq_dma1_channel3':
dma_f1.c:(.text.__irq_dma1_channel3+0x6): undefined reference to `dma_irq_handler'
.pio/build/genericSTM32F103VG/libFrameworkArduino.a(dma_f1.c.o): In function `__irq_dma1_channel4':
dma_f1.c:(.text.__irq_dma1_channel4+0x6): undefined reference to `dma_irq_handler'
.pio/build/genericSTM32F103VG/libFrameworkArduino.a(dma_f1.c.o): In function `__irq_dma1_channel5':
dma_f1.c:(.text.__irq_dma1_channel5+0x6): undefined reference to `dma_irq_handler'
.pio/build/genericSTM32F103VG/libFrameworkArduino.a(dma_f1.c.o):dma_f1.c:(.text.__irq_dma1_channel6+0x6): more undefined references to `dma_irq_handler' follow
.pio/build/genericSTM32F103VG/libFrameworkArduino.a(exti.c.o): In function `dispatch_single_exti':
exti.c:(.text.dispatch_single_exti+0x1c): undefined reference to `clear_pending_msk'
.pio/build/genericSTM32F103VG/libFrameworkArduino.a(exti.c.o): In function `dispatch_extis':
exti.c:(.text.dispatch_extis+0x20): undefined reference to `clear_pending_msk'
.pio/build/genericSTM32F103VG/libFrameworkArduino.a(usart_f1.c.o): In function `__irq_usart1':
usart_f1.c:(.text.__irq_usart1+0x6): undefined reference to `usart_irq'
.pio/build/genericSTM32F103VG/libFrameworkArduino.a(usart_f1.c.o): In function `__irq_usart2':
usart_f1.c:(.text.__irq_usart2+0x6): undefined reference to `usart_irq'
.pio/build/genericSTM32F103VG/libFrameworkArduino.a(usart_f1.c.o): In function `__irq_usart3':
usart_f1.c:(.text.__irq_usart3+0x6): undefined reference to `usart_irq'
.pio/build/genericSTM32F103VG/libFrameworkArduino.a(usart_f1.c.o): In function `__irq_uart4':
usart_f1.c:(.text.__irq_uart4+0x6): undefined reference to `usart_irq'
.pio/build/genericSTM32F103VG/libFrameworkArduino.a(usart_f1.c.o): In function `__irq_uart5':
usart_f1.c:(.text.__irq_uart5+0x6): undefined reference to `usart_irq'
```

Don't know that it matters, but this particular project is using USART, but not DMA.  And I also am getting the following related warnings, which I think are because these switched from `static __always_inline` to just `inline`, and I think you want them to be `static inline`.

```
In file included from /home/dewhisna/.platformio/packages/framework-arduinoststm32-custom/STM32F1/cores/maple/libmaple/dma_f1.c:40:0:
/home/dewhisna/.platformio/packages/framework-arduinoststm32-custom/STM32F1/system/libmaple/dma_private.h:45:6: warning: 'dma_clear_isr_bits' is static but used in inline function 'dma_irq_handler' which is not static
      dma_clear_isr_bits(dev, tube); /* in case handler doesn't */
      ^~~~~~~~~~~~~~~~~~
In file included from /home/dewhisna/.platformio/packages/framework-arduinoststm32-custom/STM32F1/cores/maple/libmaple/usart_f1.c:37:0:
/home/dewhisna/.platformio/packages/framework-arduinoststm32-custom/STM32F1/system/libmaple/usart_private.h:64:22: warning: 'rb_remove' is static but used in inline function 'usart_irq' which is not static
             regs->DR=rb_remove(wb);
                      ^~~~~~~~~
/home/dewhisna/.platformio/packages/framework-arduinoststm32-custom/STM32F1/system/libmaple/usart_private.h:63:14: warning: 'rb_is_empty' is static but used in inline function 'usart_irq' which is not static
         if (!rb_is_empty(wb))
              ^~~~~~~~~~~
/home/dewhisna/.platformio/packages/framework-arduinoststm32-custom/STM32F1/system/libmaple/usart_private.h:57:13: warning: 'rb_push_insert' is static but used in inline function 'usart_irq' which is not static
             rb_push_insert(rb, (uint8)regs->DR);
             ^~~~~~~~~~~~~~
Compiling .pio/build/genericSTM32F103VG/FrameworkArduino/libmaple/usb/stm32f1/usb.c.o
In file included from /home/dewhisna/.platformio/packages/framework-arduinoststm32-custom/STM32F1/cores/maple/libmaple/usart_private.c:33:0:
/home/dewhisna/.platformio/packages/framework-arduinoststm32-custom/STM32F1/system/libmaple/usart_private.h:64:22: warning: 'rb_remove' is static but used in inline function 'usart_irq' which is not static
             regs->DR=rb_remove(wb);
                      ^~~~~~~~~
/home/dewhisna/.platformio/packages/framework-arduinoststm32-custom/STM32F1/system/libmaple/usart_private.h:63:14: warning: 'rb_is_empty' is static but used in inline function 'usart_irq' which is not static
         if (!rb_is_empty(wb))
              ^~~~~~~~~~~
/home/dewhisna/.platformio/packages/framework-arduinoststm32-custom/STM32F1/system/libmaple/usart_private.h:57:13: warning: 'rb_push_insert' is static but used in inline function 'usart_irq' which is not static
             rb_push_insert(rb, (uint8)regs->DR);
             ^~~~~~~~~~~~~~
```

I'm also confused by them being called IRQ handlers.  I didn't take time to dig into that area of the code and so maybe they are just used by an interrupt handler, but if they are direct-called interrupt handlers that are in the interrupt vector table, I don't think those can be static or inline.  So these must just be indirect functions of some sort, since it worked before... just confusing to see...

So based on the warnings and link errors, it seems these need to be `static inline` functions instead of just `inline`.  That way, the compiler can decide if it wants to inline them or not.  The problem causing the warning was the attribute to "force" the compiler to always inline it and it deciding that it couldn't, not the static declaration.

And FWIW, this was using gcc version 7.2.1.

stevstrong2:
Hm, I cannot find in the repo the VG variant.
Where is this coming from?
And I see you are using PlatformIO. Officially, only Arduino IDE is supported.
Me I can compile SPI DMA sketches error and warning free in Arduino for blue pill.

dewhisna:
The VG is a custom variant I have for this other project that's using the STM32F103VG.  This issue has nothing to do with the target variant and would occur with any variant.  It's also not specific to PlatformIO.  You are still going to have the same problem eventually with the Arduino IDE, especially when they finally update to a modern compiler -- 4.8.3 is absolutely ancient.  I'm usually running even a 7.3.1 version, instead of 7.2.1 and that's even getting a bit long in the teeth now.

The problem here is that each piece -- the core, the variant, the framework, each Arduino library, etc., even the main app/sketch, are individually compiled into static-link libraries as separate sub-build targets.  Functions that are static are only emitted within the library and not are not visible outside the library (i.e. they aren't exported from the library).  So if you have a function calling other static functions in the same library target, then all is well.  But if one of those functions (in this case, the ones you stopped making static) are publicly visible from the library, then those functions aren't included in the final binary at all if the entire library isn't included.

And in this case, since I'm not including DMA, etc, those library pieces don't get included in my final binary, and since they aren't included, then the other locations that had their static functions calling it, now won't have a visible function to call after the linking (because it gets dropped in the linking process).

The Arduino IDE does a nearly identical division of the pieces into static libraries as well.  You probably aren't seeing the linking issues there only because the order in which the libraries are linked can affect it.  So the order it happens to be using on yours is coincidentally not causing you these linking issues.  But you will eventually have these same issues there too.

The `__always_inline` compiler attribute is what was causing the original warning of the compiler saying that inlining isn't always possible, not them being `static` or being `static inline`.  Having them as either `static inline` or `inline` only (without the special gcc attribute) would make the original warning go away, as then it's up to the compiler to decide and inline them if it can.

So you need to either make these `static inline` to match the other static functions that are also calling them (probably the best fix since these are in include files across multiple sub-build targets) or find all of the other functions calling them and make them non-static too so that they are subject to the same linking rules and then figure out how to ensure that the libraries are linked in the correct order.

I'm also a little concerned that we are all of a sudden starting to see the compiler complain that it can't always inline these functions, when some are clearly only accessing a few local variables and CPU registers.  I know this `__always_inline` attribute wasn't an issue at all back on 60d982ff25d966f15e8af70d4bb6f46bb1d1595a.  So something between that commit and these later ones is causing the compiler some grief and possibly emitting more code and consuming more flash than is necessary.

stevstrong:
I don't think that excluding the DMA function from my sketch will change anything on this.
Arduino IDE will compile all the C/C++ files and will automatically include all the header files necessary to compile them.
I will check a sketch with SPI and serial 1 using gcc 7.2.1., but I honestly cannot imagine that the compiler will differ in this amount (to generate an error where there was not even a warning before) from one version to another.

I am almost sure that this issue is generated by PlatformIO, by the compile and linking parameters which differs from Arduino.
So I would like to ask you to check your project (or at least a simple sketch) with Arduino IDE and tell me the result.

dewhisna:
Excluding DMA from your sketch has everything to do with this.  If you exclude DMA, then it won't be linking the DMA library pieces and you will most likely see the same link error I am with having the DMA irq functions missing, even on gcc 4.8.3.  It's a problem on both compilers, not just 7.2.1 -- and not actually a problem with the compiler, but the code.  Only 7.2.1 is smart enough to produce a warning first (as I listed in the above comments) that 4.8.3 doesn't, to tell you what's about to fail and why.

So please do try yours without DMA....  In the meantime, I'll try to get the Arduino IDE working here to build something with this to prove that it exists there too and with the older compiler.  Though it may take me a while to set all of that up.  I can't stand the Arduino IDE and so never use it.  I can't stand PlatformIO either -- or specifically the Microsoft VSCode IDE that it integrates with.  Qt Creator is a MUCH better IDE than those two.  However, the Python scripts of PlatformIO are great for keeping up with library content and corresponding hashes and checkout points allowing you to properly version control things (including every single library you use and all of their dependencies) to have repeatable builds, so that you don't have the "smash-it bash-it mayhem" of the Arduino IDE wherein you can compile things today but not tomorrow because something changed that wasn't tracked in version control on your project or that was installed on one computer but not another.  Unlike the Arduino IDE, each project on PlatformIO can pick the specific versions of specific libraries, etc, that it needs and it's stored in version control with your project and staged in your temp folder build area so that one project doesn't affect another and isn't affected by what other things you happen to install on the computer.

stevstrong:
I just tried a test sketch with SPI and Serial1 using gcc 7.2.1.
Result: no errors, no warnings.
So your issue must have something to do with PlatformIO.

DMA can only be separated from being compiled when you deselect `USE_DMA` within the SPI driver.
But even if I would do that, it does not change anything on that the Serial 1 part is also compiling fine, so that your argument cannot be followed.

I would suggest you to compile a test sketch with Arduino IDE and compare the output (the compiled modules and compiler switches) withe those used by PlatformIO. This way you can figure out the root cause of your issue.

I agree with you, Arduino is a terribly bad IDE (I use Notepad++, and check the Arduino option "use external editor"), but offers an easy way to compile the sketch or your whole project.
This is why this core is build around this IDE (as the name is reflecting that, too).

stevstrong2:
OK, hold on, it seems that my previous test was not quite correct, the modules were not compiled.
Now I can reproduce the issue with 7.2.1.
Strange.

PR #679, opened 2019-09-19
WireSlave Library Rework to make I2C Slave Fully Functional

### LINKED ISSUE #207: I2C/wire library: slave is not implemented
Just for the record, because from time to time there are the same questions and troubles in the forum:
No, we do not have a working slave configuration for I2c  (onReceive()....) yet.

Maybe an approach could be:
http://forums.leaflabs.com/forums.leaflabs.com/topic2429.html?id=1543

### LINKED ISSUE #613: problem with IIC2
error while uncommenting the below line in i2c scanner example....
please fix it asap.................
and also show us a way on how to use I2C2 LINE with Wire_slave.h
//TwoWire WIRE2 (2,I2C_FAST_MODE);

error that i'm getting is :
'WIRE2' was not declared in this scope

### LINKED ISSUE #625: can some one please suggest how to use I2C2 with <wire_slave> library
The example under i2c <Wire_slave.h>  for I2C2 Is giving compilation error..and including <SoftWire.h> along with <Wire_slave.h> is giving redeclaration error
could nt find how to use I2C2 with <Wire_slave.h>

dewhisna:
Rework the WireSlave library of the STM32F1 target, merging it with the core, and fixing all of the I2C quirks and problems.

This fixes #625 and fixes #613 and fixes #207 and possibly other open issues.

I have tested the new WireSlave code and I2C core enhancements as both a master, communicating with CAT24C256 EEPROMs and MCP4725 DACs, and as a slave, communicating with a STM32F746G Discovery board and with the [I2CDriver](https://i2cdriver.com/) test gadget (a very handy tool, BTW).

The WireSlave examples have also been updated where needed and test compiled.

This code has already been merged and retested with the current master at baea09a2c094430199602f0c9e5964c039d7b6d6 to pick up the other changes related to the I2C frequency selection and apply similar functionality to WireSlave as was added to Wire and is ready for integration via a simple fast-forward merge.

This one certainly took longer than I had expected, as I didn't realize I would need to rewrite so much of it to make things work.


stevstrong:
Donna,
many thanks for the huge effort.
I will shortly test a master + slave sketch (is there such an example?) and commit it if everything is ok.

dewhisna:
You are welcome.  I had to rework it anyway since I have to get this new board up running.  And since I was using code others had contributed, it only seemed responsible to share my work too.  Though I should have rebased my code last week when I started on this rework and would have saved myself some merge conflicts last night.

Here's a minor "last-minute" addendum commit that makes the signature of `TwoWire` in `WireSlave` a little more like the one in `Wire`, which may make things a little easier in the future if there's a desire to merge the two into one library.  It should also make it easier for users to switch between the two libraries.

As for example sketches for testing... Hmm...  I've mainly been testing with the code for my motor control board that I vaguely mentioned in #625.   While it was a great test case for this, it would hardly be trivial for you to replicate, as the STM32F103 on the new motor controller communicates as a slave over I2C to a STM32F746G Discovery board that is serving as the master and controlling a much larger system with lasers and many more gadgets.  All of that I already had in place back when I was using the ATmega328PB on the motor control board.  So the code and hardware was there and waiting on a functioning I2C slave.

There are some examples in the `WireSlave` library, such as `selftest1`, that implements a master on one I2C and a slave on the other.  While I checked that it compiles and that it looked like somewhat sain code, I didn't actually test it on hardware.  But if you have a blue-pill board or something lying around, you could tie the two I2C ports together and add a couple pull-up resistors, and that would probably work well as a simple test.

dewhisna:
Actually, as I think more about it, I guess it won't help users switch between the two libraries, since that's a protected constructor, but it certainly should help with a future merge of the `Wire` and  `WireSlave` libraries.  These constructors are protected because the hardware state is important and having multiple simultaneous objects trying to control the same hardware would cause problems.  There would need to be an additional class inserted to serve as an interlock between "virtual I2C ports" and "physical I2C ports".  But that's for a future "free-time" project.

TheKikGen:
Just to add my 2 cents...I fully tested that version in a complex I2C project.  
No regression at all and much stable as said.

TheKikGen:
Dewhisna.  Stable : that what I said with some part of my project. But when I tested with a 'more real time' running condition (midi requires real time) , I mean without delays between calls to trace them, it crashed quite rapidly.  It is not the case of the Stm32Duino ware_slave head version currently.  I tested exactly the same code without any modification on both.  I attach a master and a slave I , hoping you will be able to reproduce that. Master and  slave  hardware are 2 bluepills (STM32F103C ).
[I2CMidi.zip](https://github.com/rogerclarkmelbourne/Arduino_STM32/files/3667754/I2CMidi.zip)

dewhisna:
@TheKikGen can you elaborate on the test you describe?  Which side has issues?  The Master?  The Slave?  Or both?  I also see a ton of debug prints in your code which is going to severely slow down communications and is likely to interfere with the timing.  It seems you are trying to debug print much more data than your main I/O stream itself is even sending.  Standard MIDI is designed for 31250 baud and your serial debug stream is only 115200, which is just shy of 4 times standard MIDI speed.  With the amount of debug data you are printing for every message, there's no way it could keep up with it.  I'm not sure that even increasing the debug channel to 1mbps is enough.  So my question is have you tried it without all of the debug prints??

You also say that "It is not the case of the Stm32Duino wire_slave head version currently", but you don't say which repo that's from.  It can't be this repo, as the original `WireSlave` library in this repo didn't work at all in Slave mode and only accidentally worked in Master mode on `I2C1` (but not `I2C2`) due to the multiple definitions of the `i2c_dev` structures [that I described here](https://github.com/rogerclarkmelbourne/Arduino_STM32/issues/625#issuecomment-530407251) with one structure stepping on top of the other.  So what code base are you saying it works with??

You say that it "crashed quite rapidly".  What does that mean?  That it just couldn't keep up?  Did the code lock up with a stack/heap overflow or machine check?  Have you tried disabling all of the debug prints where you are debug-printing so much more data than your main I/O stream is even sending?

I need more details about what you are trying and exactly what is and isn't working to have any chance in trying to replicate it.

TheKikGen:
@dewhisna.  At first, you could simply run the example I provided in my last post with the official stm32duino core (the repo you forked on which you posted that pull request)  vs your version and understand why one is stable  and not the other.  Simply let it run... 
Code is simple enough. Goal from the POC  is reached . I'm trying to send midi msg via I2C from a master to a slave.  The point is the stability of I2C. 
Concerning "Debug print" : almost all commented if you check the code,, at the exception of some,  outside ISR whatever.   The code compile and works with the current core https://github.com/rogerclarkmelbourne/Arduino_STM32 if you only include wire_slave (maybe accidentally...).    I use I2C1 ( the default one ).
Concerning the baud rate speed point : I'm debugging on a Serial USB at 115200 bauds  2 midi notes on note off  delayed msg sent and received by a very slow midi serial rate of 9600 bauds. Not talking about the 100hz of I2C....Absolutely no contention at all. I already did that on other hardware in the past.... 
The only serious point could be to print in the 2 ISR Receive/RequestEvent . But sometimes, it is necessary for debugging purpose (and commented anyway as said).  I tested obviously without any printing. Same issue.   
I was faced my self with the multiple definitions of i2c_dev, and at a higher level, the confusion existing between wire and wire_slave. So I understand quite well your point.  It is why your very welcome pull request has got my full attention back !

dewhisna:
@TheKikGen .... there is one possibility I thought of.  Unlike the original `WireSlave` code, which can hang indefinitely waiting on the other side, my new code actually supports the timeout logic correctly.  And, it uses the underlying timeout functionality in the base `Stream` class, meaning that it defaults to 1000 milliseconds (or 1 second), instead of waiting forever.

Perhaps you need a longer timeout?  I would suggest first trying setting it to the original timeout of the old code to see if the old functionality returns.  So add:

```
Wire.setTimeout(0);
```

in your `setup()` function on both sides.  That should disable the timeout to be like the original code.

If that restores the functionality, then it means there is a timing issue somewhere and perhaps the default 1-second timeout isn't sufficient (even a few debug prints could push it over that limit when the serial buffer gets full).

As for the stack/heap question, I was trying to figure out what exactly is "crashing" for you.  While you probably aren't using dynamic memory allocation, the `WireSlave` code does.  I left that part of the original `WireSlave` somewhat intact, where it dynamically grows the master transmit buffer to be whatever you put in the buffer prior to calling `endTransmission`.  Meaning if you write too much data before calling `endTransmission`, it could run out of heap and collide with the stack.  Or if there is a bug of some sort in that part of the `WireSlave` code.  There is a similar dynamic receive buffer on the master `requestFrom`, as well.

I left those buffers on the master-side dynamic since the code in the master, if well written, can avoid memory overflows because it controls the calls to those functions.  The slave side, however, has a fixed receive buffer, as a rogue device transmitting data to a slave device shouldn't crash the slave.

So... I was asking about stack/heap collision and such to help in trying to troubleshoot where the problem is at.  As long as you aren't calling `write` or `print` too much before `endTransmission`, I doubt there is a memory issue.

I'm leaning toward the timeout functionality, since that part is different between the old and new code.  I left its initial value to be the default of the `Stream` class since that's the functionality that `Wire` in general (based on Arduino documentation) should be following -- that and a 1 second timeout is very reasonable for most applications and helps avoid stuck code.  But it could be that you need a slightly longer timeout to give the serial buffers time to get caught up on the debug prints or something.  So give that a go.

TheKikGen:
@dewhisna .  The Wire.setTimeout(0) on the master seems to stabilize I2C. No more hangs...
So I think it is probable that the timeout handling lets the bus in an unstable state...when it is triggered.
Seems also strange that 1s default  is not enough ??? I tried to set that timeout to any value, let's say 10000 (it is like "forever" for a STM32F103C at 72 Mhz) , it crashes again...Something is wrong in this area I think.
[I2CMidi_WireTimeout.zip](https://github.com/rogerclarkmelbourne/Arduino_STM32/files/3672154/I2CMidi_WireTimeout.zip)

dewhisna:
@TheKikGen  Thanks for the update.  It's possible that I failed to reset something on the I2C peripheral in the timeout condition.  That path is a bit harder to test than the others.  I had based it on some code I did for the F4, which uses the same I2C peripheral as the F1 and it resumes OK there after a timeout, but it was built on top of a .NETduino based core instead of libmaple so a lot was different with the clocks and timeout logic in general.

So, let me do some investigation on the timeout path.  The STM peripherals are a bit insane (particularly this I2C one -- no wonder they've changed it on newer devices like the F7).  I probably forgot to clear some bit or reset interrupts or otherwise didn't hold my head at quite the right angle.

At least the `setTimeout(0)` is working for you so you can continue your development while I investigate.  Plus, this is good testing and seat-time for the PR before they merge it.  So thanks for testing it out!

TheKikGen:
@dewhisna You're welcome. A 150 µs timeout is often used (cf application notes of many I2C peripherals providers) as it the double of longest I2 C message at 100khz.  1s is a very very large one !  I'm aware of the bad I2C  implementation made by STM....

dewhisna:
@TheKikGen I think I might have found it -- at least why the bus is being left in a bad state.  It was the one part of the error path from the original `WireSlave` code that I didn't change.  It resets the bus on a protocol error, but not on the timeout condition.  So my guess is that the stop condition never comes and both sides are left "stuck".  The master won't start the next transmission because it thinks the peripheral is still busy (i.e. busy flag still set), and the slave is in limbo waiting on the master to send the stop (currently the timeout logic is only on the "master" side of the world, which means I probably need to enhance the slave side a bit more to handle timeouts too).

This still doesn't explain why you hit the original timeout, since as you point out, 1s is a large time and even more so in your 10s test, so I need to go through that code with a fine-toothed comb too.  Though this timeout isn't for a single byte, it's for the entire message transmission from the master to the slave including the time for the slave to send all of its ACK's, so should probably be larger than the time you cite from I2C specs for timeout.  The timer here starts when the transmission engages (i.e. interrupts enabled and the start-bit set) to start sending the slave address and data, and it doesn't stop until it either times out or the state transitions to done (or a protocol error happens) at the end of the entire message, and it includes the time in the slave for it to call its transmit and receive callbacks to process it too for it to send its ACK/NAK signals.  My guess is the slave is keeping it busy somehow in one of those callback functions or something -- and who knows, you might have found another bug somewhere in the slave code and its sending of the ACKs and NAKs.  Though I have beat on the slave code quite a bit in one of my projects here.

I have to do some shuffling on my local clone of the repo to get it back to where I can push updates on this branch (plus my day-job will get in the way), but later this evening or early tomorrow I will add some additional updates to this PR that will hopefully fix it.

dewhisna:
Actually, what I said is wrong.  The timeout is per I2C interrupt cycle.  Not the whole message.  I had forgotten that the `_i2c_irq_handler` function updates the timestamp.  So the timeout should be from one event to the next.  But the callback functions to get/put the data on the slave are still in the loop since the master is waiting for the ACK or NAK from the slave before proceeding.

dewhisna:
@TheKikGen  Hopefully these two additional commits will fix your issues.  I did beat on it a bit more with my project, which uses the STM32F103 as a coprocessor doing motor PID control as a slave to a STM32F746 master running a laser vision system, including fast motor speed acquisition over I2C.

While working on these last commits, I re-remembered why the slave mode doesn't do more with its timeout setting.  And that's because in slave mode, all communications is handled in the IRQs and driven by the master triggering events on the bus.  As such, it never hangs the main processing code on the slave device.  The IRQ functions will call the main code callbacks when it needs data, otherwise it's content to wait on the master at the hardware level.

The only time the timeout is used on the slave is when processing `flush()`, which is used for switching from slave mode to master mode by making sure that any pending slave transmit is completed before switching modes.  In theory, it should allow for a multimaster architecture and let a slave turn around and talk as a master to other slave devices on the bus where it was formerly a slave, and I did actually try that some here, and was successful at having my F103 read the EEPROM located on the F746's I2C bus where the F103 normally serves as a slave.  But, I didn't run any bus-contention resolution tests.

As a side note -- you have me curious what your MIDI device is, as another one of my projects using this WireSlave library is a MIDI interrupter for my Musical Tesla Coil.  However, there the I2C is only used for the display and is always a master.  The MIDI all happens either over USB or USART or via SD Card playback.

TheKikGen:
@dewhisna  Here we go ! No more hangs with the default timeout (i.e. not using the setTimeout(0) method anymore) !   2 words about my project "Midi over I2C" project.  
I already developed a usb midi firmware 3IN/3OUT allowing deep midi routing , and designed a "midi shield" PCB  for the Bluepill.   I want to to extend the midi capacity of to 15 IN/ 15 OUT with 5  Bluepill boards communicating via I2C (the USB Midi standard is limited to 16 cables anyway) .   
The master is the only one connected to usb and dispatch and collect midi msg to/from all of the others slaves, apply routing rules, etc...
Currently, I'm testing 3 devices (1 master , 2 slaves) with a quite aggressive midi clock of 300bpm  from an USB apps  using your version. I will let it run all night long... Let the music play on
On verra demain matin. French time.

dewhisna:
I'm so glad to hear that it's working now!  Thanks again for testing it.  Maybe a few more will join in testing as well.  I'm not sure how much testing they need before landing this PR.
Et oui, jusqu'à demain matin.

dewhisna:
@TheKikGen ... After some experimenting tonight, I think there's a race condition between the `wait_for_state_change` function and the I2C interrupts, particularly with the timestamp for the timeout.  I was seeing what appeared to be glitches at approximately the 16-bit rollover point in the timeout when doing large block transmissions to this I2C display.   This might be what initially triggered the timeout issue for you as well -- not a real timeout, but an artifact of the race condition.  I'm betting you also see small random glitches in your overnight test too, where it will occasionally think the communications times out.

I'm working on a fix.  In fact, I have a fix, but need to do more testing before sending it your way.

TheKikGen:
@dewhisna Exact.  My tests showed also that. The communication slowed down due to the timout. Re-tested again with a setTimeout(0). No issue .

dewhisna:
@TheKikGen -- OK, this last commit should fix the race condition in `wait_for_state_change`.  I think this was why the original I2C code didn't implement any of the timeout logic.  They just had a comment to fix the rollover and the race condition, but left the broken code and only had the timeout set to zero.  They technically also had a race condition for the very state variable they were testing, but "accidentally" got away with it because the enum values for the state fit in a single byte.

However, it's good we had this race condition issue because it let us find that previous bug that would cause the hang.  That one would have been very difficult to find without something false triggering the timeout.

Everything seems to be working great at the moment and I think this code is good enough to merge to master once they give it the green-light.

TheKikGen:
@dewhisna . Hi Donna.  I still got some strange things with your last commit. I modified my test master and slave, and let the default timeout as is (1sec).
But if i measure the time before the RequestFom and a return value = 0 (meaning a timeout), the time is like this :
Timeout count packet : t=1
Timeout get packet : t micros =629
Timeout get packet : t micros =629
Timeout get packet : t micros =511
Timeout get packet : t micros =83
etc...in some situation, this slows down the comm with the bus...
The setTimeout(0) still works as a workaround.
[I2CMidi-Wire_slave_timeout2.zip](https://github.com/rogerclarkmelbourne/Arduino_STM32/files/3677884/I2CMidi-Wire_slave_timeout2.zip)

dewhisna:
Now that one could be a legitimate timeout.  It means your slave took longer than 1 second to respond (i.e. get its data together).  If it's writing to serial or a display or something, I could certainly see that happening on occasion.  I have no reason to believe that this case is an issue with the i2c master code.

Maybe you legitimately need longer than a 1sec timeout?  As I mentioned earlier, that timeout (while normally byte-to-byte for ACK's on the bus) also includes the entire code execution path on the slave to respond and create the message to send, which happens during the ACK cycle with the slave address (which includes your debug prints and all too).

I see you have some active debug prints in your `I2C_SlaveRequestEvent` and that alone could cause a periodic timeout.  Try disabling those prints and see what happens.  If you want to print something as a result of the function call, set a global flag there and have your main `loop()` do the debug printing.  One important reason is that the `Serial` calls in that function may encounter a buffer full case and end up trying to block waiting on the serial interrupts to happen.  But, your event handlers are already inside an existing interrupt handler and so addition interrupts cannot happen until you exit.  It's important to not call any functions inside these event handlers that rely on interrupts running -- as the processor will currently have the interrupts disabled for the interrupt it's already processing.

Also, why are you calling `Wire.available` and `Wire.read` during the `Request` event??  Those should all happen in the `Receive` event.  You should capture all inbound data from the master in the `Receive` event and set a state-machine or flags to process in the `Request` event.  The `Request` event should only consists of transmitting data back to the master -- i.e. calls to `write` or `print`.  Your code isn't correctly formed.

Now, you can somewhat get away with it because this `WireSlave` code uses separate buffers and doesn't force-clear them until another inbound message comes.  But to be fully Arduino API compliant, `receive` should only receive/read data and `request` should only write/send data. 

Your code for calling `Wire.available` and `Wire.read` to get the command, etc, should be moved to your `I2C_SlaveReceiveEvent` event.  That code should set a variable to tell the `I2C_SlaveRequestEvent` function what to send when it gets called shortly thereafter.

TheKikGen:
@dewhisna.  I perfectly know how interrupts work. 
If we concentrate on the initial discussion concerning the timeout issue : the RequestFrom is called by the master and no ISR are used at the master level.  Why when I let the default timeout (1000ms), i receive a 0 return value from Wire.requestFrom before this delay in some cases ?   
Again something strange here.

The  "print" concerns only the slave, and only in a specific case for debugging purpose. it can't  be the cause of the erratic behaviour of the wire timeout on the master side. And remember this is a pure test program, not a production one, so a one shot printing in ISR can help if you know what you are doing.  And besides, Wire's examples are themselves full of print in ISR...not the best way to show how an interruption works !   Personally, I use ring buffers to store received messages in the full project, which I then read outside ISR for processing.

Your other points...

Concerning the logic : A command to a slave is always followed by a request, so why would this way of dealing with events necessarily be "badly formed"? Does the AVR Arduino Wire clear the buffer immediately after the receive ISR handler ?  I don't think so.  I however agree that  when you do not need to wait for a response the request event will never be triggered, and in that case, it is better to use a global volatile.  To keep it simple I will modifiy my tests pgm in that way, and will post here the result later.

dewhisna:
@TheKikGen 

> If we concentrate on the initial discussion concerning the timeout issue : the RequestFrom is called by the master and no ISR are used at the master level. Why when I let the default timeout (1000ms), i receive a 0 return value from Wire.requestFrom before this delay in some cases ?
> Again something strange here.

I think this means the slave is taking longer than 1sec to respond.  The master is sitting there spinning in the `wait_for_state_change` function, waiting for the slave to respond.  If the slave doesn't respond, its I2C interrupt never fires and the function times out waiting for the ACK or NAK from the slave.
 
> The "print" concerns only the slave, and only in a specific case for debugging purpose. it can't be the cause of the erratic behaviour of the wire timeout on the master side. 

It certainly can cause the master side to timeout.  The master spins waiting on either an ACK or NAK for the slave at the end of every byte sent.  If the slave takes more than the master's timeout, the master gives up and aborts the communications with the slave.  So yes, the prints on the slave side certainly affect the master.  The entire I2C bus is suspended until the slave responds to the master or the master times out and sends a `stop` signal to halt the attempt.

On the slave side, this `requestEvent` callback gets called during the IRQ on the slave when it is to about to send the ACK for its address.  It knows at this point that it's an incoming read from the master and it calls the `requestEvent` prior to sending its ACK for the address.  This suspends the master to wait for your data until you have written it in your `requestEvent` -- otherwise, your data wouldn't be ready by the time it's supposed to be sent.  If the code in your `request` callback takes too long, the master times out and thinks the slave isn't there at that address to even answer.

The `requestEvent` isn't called for every byte communicated.  It's called once at the beginning of the message and the callback needs to fill the buffer with the entire message to send as quickly as possible.

>And remember this is a pure test program, not a production one, so a one shot printing in ISR can help if you know what you are doing. And besides, Wire's examples are themselves full of print in ISR...not the best way to show how an interruption works ! Personally, I use ring buffers to store received messages in the full project, which I then read outside ISR for processing.
> 

Sure, other example code uses prints and such in interrupt handlers -- it doesn't mean it's a good example.  I even used some debug prints when initially developing these I2C interrupt handlers.  I only bring it up because it could certainly be the source of the timeouts you are seeing if it's causing your slave to not respond quickly enough to the address ACK that the master is trying to send.


> Your other points...
> 
> Concerning the logic : A command to a slave is always followed by a request, so why would this way of dealing with events necessarily be "badly formed"? Does the AVR Arduino Wire clear the buffer immediately after the receive ISR handler ? I don't think so. I however agree that when you do not need to wait for a response the request event will never be triggered, and in that case, it is better to use a global volatile. To keep it simple I will modifiy my tests pgm in that way, and will post here the result later.

That is just how the Arduino API is written.  If you look at the specs in their documentation, it's quite clear that `receive` should only read and `request` should only write.  It may not clear its buffer between the two calls.  It's up to the particular I2C implementation (and `Wire` library or `TwoWire` class, etc).  By the spec, it's certainly free to do so and may surprise you later if you ever port your code to another platform.

It also may depend on how the master is sending it -- i.e. whether it uses a repeated start signal or if it does a full-stop after the write followed by a start for the read.  The I2C implementation is certainly free to clear its buffers in the full-stop scenario.

TheKikGen:
@dewhisna Donna, back again !  
Same issue.  I have suspicious timeouts at the master level less than 1ms and some random error 4 in endTransmission.  

> I think this means the slave is taking longer than 1sec to respond
1 sec to process 4 bytes at 100Khz ? Impossible I didn't catch that.  The average time to ask for availability and request a 4 bytes packet (2 requests) from the slave is less than 1,5ms .

I set a 10sec timeout with the Wire.setTimeout (ridiculously long). I still got random timeout less than 1ms.   No print at all on the slave ( to avoid furthermore debate on this), only timeout report on the master : still got timeout beyond 1ms...and errors 4 (=others...).  
With 10sec, I shouldn't have any timeout..., The Wire.setTimeout function is not reliable. 
An issue is still here somewhere...

Our side discussion about reading in the requestEvent:
I checked the source code of the AVR I2C core : rx buffer is not resetted before the requestEvent. So reading is fully possible.  RX buffer is resetted only when new data are sent by the master, not during a requestFrom.  I find that very logical personally. Not reading the command byte could be a way to protect it...but I agree this is an unusual way to use Wire events.

dewhisna:
The "some random error 4 in endTransmission" could be a timeout.  Yes, it's defined as "other" but the Arduino API has no error code for timeout.  And since I was making it conform to the Arduino API, I didn't invent any new codes.  They define only four codes (five if you count "success" too):
    0:success
    1:data too long to fit in transmit buffer
    2:received NACK on transmit of address
    3:received NACK on transmit of data
    4:other error 

There is no code for "timeout" so 4 was used.  I didn't invent a new one because it wouldn't be standard and so existing code wouldn't recognize it.  However, 4 isn't exclusively "timeout".  It can also mean bus or arbitration error if something is electrically wrong with the bus.  That's the two definitions in this `WireSlave` library, but other library implementations could use 4 as a catchall for additional errors.

You say you are getting "random timeout less than 1ms" -- how do you know that's a timeout??  Are you saying that it returned within 1ms from `requestFrom` and had 0 bytes?  Is that what you are calling a timeout?

That could just mean that your slave didn't transmit any data.  As receiving 0 bytes doesn't necessarily mean a "timeout" condition.  If you write no bytes to the `Wire` in the `requestEvent` then nothing gets sent by the slave.  Or, if there is some electrical issue on the bus, it could have a premature stop condition.  You do have 4.7K (or so) pullup resistors on your SDA and SCL wires, right??  If not, that could contribute to it.  There are weak pullups in the micro, but they aren't strong enough and could lead to bus issues.

Unfortunately, the `requestFrom` function has no error return code -- only the length of bytes received.  And it's defined as an unsigned type, meaning that negative numbers can't even be used for errors.  This means that zero bytes received could be an error (timeout or other error) or it could be that the slave actually sent zero bytes back.  They aren't distinguishable.

There are lots of paths in the `requestEvent` of your slave code where it could return 0 bytes back to the master.  In fact, all of these cases when there's 1 byte in the `Wire` buffer:

```C++
  if (cmd == B_CMD_USB_NO_CX ) {
    //midiUSBCx = false;
    return;
  }

  if (cmd == B_CMD_ENABLE_INTELLITHRU ) {
    //intelliThruActive = true;
    return;
  }

  if (cmd == B_CMD_DISABLE_INTELLITHRU ) {
    //intelliThruActive = false;
    return;
  }
```
Plus any time the `Wire` buffer has more than 2 bytes, nothing is returned.

At this point, I'm out of things for you to try and don't see how the timeout setting should affect it.  Since `setTimeout(0)` is working for you, my only recommendation for now is to just set it to zero.  Perhaps later on in your development you'll come across something that will shed some light on this.

dewhisna:
I said there are two cases where `WireSlave` can return error 4 (timeout and bus/arbitration error), but there's actually a third way it can return error 4 that I forgot about. It can return error 4 from `endTransmission` if that function was called and the `Wire` is either not in master mode or `beginTransmission` wasn't called first.

TheKikGen:
@dewhisna the switch case you mentioned doesn't trap anything and was purely removed in my last tests, with exactly the same result (was just copy past from the project code). 

> you say you are getting "random timeout less than 1ms" -- how do you know that's a timeout?? Are you saying that it returned within 1ms from requestFrom and had 0 bytes? Is that what you are calling a timeout?

Exactly, from the master point of view.

> As receiving 0 bytes doesn't necessarily mean a "timeout" condition.

So what does a 0 value returned by RequestFrom means ?
If we are in that ISR, it is because the master is waiting for data.
If data are sent by the slave, usually, the nb of bytes is returned.
If a timeout is set, and reached, then a 0 value is returned.
Else the master waits...
else....any other case could return 0 also .???? not clear.
It is clear however that getting 0 when no timeout set is an error.

>  If you write no bytes to the Wire in the requestEvent then nothing gets sent by the slave

Is it an issue if I'm not waiting anything ? 

> Or, if there is some electrical issue on the bus, it could have a premature stop condition. You do have 4.7K (or so) pullup resistors on your SDA and SCL wires, right?? 

Right.  3,2 K . +  I have an oscilloscope on the data line to check the quality of the signal. rise-time is short.  Pullup are mandatory with all I2C design.

> Since setTimeout(0) is working for you, my only recommendation for now is to just set it to zero

I agree. I think it is the only recommendation acceptable at the moment. setTimeout is not reliable enough.

> It can return error 4 from endTransmission if that function was called and the Wire is either not in master mode or beginTransmission wasn't called first.

This should be better documented. Interesting despite it is not that case here however.

=> So I go with a forced setTimeout(0)

TheKikGen:
@dewhisna . Donna, I rechecked everything.
1/ I changed SDA/SCL pullup resistors  by 2 new one 
2/ I rechecked all my prototype and wire connections on the test board, and removed the oscillo from the loop for now.
3/ I pulled the branch Rewire Work from your repo again (got 1 update ? maybe i wasn't on the last update...)
4/ I simplified as much as possible the master and slave

Guess what : no more errors or timeout with the default setTimeout to 1 sec.
Whatever the reason was exactly, it is now running perfectly with the master and slave test code.
So yes, I think your PR is now ready for merging !

dewhisna:
@TheKikGen  That's great news!  I was out of things to try here.  I was even looking at the assembly code that the compiler was creating, walking through that by-hand and couldn't find any issues with it.  I thought perhaps it wasn't treating the `systick_uptime_millis` global as volatile.  But it was.

I'm so glad to hear that everything is working correctly for you now.  Maybe you just had the problem I often do and forgot to switch branches on your git checkout or something and then when you do, forget to do a clean build?  That happens to me all the time.  In any case, this is good news.  And good luck with the rest of your project!

TheKikGen:
@dewhisna Hi Donna.  A rapid update.
Your Wire_slave is currently fully integrated in my production project and I tested it successfully with 5 midi devices on the bus (including 1 master) and default timeout !!
As tests shown before, everything is ok.  However I had a bug in a very specific case that made me wonder again....But it was on my side. The master was requesting 5 bytes, and the slave was sending 4 only..that gave the feeling of a timeout, because the 5th byte was sent on the next msg...So, lesson learn is that it is critical to check if master and slaves are aligned on the size of msg they are exchanging !  
@stevstrong do you think that PR could be merged now ?  Tests were quite intense...

stevstrong:
I don't have possibility to test this, but obviously was intensively tested by @TheKikGen, so I will merge it.
Should issues come up, we can still restore the previous state.

dewhisna:
@stevstrong  Thanks...  and if there are any issues.  Instead of reverting it, let me know and I'll gladly create a new PR to fix them, as I and @TheKikGen and many others are relying on it as a working master/slave solution.

jjbubka:
Hi folks, I have followed your conversation these days... I only write to thank you for your incredible work. 👏 👏

Issue #655, opened 2019-08-06
Serial numbers are shifted if not using SERIAL_USB


saloid:
Hello, if I am using board without USB, all serial ports are shifted, so I should use Serial3 when i mean 4 port (in stm32CubeMX it's 4 port, and if i turn on `SERIAL_USB` it is 4 port too).
https://github.com/rogerclarkmelbourne/Arduino_STM32/blob/a3a56866505eeca4b49fc435700dcd4bd5540c8c/STM32F1/cores/maple/HardwareSerial.h#L184-L222
I think port numbers should be always the same (in old versions of this file they were same) and for example if `SERIAL_USB` disabled define `Serial` as `Serial1`

saloid:
Example: I am using STM32F103RCT6 mcu.
```
| Pins       | CubeMX | with USB | without USB |
|------------|--------|----------|-------------|
| PC10, PC11 | UART4  | Serial4  | Serial3     |
| PA2, PA3   | USART2 | Serial2  | Serial1     |
```

stevstrong:
Yes, you are right.
The solution would be to assign Serial x always to the corresponding U(S)ART interface, and leave Serial only for USB serial.
Can you prepare a PR which solved that?

saloid:
I have checked and found that Serial number assigns in every board individually.
https://github.com/rogerclarkmelbourne/Arduino_STM32/blob/e009a060eb58e4a7a619e07557ac7f476fc20915/STM32F1/variants/generic_stm32f103c/board.cpp#L127-L139
How better to do in this case? Change it in every board and add #warning that Serial has changed?

saloid:
And if no SERIAL_USB in this board, maybe better to define Serial as Serial1?

stevstrong:
I think no warning is necessary because when SERIAL_USB is not defined then
`#define Serial Serial1`
should do the trick.

saloid:
I mean if somebody already use shifted numbers (like me), and they will be no more shifted - in this case I'd prefer to be warned. And how about changing this - I need to change it in all boards, right?

stevstrong:
Yes, you need to be change that for all boards.

victorpv:
This could break compatibility with older sketches that expect that behavior. I remember it has been like that forever.

stevstrong:
PR merged.

PR #682, opened 2019-09-22
Fix Serial numbers shifting when no SERIAL_USB 

saloid:
#655


saloid:
Not tested - need somebody to test.

stevstrong:
I have tested this for several examples and it builds error free.

stevstrong:
@saloid 
could you please remove from this PR the changes related to in-lining some functions, being not relevant to serial definitions?

saloid:
OK, I will revert inlines and open for them separate PR, right?

stevstrong:
Are the inlines needed for the serial stuff?

saloid:
I don't think so, they are even not from my commits. I have pulled them from main repo before making any changes in my fork (to be synced with original).

stevstrong:
Ah, ok, I will then check this as it is.

PR #691, opened 2019-09-26
Prepare for audio device class

### LINKED ISSUE #687: Audio Device Class Isochronous transfers missing payload in every other packet
I am trying to add audio device class to USBComposite [here](https://github.com/soreau/Arduino_STM32/commit/b0e8f7ee491b808ec5090b9b7ee669850681a19d). Currently it uses double buffered isochronous IN endpoint transfer. The device shows up as an audio input device however it has obvious problems. When trying to record with audacity or arecord, it takes exactly 4 times as long to record as it does to play back. In the current example I send a sawtooth wave (audioDataTxCb [here](https://github.com/soreau/Arduino_STM32/commit/b0e8f7ee491b808ec5090b9b7ee669850681a19d#diff-2415201b08e9f075e95576436c33e23dR145)) with the same values for each (stereo) channel. Here the resulting sawtooth [wave file](https://filebin.net/rv8omn6wpnwf7gss) recorded with `pasuspender -- arecord -d 5 -f S16_LE -c 2 -r 16000 -t wav -D hw:4,0 sawtooth.wav`. When examining the wave file, it's clear there is missing data. With real data, the playback is sped up and garbled. When I examine the situation with usbmon and wireshark, I notice every other packet from the device is missing the payload entirely. Here is a [pcap file](https://filebin.net/uzvf8vovynqe0nde) with the packets. The host sends an urb every 1ms but doesn't get a packet from the device until right before the host sends the next urb, seemingly 1ms late. The packets without payload are 80 bytes, the packets with payload are 112 bytes. This means it only gets the payload sent every 2ms instead of 1. With pulseaudio loopback module, the audio stutters as well. I am using arduino IDE to upload the following test sketch. The problem is the same when trying with single buffer. I have also tried different sample rates. Currently using with STM32F103C8 with nothing connected except usb and st-link v2.
```
#include <USBComposite.h>

USBAUDIO AUDIO;

void setup()
{
  AUDIO.begin();
}

void loop()
{
}
```

soreau:
A couple of libmaple additions to prepare for usb composite audio device class. Allows double buffered isochronous IN endpoints to function correctly. Fixes #687.

### DIFF
--- a/STM32F1/cores/maple/libmaple/usb/stm32f1/usb_reg_map.c
+++ b/STM32F1/cores/maple/libmaple/usb/stm32f1/usb_reg_map.c
@@ -66,14 +66,14 @@ static void usb_set_ep_rx_count_common(uint32 *rxc, uint16 count) {
         if ((count & 0x1F) == 0) {
             nblocks--;
         }
-        *rxc = (nblocks << 10) | 0x8000;
+        *rxc = (nblocks << 10) | 0x8000 | (count & 0x3FF);
     } else {
         /* use 2-byte memory block size */
         nblocks = count >> 1;
         if ((count & 0x1) != 0) {
             nblocks++;
         }
-        *rxc = nblocks << 10;
+        *rxc = (nblocks << 10) | (count & 0x3FF);
     }
 }
 /*
--- a/STM32F1/system/libmaple/include/libmaple/usb.h
+++ b/STM32F1/system/libmaple/include/libmaple/usb.h
@@ -123,6 +123,7 @@ typedef struct usb_descriptor_string {
 
 #define USB_EP_TYPE_INTERRUPT             0x03
 #define USB_EP_TYPE_BULK                  0x02
+#define USB_EP_TYPE_ISO                   0x01
 
 #define USB_DESCRIPTOR_ENDPOINT_IN        0x80
 #define USB_DESCRIPTOR_ENDPOINT_OUT       0x00


rogerclarkmelbourne:
Whats the best way to test this ?

soreau:
@rogerclarkmelbourne Install the code and run the example sketch on STM32F103C8. An audio device class should appear in the host system, in addition to a serial interface. Then you can pipe the audio input to speakers with the pulseaudio loopback module. I use `pactl load-module module-loopback latency_msec=1` to load it and `pactl unload-module module-loopback` to unload it. Supply any analog signal to PA0 and/or PA1 within the 3.3 voltage range. I've been testing using an [Si4703](https://learn.sparkfun.com/tutorials/si4703-fm-radio-receiver-hookup-guide/all) to supply audio but a microphone might work too. This solves #687.

rogerclarkmelbourne:
@soreau 
Thanks.

Are you using Linux ?  I'm not sure W7 supports Serial and Audio as a composite device (but I could be wrong)

soreau:
@rogerclarkmelbourne Yes I am using Linux. I don't have a real windows installation to test with and with windows 7 in virtualbox, it installs a composite driver but it searches for other drivers and ultimately fails. I tried on a real osx laptop but it simply did not show up in audio device settings. You might be able to try removing all instances of CompositeSerial in the sketch and just use AUDIO.begin(); so it's only an audio device but perhaps it needs some inf file to help it along.

rogerclarkmelbourne:
@soreau 

I have a feeling its not supported at all by W7, without specialist drivers.

I reload one of my laptops with W10 purely as a testbed, even though I am loathed to use W10 for all the usual reasons.

W8.1 may support composite devices, and thats my eventual upgrade path when Microsoft stop supporting W7

soreau:
@rogerclarkmelbourne I think you are right about drivers, I tried in a win10 virtual machine and it recognizes the device as an audio device but reports driver error. But it could be virtualbox usb interface. Maybe it has a chance on a real installation. However I too share your opinion about win10 especially.

soreau:
@rogerclarkmelbourne I reinstalled windows 10 in virtualbox, updated the sample rate to a standard rate of 24Khz and eliminated CompostieSerial from the sketch and now it shows up as an audio device and I can enable it and play audio from it. The audio is just noise but I am hoping this is just due to running through virtualbox. So in it's current form, windows 10 at least installs an appropriate driver for it.

soreau:
> Also, is there a way to test the other times the USB Composite functionality is used , to confirm nothing has been broken with this change ?

The only changes that I can think of that might break something is the [usb_reg_map usb_set_ep_rx_count_common change](https://github.com/rogerclarkmelbourne/Arduino_STM32/pull/691/commits/54d2a2318eaca352ab013951cc966162628a7a56) (though this is per RM0008 notes for USB_COUNTn_RX) and possibly [this](https://github.com/rogerclarkmelbourne/Arduino_STM32/pull/691/commits/0a863de8359f9b9507c0eff1beb7a4fa5defd8a2#diff-690749d912d0f6a67967af383e8ba895L427) though I think it should be harmless.

soreau:
@rogerclarkmelbourne I have cleaned up the patches (and code) and consolidated them all into four commits to keep the git history clean. I think the audio.ino sketch should work on windows 10, can you test when you have time?

soreau:
@rogerclarkmelbourne I added a [tone generator](https://github.com/rogerclarkmelbourne/Arduino_STM32/pull/691/commits/4e3901ebf46ef5f4ea3be1703bfd9a398c2b9d76#diff-6abd735b1485cdaf1a6a0624bab3b8cb) sketch that should work on windows 10 so you don't have to connect anything.

rogerclarkmelbourne:
Thanks

I've not had time to install W10 yet. (And I still need to find the installation media)

soreau:
@rogerclarkmelbourne After fiddling with the descriptors and reading a bunch of traffic captures and documentation, I was able to get all of the audio examples working on windows 7 and windows 10 with default drivers. I also implemented optional audio class 2.0 descriptors. I get audio data in virtualbox too so I'm pretty sure it should work. The serial driver doesn't install in windows 7 for sketches that do serial and only audio class 1.0 works but it still allows the audio device to function. In windows 10, everything works; serial with audio, and both audio device class versions.

rogerclarkmelbourne:
Thanks

I finally managed to install W10 on an old Macbook at the weekend, but I've not had time to test this PR yet.

arpruss:
I think it would be better to add USB Audio first to the upstream USBComposite library project, which is probably more up-to-date than the library included in this core: https://github.com/arpruss/USBComposite_stm32f1

Make a PR there, I will test it, and then I will make a PR to update this core.

soreau:
@arpruss I wasn't aware of the upstream repo. The only problem I see is that [patch 2](https://github.com/rogerclarkmelbourne/Arduino_STM32/pull/691/commits/db8ae5d304864cd09fcec317501f367b1f121f08) in this series makes a change to libmaple required for isochronous IN transfers.

soreau:
I've made the changes @arpruss suggested so now this PR is just for the single libmaple patch and the bulk audio implementation is upstream [here](https://github.com/arpruss/USBComposite_stm32f1/pull/30).

arpruss:
I merged it upstream. If Roger can merge the libmaple patch, then I'll submit a PR to him to merge the latest USBComposite.

soreau:
@arpruss Great, thanks! I assume this to say it worked when you tested?

rogerclarkmelbourne:
@arpruss 

Do you mean this PR ?

soreau:
@rogerclarkmelbourne Yes it is merged [upstream USBComposite](https://github.com/arpruss/USBComposite_stm32f1/pull/30). Now I have changed the current PR here to reflect that, so it only has two patches with 3 lines changed. If you can merge it, then @arpruss will submit a PR for upstream USBComposite. See the title and description at the top here, I updated those.

rogerclarkmelbourne:
Ok.

I merged it...

soreau:
> I think it would be better to add USB Audio first to the upstream USBComposite library project, which is probably more up-to-date than the library included in this core: https://github.com/arpruss/USBComposite_stm32f1
> 
> Make a PR there, I will test it, and then I will make a PR to update this core.

@arpruss bump.. the audio code has been added to the upstream USBComposite library project but there is still no PR to get this pushed to Arduino_STM32. Any chance this will happen soon?

PR #699, opened 2019-10-03
Added Calibration for Serasidis_XPT2046_touch library

michox:
I added a calibration example and made some changes to the library so that the points given now should match the pixel on the screen. Please have a look if it works for you. In the example, you might need to use another TFT library


stevstrong2:
@Serasidis 
Vassilis, since this is related to your library, could you eventually test this?

stevstrong:
@michox can you please comment a bit more in detail what the input parameters are good for?
Specially the added variable `display_rot` in conjunction with the input parameter `drot`.

PR #706, opened 2019-10-08
Fix I2C master 2-byte read case

### LINKED ISSUE #704: Wrong values from I2C sensor on bluepill board
I'm getting wrong values from my BME280 sensor using the latest version of Adafruit library with the master branch of Arduino_STM32 on a bluepill board. The initialization of the sensor works, but the values form the `read*` methods don't make any sense. I saw a lot of I2C related changes in the last time by @dewhisna and @stevstrong, so I tried with a version from July, before all the changes were added and it worked again (a3a56866505eeca4b49fc435700dcd4bd5540c8c). I'm happy to help to narrow down the changes which cause the problem, but there are a lot of commits in the meantime. If you could give me a list of commits which are worth testing, I can test them and report where the problems starts to show up.

Here the sketch I'm using:
```C
#include <Adafruit_Sensor.h>
#include <Adafruit_BME280.h>
#include <Arduino.h>
#include <Wire.h>

Adafruit_BME280 bmeSensor;

void setup() {
  Serial.begin(115200);

  delay(5000);

  Serial.print("init BME280...");
  if (!bmeSensor.begin(0x76)) {
    Serial.println("failed");
  } else {
    Serial.println("done");
  }
}

void loop() {
  Serial.print("temperature: ");
  Serial.println(bmeSensor.readTemperature());

  Serial.print("humidity: ");
  Serial.println(bmeSensor.readHumidity());

  delay(1000);
}
```

Example output old version:
```
temperature: 25.97
humidity: 53.08
```

Example output current master, just a few seconds later:
```
temperature: -0.09
humidity: 0.00
```

dewhisna:
This fixes #704 

The I2C master mode 2-byte read was incorrectly NAKing prematurely causing the second byte of all 2-byte reads to be returned as `FF`.

As per the ST datasheet for the CR1 register when using read Method 2, it's very important to clear the ACK bit AFTER clearing the ADDR (not before the ADDR is cleared!):

>Note: The POS bit is used when the procedure for reception of 2 bytes (see Method 2:
transfer sequence diagram for master receiver when N=2) is followed. It must be
configured before data reception starts. In this case, to NACK the 2nd byte, the ACK bit
must be cleared just after ADDR is cleared. To check the 2nd byte as PEC, the PEC bit
must be set during the ADDR stretch event after configuring the POS bit.

This PR swaps the clearing of the ACK-bit to after clearing the ADDR via the read of SR2.

Thanks @bergos for finding this issue.

@TheKikGen if you get a chance, please pull this commit and give your I2C code a quick check.  Apparently neither you nor I had done any 2-byte Master read sequences where the second byte was supposed to be something other than `FF`.  This should fix that.  This change should only affect the 2-byte master read path.

### DIFF
--- a/STM32F1/cores/maple/libmaple/i2c.c
+++ b/STM32F1/cores/maple/libmaple/i2c.c
@@ -491,9 +491,9 @@ void _i2c_irq_handler(i2c_dev *dev) {
                             sr2 = dev->regs->SR2;                   // Clear ADDR bit
                             dev->regs->CR1 = (cr1 |= I2C_CR1_STOP); // Stop after last byte
                         } else if (todo == 2) {
-                            dev->regs->CR1 = (cr1 &= ~I2C_CR1_ACK); // Disable ACK
                             dev->regs->CR1 = (cr1 |= I2C_CR1_POS);  // Enable POS
                             sr2 = dev->regs->SR2;                   // Clear ADDR bit
+                            dev->regs->CR1 = (cr1 &= ~I2C_CR1_ACK); // Disable ACK
                         } else {
                             dev->regs->CR1 = (cr1 |= I2C_CR1_ACK);  // Enable ACK
                             sr2 = dev->regs->SR2;                   // Clear ADDR bit


stevstrong:
I merge this as the possible risk to break something is low.

TheKikGen:
@dewhisna Hi Donna.  I'm not reading 2 bytes. 1 or up 2. So I wasn't concern by this issue. Good to know.   I 'll give it a test.  That version of Wire is by far the most stable I have used for now.

PR #712, opened 2019-10-31
Print updates

jjbubka:
Make write(buffer, size) compatible with standard arduino convention.

This fixes some problems that some libraries not overloading function while passing stream or print class (e.g. standard ethernet library). With this fix printing a web page changes form 1sec to 40msec.

Added /.vs/ in gitignore for visual studio users like me...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ other/maple-bootloader/*~
 *.1
 platform.local.txt
 boards.local.txt
+/.vs/*
--- a/STM32F1/cores/maple/Print.cpp
+++ b/STM32F1/cores/maple/Print.cpp
@@ -51,11 +51,10 @@ size_t Print::write(const char *str) {
 	return write((const uint8_t *)str, strlen(str));
 }
 
-size_t Print::write(const void *buffer, uint32 size) {
+size_t Print::write(const uint8_t *buffer, size_t size) {
 	size_t n = 0;
-    uint8 *ch = (uint8*)buffer;
     while (size--) {
-        write(*ch++);
+        write(*buffer++);
         n++;
     }
 	return n;
--- a/STM32F1/cores/maple/Print.cpp
+++ b/STM32F1/cores/maple/Print.cpp
@@ -51,12 +51,13 @@ size_t Print::write(const char *str) {
 	return write((const uint8_t *)str, strlen(str));
 }
 
-size_t Print::write(const uint8_t *buffer, size_t size) {
+size_t Print::write(const void* buffer, uint32 size) {
 	size_t n = 0;
-    while (size--) {
-        write(*buffer++);
-        n++;
-    }
+	uint8* ch = (uint8*)buffer;
+	while (size--) {
+		write(*ch++);
+		n++;
+	}
 	return n;
 }
 
--- a/STM32F1/cores/maple/Print.h
+++ b/STM32F1/cores/maple/Print.h
@@ -36,9 +36,13 @@ enum {
 
 class Print {
 public:
-    virtual size_t write(uint8 ch) = 0;
+    virtual size_t write(uint8_t ch) = 0;
     virtual size_t write(const char *str);
-    virtual size_t write(const void *buf, uint32 len);
+    virtual size_t write(const uint8_t *buf, size_t len);
+	size_t write(const char* buffer, size_t size)
+	{
+		return write((const uint8_t*)buffer, size);
+	}
 	
 	size_t print(const String &);
     size_t print(char);
--- a/STM32F1/cores/maple/Print.h
+++ b/STM32F1/cores/maple/Print.h
@@ -36,15 +36,15 @@ enum {
 
 class Print {
 public:
-    virtual size_t write(uint8_t ch) = 0;
+    virtual size_t write(uint8 ch) = 0;
     virtual size_t write(const char *str);
-    virtual size_t write(const uint8_t *buf, size_t len);
-	virtual size_t write(const void *buf, uint32 len) {
-		return write((const uint8_t *) buf, (size_t) len);
+	virtual size_t write(const void* buf, uint32 len)
+    virtual size_t write(const uint8_t *buffer, size_t size){
+		return write((const void*) buffer, (uint32)size);
 	}
 	size_t write(const char *buffer, size_t size)
 	{
-		return write((const uint8_t *) buffer, size);
+		return write((const void *) buffer, (uint32) size);
 	}
 	
 	size_t print(const String &);
--- a/STM32F1/cores/maple/Print.h
+++ b/STM32F1/cores/maple/Print.h
@@ -39,9 +39,12 @@ public:
     virtual size_t write(uint8_t ch) = 0;
     virtual size_t write(const char *str);
     virtual size_t write(const uint8_t *buf, size_t len);
-	size_t write(const char* buffer, size_t size)
+	virtual size_t write(const void *buf, uint32 len) {
+		return write((const uint8_t *) buf, (size_t) len);
+	}
+	size_t write(const char *buffer, size_t size)
 	{
-		return write((const uint8_t*)buffer, size);
+		return write((const uint8_t *) buffer, size);
 	}
 	
 	size_t print(const String &);


stevstrong:
Please keep only [the change which solves your issue](https://github.com/rogerclarkmelbourne/Arduino_STM32/blob/345786e9c1fada935ec62f5bfb7b4b4bd19fc85f/STM32F1/cores/maple/Print.cpp#L54-L61), changing the types are not necessary.

Also, in the header keep only the corresponding definition(s), without the residual type changes.

jjbubka:
I have reverted the changed types. 😃

jjbubka:
OK, then I preserve the old original write. For my application this is working. 
Maybe this PR needs yo be tested with various devices to see if all is ok.

dewhisna:
While this should now allow the other libraries in this framework to compile and work, this will reduce performance in those other libraries, depending on which overload the target code happens to take.  Since your `virtual size_t write(const void *buf, uint32 len)` delegates to `virtual size_t write(const uint8_t *buf, size_t len)`, rather than the other way around.  It means that in those other libraries, which were overriding `virtual size_t write(const void *buf, uint32 len)` to do a full copy in one operation will be relegated to getting called with single `write(uint8_t)` calls from the loop in your `virtual size_t write(const uint8_t *buf, size_t len)` if the caller uses that overload, which bypasses the optimizations in those other libraries.

For example, in the aforementioned `WireSlave` library, it means that instead of allocating a single buffer for the data and copying the whole thing in one slice, it will get called 'n' times, allocating and reallocating the memory buffer byte-by-byte as if it wasn't even overriding the other method (since it's now "overriding the wrong one") -- which is the same problem you are trying to fix on the ethernet library (or whatever library you are having the performance issue with).  In other words, you've fixed the performance in your library and broken the performance in the other libraries in this framework.

One possibility would be for you to turn this inside out.  That is, instead of having the `virtual size_t write(const void *buf, uint32 len)` delegate to `virtual size_t write(const uint8_t *buf, size_t len)`, have `virtual size_t write(const uint8_t *buf, size_t len)` delegate to `virtual size_t write(const void *buf, uint32 len)`.  That will at least preserve the functionality within this framework for its existing libraries.

To put it another way, instead of changing the original function signature and code at all, simply add your new overload and have it call the one with the `void *` instead.  That would leave the functionality of the existing libraries in this framework untouched.

That is.  Leave this:
```C++
size_t Print::write(const void *buffer, uint32 size) {
        size_t n = 0;
    uint8 *ch = (uint8*)buffer;
    while (size--) {
        write(*ch++);
        n++;
    }
        return n;
}
```

And add this:
```C++
virtual size_t write(const uint8_t *buf, size_t len)
{
	return write((const void *) buf, (size_t) len);
}
```

While not perfect, this should fix your performance issue and preserve the performance in the existing libraries in this framework (assuming your code is actually passing the `uint8_t *` type this new function you are trying to add is expecting, as it would be the one overridden in your library).

**But my guess is that you aren't actually passing a `uint8_t *` at all but another type that was resolving to the `void *` signature rather than your library's overload and by changing these signatures, you just got the compiler to prefer the one in your library.  If so, just casting what you are passing or changing its datatype would call the right overload in your library without even needing to change the code in this repository at all.**

The other option would be for you to go through the rest of the code in this framework and find all places that inherit from `Print` and that override `write` and add the additional overload to them as well to preserve their optimizations.

Also, why is the change to `.gitignore` included in this PR?  It seems completely unrelated.

jjbubka:
Thank you for the help dewhisna. The performance issue you are describing is exactly what happens to my code. But, in fact my code is calling <printclass>.print("blahblabblah"), and it is in fact the print class itself who is casting to the default write in print, and this is not overloaded with the ethernet library. So i cannot simply cast the buffer

To clarify, this is what i am doing with simplified code:

```
void mainFunction(){
    EthernetClient client;
    ...
    ...
    printFunction(&client);

}

void printFunction(Print* printclass){
   printclass->print("sometext");
   printclass->print("sometext");
   printclass->println("sometext");
}
```

I will commit the changes you propose and i will test it on monday. Of course the most important thing is not to break what is working now in this repo.

Have a good weekend!

jjbubka:
PD. I will get rid of the gitignore change if its finally merged. The problem is that I use Visual Studio, and generates garbage to the commits if I not omit that directory. Sorry.

jjbubka:
After some testing I can't see an easy fix for this, as implementing only the virtual:
```
virtual size_t write(const uint8_t* buf, size_t len) {
		return write((const void *) buf, (uint32) len);
	}
```
will fix my issue, but will interfere with actual repo. As you can see in Print.cpp all functions that are using write, are casting to uint8_t, so will break libraries that are overloading only write(const void* buf, uint32 len).

So i think the real solution were replacing all references of write(const void* buf, uint32 len) to the standard one. Maybe I am wrong or I am missing something, what do you think?

stevstrong:
I think after all that it would be more safe if we would omit this PR.
@jjbubka , which Ethernet library do you use?
I think that you should fix/implement the right overdrive function in that lib instead of changing the core files.

jjbubka:
@stevstrong. I have used the ethernet library that is in your repo, and now I use the official arduino. Both shows the same behaviour, as both overloads write( (uint8_t*)...) . The problem is that my program passes a print object instead an ethernet one (this is actually required for me), so this is a very particular situation.

I think adding only the virtual is a fix that will work, but who knows... I will use this patch in my repo, so maybe is better to skip this PR here.

Thank you!

stevstrong:
Please open an issue for that lib in my repo, I will check there for a solution.
Please also provide a simple example where the issue can be reproduced.

jjbubka:
I take note, i will open an issue ina free temporary slot. 😄 
Thank you!

stevstrong:
I will close this because of wontfix. It seems more appropriate to solve this in the particular lib used by OP.

dewhisna:
@jjbubka -- Yes, the fix of:
```C++
    virtual size_t write(const uint8_t* buf, size_t len) {
        return write((const void *) buf, (uint32) len);
    }
```

Is exactly what I was saying will work for you and not break anything.  It will add a vtable slot for the `const uint8_t *` variant, which your library is overriding.  All of the functions in the existing `Print` class are already casting to `const uint8_t *` for calling this function.  It's just that today without the above in the class, the compiler is falling back to the `const void *` path.

However, the above will call the `const void *` version, which means that classes, like `TwoWire` in `WireSlave` that only overload the `const void *` version will still work, as the above overload calls it.  And classes like the Ethernet library you are trying to use, that only overload the `const uint8_t *` version will work correctly, because that's what `Print` is already calling.

At most, the above fix only slows things down by one level of call indirection if the compiler can't optimize it out due to it being a `virtual` function and it will add 4-bytes to every class instance because the vtable will grow by one entry (meaning more RAM usage).  But that isn't that great of a cost, relatively speaking.

So yes, the above is exactly what I was saying to do to fix it and not break anything.  The other solution, as @stevstrong suggests, is to change the overloads in the Ethernet library to add both forms there -- specifically the `const void *` version so that your library overrides the one in the base `Print` class.

stevstrong:
I think the function `write(const void * buffer, uint32 len)` only serves as a fallback solution if the inheriting library does not provide such a function.

But as I mentioned above, usually the libraries should provide their own specific `write(const uint8 * buf, uint32 len)` functions, so that the one with a pointer to a void buffer from the original Print class will be not used at all.

I also added this new function to my Ethernet library 6 moths ago, check https://github.com/stevstrong/Ethernet_STM32/commit/b85cb297bf8294ba4f55fb016a687941f0a04a62.

jjbubka:
@dewhisna, you were right (as always 😄 ). After all the tests this morning I realized well what was happening.
If you are confident that the patch will not break anything, maybe @stevstrong can reopen the PR.

@stevstrong, if you pass a Print class then you dont have reference to the optimized ethernet write function, you need it defined the same way in the Ethernet class and the print class.

stevstrong:
> @stevstrong, if you pass a Print class then you dont have reference to the optimized ethernet write function, you need it defined the same way in the Ethernet class and the print class.

Not quite.
As I said, if you define your own function in the lib, it will overrule the default Print class function, which is declared as "virtual" exactly for this reason.

PR #718, opened 2019-11-25
Keep I2C IRQ processing from affecting system timing

dewhisna:
Disable only I2C IRQs when avoiding IRQ race-condition instead of global interrupts.  This avoids affecting the entire system timing when waiting for I2C state change.


stevstrong:
I think it makes sense, so I will merge it.

dewhisna:
@stevstrong  Thanks!  I am trying to milk every ounce of clock cycles out of this processor, and I was noticing that when I2C usage was up, the other IRQs, like SPI and Timers, were getting a little behind and realized that this code should only enable/disable the I2C interrupts anyway.  Most applications wouldn't even notice the difference, but when you need a faster processor in your design, it can make all the difference.

Thanks for getting it merged!

PR #726, opened 2019-12-05
Add a way to remap I2C1 pins after construction of standard Wire instance

gmcn42:
Normally there is no way to edit the dev_flags variable after Wire has been constructed. It is sometimes helpful to do so though. Example: u8g2 uses the builtin Wire instance to communicate with I2C displays which is normally set to the standard pins.

Using the added functions from this PR it is possible to:

```
//during setup():
uint8_t df = Wire.getDevFlags();
df |= I2C_REMAP;
Wire.setDevFlags(df);
```

After that u8g2 will nicely communicate with an I2C display connected to the alternate pins.

### DIFF
--- a/STM32F1/libraries/Wire/Wire.cpp
+++ b/STM32F1/libraries/Wire/Wire.cpp
@@ -102,4 +102,12 @@ void TwoWire::setClock(uint32_t frequencyHz)
 	}
 }
 
+void TwoWire::setDevFlags(uint8 df) {
+    dev_flags = df;
+}
+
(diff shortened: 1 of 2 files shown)


stevstrong2:
@gmcn42 ,
could you eventually define these functions in the header file only, so that the Wire.cpp remains untouched?

gmcn42:
Sure. Amended PR.

stevstrong:
I think there is a small mistake, the input parameter should be `df` if I'm not wrong.
Can you please correct that?

PR #727, opened 2019-12-08
support for fake stm32f103c8t6

arpruss:
This PR adds support for the 32K flash/10K ram blue pill boards:

 https://embeddedtronicsblog.wordpress.com/2018/12/29/fake-stm32-blue-pill-boards/

The current bootloader doesn't work with these, but serial upload works fine.

### DIFF
--- a/STM32F1/boards.txt
+++ b/STM32F1/boards.txt
@@ -358,6 +358,13 @@ genericSTM32F103C.menu.device_variant.STM32F103CB.build.ldscript=ld/jtag.ld
 genericSTM32F103C.menu.device_variant.STM32F103CB.upload.maximum_size=131072
 genericSTM32F103C.menu.device_variant.STM32F103CB.upload.maximum_data_size=20480
 
+## STM32F103FEBK6 (fake STM32F103C8T6) -------------------------
+genericSTM32F103C.menu.device_variant.STM32F103C6=STM32F103FEBK6 (10k RAM. 32k Flash)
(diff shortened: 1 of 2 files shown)


arpruss:
That was quick. I had a change of heart after this PR, though, since these are not really stm32f103c units, so I think they should have a separate main menu entry rather than submenu variant. I've also since then got the bootloader working. Should I make another PR?

rogerclarkmelbourne:
Please make PR for the bootloader.

Btw. As the chips ate not STM32, perhaps we should  call them something else.

E.g.  FSTM32 or FAKE32... etc

arpruss:
You don't think it's just a relabel of a lower end model, like the C6? I could run some tests to identify it if you have ideas.

rogerclarkmelbourne:
Ok.

It could be a relabel, but it may also be a CS32 or a GD32

arpruss:
I should check if Serial3 works. The C6 doesn't have it.

rogerclarkmelbourne:
Good idea

arpruss:
USART1 and USART2 work, but no USART3 on PB10/PB11. This fits with the stm32f103c6 hypothesis. 

In any case, the core support should work for the c6 as well, so I might as well just call it a stm32f103c6 in the menu.

stevstrong:
> In any case, the core support should work for the c6 as well, so I might as well just call it a stm32f103c6 in the menu.

This would make sense.

PR #729, opened 2019-12-11
stm32f103c6 support improvements

arpruss:
I've added ld scripts supporting the serial and HID boot on the low-density 32K flash / 10K ram stm32f103c6 (which is what the fake stm32f103c8 units seem to be). To make this work with the various overrides in boards.txt, I've had to move the support from the stm32f103c submenu into a separate menu item.


stevstrong:
I have tested it for several examples and it builds error free.

PR #733, opened 2019-12-18
added an example for USB-serial-uart sketch

ag88:
Added an example for usb-serial-uart (usb-uart dongle) sketch
with ESP-01 (esp8266) connectivity and notes on using this same
 sketch to flash roms on stm32(f103) over uart


stevstrong:
As this is not a "library", but just an example how to use the HW serial port, I think it would make more sense to move it to: https://github.com/rogerclarkmelbourne/Arduino_STM32/tree/master/STM32F1/libraries/A_STM32_Examples/examples

ag88:
oops, thanks didn't notice that

ag88:
ok moved into [STM32F1/libraries/A_STM32_Examples/examples/Communication/USB-uart-w-signals/]
(https://github.com/ag88/Arduino_STM32/tree/master/STM32F1/libraries/A_STM32_Examples/examples/Communication/USB-uart-w-signals)

stevstrong:
Shouldn't the example folder have the same name as the INO file? Or vice versa.
I think this is required by Arduino IDE.

ag88:
done renamed sketch to be same as folder and added parity handling for SetLineCoding command from host.

PR #738, opened 2019-12-27
correct some files form the last commit

stevstrong:


stevstrong:
This is needed because the previous merge contained some files which caused build error.

PR #742, opened 2020-01-02
Remove warning about arithmetic with void* pointer

arpruss:
There is a compilation warning generated that we can avoid with a cast.


victorpv:
Just out of curiosity, for my own learning, but why does it throw a warning in the first place? Do you see a reason for that?
From what I see the USB_PMA address is declared as (void *), then in the return we have a (uint32 *) cast, which seems the correct return type for that function. So it seems we will be casting from (void *) to (unit8 *) and then to (uint32 *), correct?
Is it that the compiler is throwing a warning where it shouldn't?

dewhisna:
@victorpv -- no, it's not the compiler throwing a warning where it shouldn't.  It's throwing a warning because the size of a `void` isn't defined.  There are many good tutorials online on pointer arithmetic.  I've picked [one at random here to link](https://www.learncpp.com/cpp-tutorial/6-8a-pointer-arithmetic-and-array-indexing/).

If you have a pointer variable, for example, to a `uint32_t` type, such as:
```
uint32_t *ptr;
```
If you add `1` to that pointer with either `++ptr` or `ptr+1`, it increments the pointer by the size of what it's pointing to.  Here's an example:
```
uint32_t myVariable;
uint32_t *ptr = &myVariable;
// Assume that the address of myVariable is 0x1234
// Initially ptr will be equal to 0x1234
// if we excute:
++ptr;
// now, ptr will be equal to 0x1238
ptr = ptr + 1;
// now, ptr will be equal to 0x123C
```

However, if the type for `ptr` was `uint8_t *` instead, then after the `++ptr` above, then `ptr` would be equal to `0x1235` instead, and will be `0x1236` after the `ptr=ptr+1`.  It increments by the size of what it's pointing to -- making it useful for array indexing and such.

But if `ptr` was defined as `void *ptr` and you did `++ptr`, what's the size of a `void`?  How much does it increment by??

The default on many compilers, if you ignore all the warnings, is for the compiler to just treat it as a size of 1.  But that isn't really correct and so the compiler is correct to complain about -- just be happy that the compiler doesn't treat it as a totally undefined operation and add something random to the pointer just to prove its point.  Too many people stick their heads in the sand and ignore warnings that they shouldn't, or worse yet, use compile options to turn them off instead of fixing their code.

Here's some more links for reference:
https://stackoverflow.com/questions/3377977/void-pointer-arithmetic
https://stackoverflow.com/questions/1864352/pointer-arithmetic-when-void-has-unknown-size

victorpv:
But we had a cast to (uint32 *) already, would not be enough moving that cast inside the parenthesis?
`return ((uint32*)USB_PMA_BASE + 2 * offset);`

stevstrong:
@arpruss , can you please tell me under which conditions do you get a compilation warning?
I do not get any.
Any why did you cast it as `(uint8_t*)` and not as `(uint32_t*)` ?
At the end, the return value is anyway a `(uint32_t*)`, and the constant `USB_PMA_BASE` is not multiplied, so the case Donna was explaining is not relevant for the constant.

arpruss:
I've had the warning show up when compiling my USBComposite library with all warnings on.

Casting to uint32* would break both my and core USB code. The function is supposed to take a byte offset from the beginning of the USB Private Memory Area and return a pointer to that part of the Private Memory Area. If we cast to uint32, it will return a pointer with offset equal to sizeof(uint32)*offset in the PMA. This will overflow the PMA in the call `usb_pma_ptr(USB_CDCACM_TX_ADDR)` in https://github.com/rogerclarkmelbourne/Arduino_STM32/blob/9270ebe4081ce357fe4e8e19edaf1aa87052a25d/STM32F1/cores/maple/libmaple/usb/stm32f1/usb_cdcacm.c 
```
(USB_CDCACM_TX_ADDR is 0xC0. If we multiply it by sizeof(uint32)=4, then we leave the 0x200 byte PMA.)

By the way, some people in this thread unfamiliar with usb32f1 USB may be confused by the 2*offset in the definition of the function and the uint32* cast at the front. Why multiply by 2 and why cast to uint32* at all? 

The reason is that the PMA on these controllers is weird. You write two bytes at a time to the PMA. You write two bytes to the PMA by writing a uint32 to the PMA address and the two high bytes of it are ignored. But all the writes must be uint32-aligned. This means that if you want to write six bytes to the PMA, you do this:
```
uint32* pma = USB_PMA_BASE;
pma[0] = data[0] | ((uint32)data[1] << 8);
pma[1] = data[2] | ((uint32)data[3] << 8);
pma[3] = data[4] | ((uint32)data[5] << 8);
```
And the same goes for reading:
```
uint32* pma = USB_PMA_BASE;
uint16 w;
w = pma[0];
data[0] = w&0xFF;
data[1] = w>>8;
w = pma[1]
data[2] = w&0xFF
data[3] = w>>8
...
```

In other words, you treat the PMA as an array of uint32's with the upper two bytes of each uint32 ignored, and all the data stored in the lower two bytes.

That means, that if n is even, the nth byte you will be stored as the low byte of pma[n/2]. But USB_PMA_BASE is basically (uint32*)0x40006000. So, the CPU address of pma[n/2] is 0x40006000+sizeof(uint32)*(n/2) = 0x40006000+2*n. And that's where the multiply by two comes from in the definition of `usb_pma_ptr`.

There is, thus, a factor of 2 difference between CPU and PMA addressing. This is all confusing.

**Comment by stevstrong** [comment]
I understand how the PMA addressing works, but I think your logic is not quite correct.
For example, if n=3, then the address would be 0x40006000+2*3 = 0x40006006, which would be incorrect.
I think the written bytes are always rounded up to the next n%4 address. So correctly the pointer should be casted to (uint32*) and added to (n/2) according to your first equation.

**Comment by arpruss** [comment]
I think you're only supposed to call this for even offsets. And for even n, the two formulas give the same answer. Dividing by two and then multiplying by four will probably generate more code, however, and this is inline so we want code size to be small.

**Comment by victorpv** [comment]
Related to USB, and the PMA, why is the function "void usb_copy_to_pma(const uint8 *buf, uint16 len, uint16 pma_offset)" so complicated?
Can't it be like this?
`void usb_copy_to_pma(const uint8 *buf, uint16 len, uint16 pma_offset) {
    uint32 *dst = usb_pma_ptr(pma_offset);
    uint16 n = len >> 1;
    uint16 i;
    for (i = 0; i < n; i++) {
        *dst++ = *buf++;
    }
    if (len & 1) {
        *dst = *buf;
    }
}`

**Comment by arpruss** [comment]
Victor:

Good question.

My first thought was that this won't work if pma_offset is odd. But the core code won't work for odd pma_offset either.

My second thought is that the authors of the code are afraid of unaligned uint16 access. Such access _might_ work, but ARM recommends against it: http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0552a/BABFAIGG.html

But that doesn't explain why their copy_from_pma uses the quick and dirty approach, which will potentially do unaligned uint16 writes, while copy_to_pma is much more careful.

**Comment by victorpv** [comment]
That was my thought about alignment, if it's fine for copy_from_pma, why not for copy_to_pma? The original code, the one commented out, mentions that's pretty much a traight copy from STM code. I checked the STM official core, and uses the same convulted code as libmaple.
I then went checking the code in libopencm3, and saw that their code doesn't care about the aligment for either, to or from. I went several times over our code trying to figure out it that was the simplest, and came up pretty much to the same as their code, which just a stright copy, one half word at a time.
I'm not sure what you mean with pma_offset being odd. I believe every transfer starts from the bottom of the buffer. The thing that could be odd is the number of total bytes. In that case the function will copy one extra byte in PMA buffer, but shouldn't matter what's copied. If you mean to send a packet that's lets say 15 bytes, then the 16th byte that was copied in the buffer shouldn't matter anyway because you tell the usb peripheral to send a packet that's 15 bytes long, so will ignore the 16th byte, same as it will ignore all the other data left in the PMA buffer from the previous transfer, which is never zeroed.

**Comment by dewhisna** [comment]
For what it's worth, here's the code from ST's own official STM32Cube generated code in `Drivers/STM32F1xx_HAL_Driver/Src/stm32f1xx_ll_usb.c`:

```C
/**
  * @brief Copy a buffer from user memory area to packet memory area (PMA)
  * @param   USBx USB peripheral instance register address.
  * @param   pbUsrBuf pointer to user memory area.
  * @param   wPMABufAddr address into PMA.
  * @param   wNBytes: no. of bytes to be copied.
  * @retval None
  */
void USB_WritePMA(USB_TypeDef *USBx, uint8_t *pbUsrBuf, uint16_t wPMABufAddr, uint16_t wNBytes)
{
  uint32_t n = ((uint32_t)wNBytes + 1U) >> 1;
  uint32_t BaseAddr = (uint32_t)USBx;
  uint32_t i, temp1, temp2;
  __IO uint16_t *pdwVal;
  uint8_t *pBuf = pbUsrBuf;

  pdwVal = (__IO uint16_t *)(BaseAddr + 0x400U + ((uint32_t)wPMABufAddr * PMA_ACCESS));

  for (i = n; i != 0U; i--)
  {
    temp1 = *pBuf;
    pBuf++;
    temp2 = temp1 | ((uint16_t)((uint16_t) *pBuf << 8));
    *pdwVal = (uint16_t)temp2;
    pdwVal++;

#if PMA_ACCESS > 1U
    pdwVal++;
#endif

    pBuf++;
  }
}

/**
  * @brief Copy a buffer from user memory area to packet memory area (PMA)
  * @param   USBx: USB peripheral instance register address.
  * @param   pbUsrBuf pointer to user memory area.
  * @param   wPMABufAddr address into PMA.
  * @param   wNBytes: no. of bytes to be copied.
  * @retval None
  */
void USB_ReadPMA(USB_TypeDef *USBx, uint8_t *pbUsrBuf, uint16_t wPMABufAddr, uint16_t wNBytes)
{
  uint32_t n = (uint32_t)wNBytes >> 1;
  uint32_t BaseAddr = (uint32_t)USBx;
  uint32_t i, temp;
  __IO uint16_t *pdwVal;
  uint8_t *pBuf = pbUsrBuf;

  pdwVal = (__IO uint16_t *)(BaseAddr + 0x400U + ((uint32_t)wPMABufAddr * PMA_ACCESS));

  for (i = n; i != 0U; i--)
  {
    temp = *(__IO uint16_t *)pdwVal;
    pdwVal++;
    *pBuf = (uint8_t)((temp >> 0) & 0xFFU);
    pBuf++;
    *pBuf = (uint8_t)((temp >> 8) & 0xFFU);
    pBuf++;

#if PMA_ACCESS > 1U
    pdwVal++;
#endif
  }

  if ((wNBytes % 2U) != 0U)
  {
    temp = *pdwVal;
    *pBuf = (uint8_t)((temp >> 0) & 0xFFU);
  }
}
```

Looks like they are keeping it all as plain integer arithmetic up until the final cast at the end to a pointer (which is a `uint16_t *`).

And if you are wondering about the `0x400U`, it's because the `USB_PMAADDR` is at that offset relative to the `USB_BASE` (see Drivers/CMSIS/Device/ST/STM32F1xx/Include/stm32f103xb.h):

```C
/* USB device FS */
#define USB_BASE              (APB1PERIPH_BASE + 0x00005C00UL) /*!< USB_IP Peripheral Registers base address */
#define USB_PMAADDR           (APB1PERIPH_BASE + 0x00006000UL) /*!< USB_IP Packet Memory Area base address */
```

arpruss:
Victor:

It's still risky to have code, without any comment, that has to be called with an even offset. One might, e.g., try to write a HID report piece by piece to the PMA buffer, and one of the pieces could have odd-length. 

I just checked USBComposite, and all the code there that calls the PMA copy code does so with the offset at the start of the PMA area for the particular endpoint. However, currently nothing stops a user from setting an odd packet size (e.g., CompositeSerial.setRXPacketSize(23)) and that will screw up the PMA copy code for any subsequent endpoints. This is reminding me that I'm going to have to modify things to force even alignment.

dewhisna:
@arpruss such as cut-and-paste the official code from ST's STM32Cube that I pasted above, which takes the argument as bytes, but then divides it by 2 (via shift right 1) and then copies it as 16-bit words to force the alignment.  ...  And all you'd have to do then is make a few small tweaks to make ST's code fit the code-signatures of this core. ... Though it looks like I forgot to paste the definition of `PMA_ACCESS`.  It's value is 2, as defined in `Drivers/STM32F1xx_HAL_Driver/Inc/stm32f1xx_hal_pcd.h`:
```C
#define BTABLE_ADDRESS                         0x000U
#define PMA_ACCESS                             2U
#endif /* defined (USB) */
#if defined (USB_OTG_FS)
#define EP_ADDR_MSK                            0xFU
#endif /* defined (USB_OTG_FS) */
#if defined (USB)
#define EP_ADDR_MSK                            0x7U
#endif /* defined (USB) */
```

You can find all of the above code-base by going to ST's website and downloading the latest STM32Cube or STM32CubeMX or whatever they are calling it these days (i.e. there used to be separate STM32Cube versions for each series, but they finally combined it into a single one so you didn't have to install multiple environments when working with multiple series, where 90+% of it's the same).  Then, launch it, start a new project, select one of the F1 micros (my example above came from a generation-run for a STM32F103CBT6, since that's what I'm using on [one of my circuits](https://oshpark.com/shared_projects/8Cf51huB)), enable the USB port in your project, and click "generate code", and then look in your project folder and see what it generated (the above snippets came from my project folder that it generated). ....  Now, I'm not using their generated code on my project.  I'm using this Arduino_STM32 core for it.  But I generated their core code for it so I could compare when things like this in the core need some attention.

victorpv:
@arpruss that problem with setting a PMA buffer area with an odd lenght would affect the other function today anyway. Is having such buffer size even supported? I thought the buffer size alignment was 4 bytes minimum from something I read in the reference manual.

arpruss:
This is just a reminder that warning still hasn't been fixed.

stevstrong:
I think it is safe to merge this, I tested it and worked ok.

PR #753, opened 2020-02-15
F1: RTCAdjustment example add incremental calibration

ag88:
Incremental calibration allows the user to call calibratertc() repeatedly.
This is done by running synctime() once and the user can then call calibratertc(() more than once.

calibratertc() computes the drift adjustments by adding up the cumulative adjustments each time the user call calibratertc(). By doing this, it accounts for the smaller differential errors detected from the elapsed time since the initial synctime() call, and hopefully provide more accurate adjustments over a longer time period.

so initially call synctime() then afterwards you can call calibratertc() more than once with the new time to sync, for each time it measures the total elapsed time difference from the initial time sync to get the differential error. And add this incremental error to the drift_duration which is basically the 1 in N seconds error with each calibratertc() measurements


stevstrong:
As it does not change any core files I see low risk so I will commit this.

PR #763, opened 2020-03-12
[FIX] Missing initialization for doubleBuffer and begun, may lead to …

ccestari05:
…undefined behaviour.


rogerclarkmelbourne:
OK.

I think by default gcc must have been setting both these variables to zero, but I agree its safer to explicitly set them to zero / false

PR #770, opened 2020-04-22
Added digitalPinToInterrupt macro for STM32F4

jgromes:
Hi,

digitalPinToInterrupt macro seems to be missing for STM32F4, which causes an issue if some Arduino library happens to be using it ;)

I added it the same way it was done on STM32F1.

### DIFF
--- a/STM32F4/cores/maple/wirish.h
+++ b/STM32F4/cores/maple/wirish.h
@@ -67,6 +67,8 @@
                                                    bitClear(value, bit))
 #define bit(b)                         (1UL << (b))
 
+#define digitalPinToInterrupt(pin)     (pin)
+
 typedef uint8 boolean;
 typedef uint8 byte;
 


stevstrong:
Shouldn't be (pin%16) ? Because there are only 16 interrupts...

jgromes:
You're probably right, I just took what was in STM32F1's wirish.h - that one also has only (pin). So is it correct in F1?

stevstrong:
Well, it depends on what type of input do you deliver for that macro.
For Arduino AVR boards, that is the plain number on the board pin.
Fro STM32 it is the pin number, which is PA0..15, PB0..15, PC13..15. In this case the macro should be as I indicated. It seems that  it is wrongly defined for F1 as well.
I am not sure how many people are using this macro a nyway. Most of the users just put the "bit" number there (0..15), which is different from the "pin" number (0..46).

stevstrong:
I just checked, the attachInterrupt() function for both F1 and F4 requires the "pin" number (0..47) so it seems the macro is correct. I will commit it.

Issue #620, opened 2019-04-04
"sprintf" and "dtostrf" does not work when compiling as "--specs=nano.specs" added in platform.txt


HamidSaffari:
Hi and best regards to creators.
"sprintf" and "dtostrf" does not work properly when compiling as "--specs=nano.specs" option added to the platform.txt .
as discussed here:
 http://www.stm32duino.com/viewtopic.php?f=28&t=1596&p=20642#p20627 
by adding :   --specs=nano.specs   at the very end of line in Documents\Arduino\hardware\Arduino_STM32\STM32F1\platform.txt on the line that starts with "recipe.c.combine.pattern=...."
 the flash usage reduced to almost half, but in price of loosing some huge-flash-using functions as they said. and one of them that I realized is "sprintf" and "dtostrf" in some form of use like in my case floats: sprintf(buff, "%5.1f", Temperature_C);
but since the Arduino_Core_STM32:  https://github.com/stm32duino/Arduino_Core_STM32 
has used that option by default, he also has written new version of "dtostrf" that works perfectly with this option and I used their function in Arduino_STM32 platform and works fine. here it is:
https://github.com/stm32duino/Arduino_Core_STM32/blob/master/cores/arduino/avr/dtostrf.c
now I think it's better replace this with your dtostrf.c library.
and also using "--specs=nano.specs" option by default and see the other lost to try fix it.

stevstrong:
What is the status of this issue?

stevstrong:
@HamidSaffari 
Can you please provide a PR?

HamidSaffari:
when you use "sprintf(buff, "%5.1f", Temperature_C);" with original framework and choose
"--specs=nano.specs" for compile the output is empty, but when U do what I described then it print out the number correctly.

danieltr3s:
@stevstrong I think this [pull request](https://github.com/rogerclarkmelbourne/Arduino_STM32/pull/790) is what @HamidSaffari was requesting.

Also see this [Pull Request](https://github.com/rogerclarkmelbourne/Arduino_STM32/pull/787)

Best regards,
Tr3s

stevstrong:
I will check the first PR.

stevstrong:
#790 is a fix for this and was merged so that this can be closed.

PR #790, opened 2020-05-25
Fix for sprintf when compiling with nano.specs option set.

danieltr3s:
Fix suggested by @HamidSaffari in issue https://github.com/rogerclarkmelbourne/Arduino_STM32/issues/620 .

### DIFF
--- a/STM32F1/cores/maple/avr/dtostrf.c
+++ b/STM32F1/cores/maple/avr/dtostrf.c
@@ -1,7 +1,7 @@
 /*
   dtostrf - Emulation for dtostrf function from avr-libc
   Copyright (c) 2013 Arduino.  All rights reserved.
-  Written by Cristian Maglie <c.maglie@bug.st>
+  Written by Cristian Maglie <c.maglie@arduino.cc>
 
   This library is free software; you can redistribute it and/or
   modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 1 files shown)


kad:
+1 for this fix

stevstrong:
Was this tested by anyone?

kad:
> Was this tested by anyone?

I tried it with small hack in Marlin 2.0.7.2 firmware. builds fine, in places where it was broken, now seems to print floats ok.

https://github.com/kad/Marlin/commit/90878465bbc458ed516d325e9fa7629a056f3039#diff-f8926a81e11f2d9a94bb619e586b96c7d3cd4452a5e63d1b2c215efd0071a346

stevstrong:
Is there a simple skecth wich shows the problem and the solution?

fpistm:
If it can help, it is used since a while in STM32 core.
Note that is is also used by String class:
https://github.com/rogerclarkmelbourne/Arduino_STM32/blob/59227d1ecab7e931a390119c20655f1c4147a668/STM32F1/cores/maple/WString.cpp#L114

You can use this sketch to test:

```c++
void setup() {
  bool validate = true;
  char sout[256] = "";

  Serial.begin(9600);
  while (!Serial);

  Serial.println(dtostrf(1.23, 7, 2, sout));
  if (strcmp(sout, "   1.23") != 0) {
    validate = false;
  }
  Serial.println(dtostrf(1.899, 2, 2, sout));
  if (strcmp(sout, "1.90") != 0) {
    validate = false;
  }
  Serial.println(dtostrf(-123456.78910, 127, 6, sout));
  if (strcmp(sout,
             "                                                                                                                 -123456.789100"
            ) != 0) {
    validate = false;
  }
  Serial.println(dtostrf(-123456.78910, 128, 6, sout));
  if (strcmp(sout,
             "-123456.789100                                                                                                                  "
            ) != 0) {
    validate = false;
  }

  if (validate) {
    Serial.println("SUCCESS");
  } else {
    Serial.println("FAILED");
  }

}

void loop() {
}
```

stevstrong:
I have tested it and it seems to work, so I will merge this.

stevstrong:
this also solves #620.

PR #802, opened 2020-06-16
Fix micadc example

### LINKED ISSUE #801: micadc example: no matching function for call to 'STM32ADC::setDMA(uint16_t [2], int, int, NULL)'
There is a compilation error in the micadc example for the USBComposite library. It fails with an error `no matching function for call to 'STM32ADC::setDMA(uint16_t [2], int, int, NULL)'`.
AFAIK, it's the line
https://github.com/rogerclarkmelbourne/Arduino_STM32/blob/9b81135699acc7356721957d7013fde052a0b55f/STM32F1/libraries/USBComposite/examples/micadc/micadc.ino#L29
The parameters for `setDMA` have changed and that invocation is no longer valid. I'm preparing a pull request that (should) solve it.

jreyesr:
Fixes #801


jreyesr:
Self-closing because it was fixed by https://github.com/rogerclarkmelbourne/Arduino_STM32/commit/5236648ade3e9ae17c38e372646bfc70f414d5aa

Issue #813, opened 2020-08-24
Java Runtime PATH problem under Windows (EXCEPTION_ACCESS_VIOLATION)


dadul96:
This problem occurs when a second (different) Java version is installed on a windows PC (besides the one provided with the Arduino IDE). Instead of using the provided Java version of the Arduino IDE the `maple_upload.bat`-file uses the Java version found in the environment path (oftentimes this is a different one).

This results in the following error message:
```
maple_loader v0.1
Resetting to bootloader via DTR pulse
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  EXCEPTION_ACCESS_VIOLATION (0xc0000005) at pc=0x000000007110b5db, pid=12532, tid=0x000000000000127c
#
# JRE version: Java(TM) SE Runtime Environment (8.0_261-b12) (build 1.8.0_261-b12)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (25.261-b12 mixed mode windows-amd64 compressed oops)
# Problematic frame:
# C  [jSSC-2.8_x86_64.dll+0xb5db]
#
# Failed to write core dump. Minidumps are not enabled by default on client versions of Windows
#
# An error report file with more information is saved as:
# C:\Users\Daniel\Documents\Arduino\hardware\Arduino_STM32\tools\win\hs_err_pid12532.log
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.java.com/bugreport/crash.jsp
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#
```

To solve this problem the following line should be added to the maple_upload.bat:
`if exist "C:\Program Files (x86)\Arduino\java\bin" set PATH=C:\Program Files (x86)\Arduino\java\bin;%PATH%`

This line temporarily changes the PATH during the execution of this batch file and ensures that the correct Java version is used.

This issue also has been discussed [here](https://www.stm32duino.com/viewtopic.php?t=604).

cad435:
Happens to me too, but for PlatformIO (wich afaik directly calls java with arguments without calling the .bat file).
Happened only after upgrading Java8 to newest release.

Solved by uninstalling Java8 and installing Java7 again... (Not recommended though, it is heavily outdated)

stevstrong:
I have to test the PR.

gibsonmb:
[...]

PR #814, opened 2020-08-24
Fix Java PATH problem on Windows (Fix Issue #813)

dadul96:
This solves the problem where this batch file may use a different Java version compared to the one provided by the Arduino IDE. 
(see Issue #813)

### DIFF
--- a/STM32F1/platform.txt
+++ b/STM32F1/platform.txt
@@ -118,7 +118,7 @@ tools.maple_upload.path.linux={runtime.hardware.path}/tools/linux
 tools.maple_upload.path.linux64={runtime.hardware.path}/tools/linux64    
 tools.maple_upload.upload.params.verbose=-d
 tools.maple_upload.upload.params.quiet=
-tools.maple_upload.upload.pattern="{path}/{cmd}" {serial.port.file} {upload.altID} {upload.usbID} "{build.path}/{build.project_name}.bin"
+tools.maple_upload.upload.pattern="{path}/{cmd}" {serial.port.file} {upload.altID} {upload.usbID} "{build.path}/{build.project_name}.bin" "{runtime.ide.path}"
 
 #Added tool for generic STM32 upload via serial to Serial Port 1 (pins PA9 and PA10) - note. Boot0 line needs to high on board reset to enable upload via serial
 # at the end up the upload the program is automatically run, without the board being reset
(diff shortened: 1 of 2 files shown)


stevstrong:
There might be a problem with this PR: as I can see, this will add the same string to the PATH each time the batch file will be executed.
I think there should be a safeguard implemented to add the string only if the string does not already exist in PATH.

stevstrong:
According to [this post](https://www.stm32duino.com/viewtopic.php?t=604) the solution would be different, but I am not sure that such change is really needed.
Usually, if you install Java it will add the new installation path to global PATH, so that after a PC restart the batch file should run ok.

dadul96:
> 
> 
> There might be a problem with this PR: as I can see, this will add the same string to PATH each time the batch file will be executed.
> I think there should be a safeguard implemented to add the string only if the string does not already exist in PATH.

As far as I understand the PATH is only set temporarily. See: [set-command](https://ss64.com/nt/set.html)

stevstrong:
Ok, that might be true. However, the path you set here is a custom path. Many users (just like me) use the standalone package of Arduino IDE, which means that it may be then installed into another custom folder. In these cases this PR will not work, right?

dadul96:
Yes, that is correct. This PR will only affect users which installed the Arduino IDE in the default directory, but it won't break anything for other installation variants.

stevstrong:
I think the best solution is to pass the Arduino IDE installation path to the batch file so that the correct path to Java can be set independent on the user preferences.

I have tested it and it works.

Necessary changes:
- in platform.txt (line 121):
```
- tools.maple_upload.upload.pattern="{path}/{cmd}" {serial.port.file} {upload.altID} {upload.usbID} "{build.path}/{build.project_name}.bin"
+ tools.maple_upload.upload.pattern="{path}/{cmd}" {serial.port.file} {upload.altID} {upload.usbID} "{build.path}/{build.project_name}.bin" {runtime.ide.path}
```
- in maple_upload.bat (line 8):
```
- java -jar maple_loader.jar %1 %2 %3 %4 %5 %6 %7 %8 %9
+ set PATH=%5\java\bin;%PATH%
+ java -jar maple_loader.jar %1 %2 %3 %4
```
Can you please try these changes and adapt this PR accordingly?

dadul96:
This solution is way better, thanks! I will update the PR tomorrow.
There is only one change necessary to work properly:
```
- tools.maple_upload.upload.pattern="{path}/{cmd}" {serial.port.file} {upload.altID} {upload.usbID} "{build.path}/{build.project_name}.bin" {runtime.ide.path}
+ tools.maple_upload.upload.pattern="{path}/{cmd}" {serial.port.file} {upload.altID} {upload.usbID} "{build.path}/{build.project_name}.bin" "{runtime.ide.path}"
```
This allows having a space in the path string.

Issue #454, opened 2018-03-10
troubleshhoting the "STM32duino bootloader" upload method - Linux Mint 18.3 (64bit)


lucky62:
Hello,
I have burned the bootloader (generic_boot20_pc13.bin) to the my "blue pill" board, but I cannot load the sketch using  "STM32duino bootloader" method.

bootloader is loaded successfully - I see these messages when connected to ttyACM0:
```
Congratulations, you have installed the STM32duino bootloader
See https://github.com/rogerclarkmelbourne/STM32duino-bootloader

For more information about Arduino on STM32
and http://www.stm32duino.com

```

I already have the 1k8 resistor between PA12 and 3.3V.
I have increased upload-reset value in the maple_upload script.

Still every time I got the result:
`dfu-util: No DFU capable USB device available`

Kernel messages are showing that first is connected device 1EAF:0003, then it is disconnected and new device 1EAF:0004 is connected.

```
Mar 10 09:33:42 T470 kernel: [ 6236.657072] usb 1-2: new full-speed USB device number 47 using xhci_hcd
Mar 10 09:33:43 T470 kernel: [ 6236.802412] usb 1-2: New USB device found, idVendor=1eaf, idProduct=0003
Mar 10 09:33:43 T470 kernel: [ 6236.802416] usb 1-2: New USB device strings: Mfr=1, Product=2, SerialNumber=3
Mar 10 09:33:43 T470 kernel: [ 6236.802419] usb 1-2: Product: Maple 003
Mar 10 09:33:43 T470 kernel: [ 6236.802421] usb 1-2: Manufacturer: LeafLabs
Mar 10 09:33:43 T470 kernel: [ 6236.802423] usb 1-2: SerialNumber: LLM 003
Mar 10 09:33:44 T470 kernel: [ 6238.128071] usb 1-2: USB disconnect, device number 47
Mar 10 09:33:44 T470 kernel: [ 6238.424770] usb 1-2: new full-speed USB device number 48 using xhci_hcd
Mar 10 09:33:44 T470 kernel: [ 6238.566776] usb 1-2: New USB device found, idVendor=1eaf, idProduct=0004
Mar 10 09:33:44 T470 kernel: [ 6238.566781] usb 1-2: New USB device strings: Mfr=1, Product=2, SerialNumber=0
Mar 10 09:33:44 T470 kernel: [ 6238.566784] usb 1-2: Product: Maple
[...]

PR #818, opened 2020-09-05
recompile upload-reset for 64-bit systems

midnight-wonderer:
The precompiled upload-reset in linux64 is actually for 32-bit systems.

Try `file ./tools/linux64/upload-reset` and you'll get
~~~
ELF 32-bit LSB executable, Intel 80386, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux.so.2, for GNU/Linux 2.6.24, BuildID[sha1]=dc8dc5ad6de94179a9d534b1a34aeb6d906640b1, not stripped
~~~

That's why it does not work.

Recompiled with
~~~
gcc -o ./upload-reset -O3 ./src/upload-reset/upload-reset.c
~~~


midnight-wonderer:
Probably related:
https://github.com/rogerclarkmelbourne/Arduino_STM32/issues/454

Credit:
https://community.platformio.org/t/upload-reset-no-such-file-or-directory/8656/5

stevstrong:
Can anyone else test it? I don't have Linux.

ddrown:
Testing on Ubuntu 20.04.1 and Fedora 32, this looks good.

I verified:
* Ubuntu gcc used compiles the exact same binary
* strace output from the 32 bit and 64 bit binaries have the exact same ioctl/sleep sequences to manipulate the serial port

+1 from me

stevstrong:
I have no Linux distro where to check, but I trust the two testers, so I will merge this.
It can be reopened if problems occur.

PR #820, opened 2020-10-02
Add GD32F103CB board support

stefkpl:
For use with 12Mhz crystal @96Mhz and USB serial.


rogerclarkmelbourne:
This repo originally included GD32 support, but the GD32 chip is not 100% compatible with the STM32 chip.
Your PR just uses the variant folder that already exists.

People reported problems with SPI and other interfaces not working on the GD32 the same as on the STM32.

So I had to remove the GD32 from the list of available boards.

I retained the GD32 variant folder, so that people like you, who want to modify the repo can try the GD32 variant, however because the GD32 is not 100% compatible with the STM32 I can't accept this PR, as other people will try to use the GD32 and raise bugs because of the incompatibility.

BTW. There was a big list of incompatibilities and differences posted on a Russian web site, but that was several years ago, so the post may no longer exist.

stefkpl:
Ok I understand, thanks.
Good job :)

PR #825, opened 2020-11-11
Fix "unknown type uint8_t" error

Craft4Cube:
Added missing include to "inttypes.h"

### DIFF
--- a/STM32F1/cores/maple/avr/dtostrf.c
+++ b/STM32F1/cores/maple/avr/dtostrf.c
@@ -21,6 +21,8 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <inttypes.h>
+
 
 char *dtostrf(double val, signed char width, unsigned char prec, char *sout)
 {


rogerclarkmelbourne:
Can you list the steps to reproduce the error

GitMoDu:
Why not just #include <stdint.h> where needed?

ArminJo:
@rogerclarkmelbourne Please merge it fast,
You can see errors [here](https://github.com/ukw100/IRMP/runs/1395315243?check_suite_focus=true#step:3:659)
and [here](https://github.com/ArminJo/ServoEasing/runs/1396135525?check_suite_focus=true#step:3:903)
and more [here](https://github.com/ukw100/IRMP/actions/runs/361418390)
**It crashes all my CI Workflows using your core!**

ArminJo:
And please consider to release it fast, since all my compiles crashes.

stevstrong:
@ArminJo ,
the best solution is to change in line 34
```
uint8_t negative = 0;
```
to 
```
unsigned char negative = 0;
```
Please change this in you PR and then I will commit it.

STM32F1/cores/maple/avr/dtostrf.c
```diff
@@ -21,6 +21,8 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <inttypes.h>
```
stevstrong, line 24: Remove this line.

ArminJo:
@stevstrong I totally agree, but the PR is not from me.
But of course I can provide a new PR for this, if it helps, but maybe it is easier to just change the line manually and then close this PR?
And thanks for the fast reply! 👍

stevstrong:
This should be fixed by https://github.com/rogerclarkmelbourne/Arduino_STM32/commit/194a45ae256c6ede786d58525fe1b7550c9e6614.

ArminJo:
T H A N K S 👍

PR #829, opened 2020-11-25
Adafruit_ILI9341_STM (stm32f1): Compatibility with current HEAD versions of Adafruit GFX

kiki-lamb:
The ILI9341 has depended on a fairly ancient (1.5.3) version of the Adafruit GFX Library. 

This commit makes it compatible with the current HEAD versions of the Adafruit GFX Library by making the following two changes:

- Update references to use textsize_x (renamed in version 1.5.4 of the graphics library) when calculating positions/sizes in Adafruit_GFX_AS.cpp.
- Provides the SPIClass::transfer(...) method expected by Adafruit BusIO's SPIDevice class.

### DIFF
--- a/STM32F1/libraries/Adafruit_GFX_AS/Adafruit_GFX_AS.cpp
+++ b/STM32F1/libraries/Adafruit_GFX_AS/Adafruit_GFX_AS.cpp
@@ -139,14 +139,14 @@ byte line = 0;
 for(int16_t i=0; i<height; i++)
 {
   if (textcolor != textbgcolor) {
-    if (textsize == 1) drawFastHLine(x, pY, width+gap, textbgcolor);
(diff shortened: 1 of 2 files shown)


stevstrong:
While changing all occurrences is one way to solve the issue, wouldn't be nicer to use 
`#define textsize textsize_x`
before the first occurrence?

kiki-lamb:
I think whether a #define would be 'nicer' is sort of a toss-up: it's more concise, but on the other hand it's one more macro to remember. If you'd prefer to solve it that way I'll happily concede, though.

stevstrong:
It is not a clean solution but I will merge this though.
If anyone needs more `transfer` functions for SPI, then I will recommend to use my repo.

PR #830, opened 2020-12-07
SPI: add parameter to spiclass::on transmit/on receive

mean00:
This PR adds a parameter/cookie to the spi class onTransmit/onReceive callbacks
This is useful to go back to the client class when the callback is invoked 
By default the parameter is null, so it should be compatible with the current API

It is especially useful when using a small RTOS like FreeRTOS

To give an example,  a transfer to the screen is armed and the task waits on a semaphore
The callback cookie is actually the client instance, and upon dereferencing, the semaphore is unlocked

i.e. 
static void Screen::screencallback(void * a) { myScreen *s=(Screen *)a; s->transferDone();}

### DIFF
--- a/STM32F1/libraries/SPI/src/SPI.cpp
+++ b/STM32F1/libraries/SPI/src/SPI.cpp
@@ -565,8 +565,9 @@ uint8 SPIClass::dmaSendAsync(const void * transmitBuf, uint16 length, bool minc)
     Victor Perez 2017
 */
 
-void SPIClass::onReceive(void(*callback)(void)) {
+void SPIClass::onReceive(void(*callback)(void *),void *cookie) {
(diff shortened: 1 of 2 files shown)


stevstrong:
I'm afraid the compatibility is not kept because you changed the first parameter (the callback function type) from `void(*callback)(void)` to `void(*callback)(void *)`.

mean00:
Indeed

stevstrong:
I will not merge this if it does not keep backward compatibility.

mean00:
yes, that goes without saying
Even if , as it is, the api is a bit awkward and probably not used much

stevstrong:
Closed due to missing backward compatibility.
You may reopen this PR if you re-write it to be compatible.

PR #833, opened 2020-12-10
Fix spaces in paths for F4

ThexXTURBOXx:
Two Quotation marks were missing and therefore the build fails for some users with spaces in their temp path (like me).


stevstrong:
I have tested, it works, thanks!

PR #846, opened 2021-02-17
Fix dfu_util upload

vinzenz:
"{runtime.ide.path}" was appended to the dfu-util call which is nonsense

### DIFF
--- a/STM32F1/platform.txt
+++ b/STM32F1/platform.txt
@@ -118,7 +118,7 @@ tools.maple_upload.path.linux={runtime.hardware.path}/tools/linux
 tools.maple_upload.path.linux64={runtime.hardware.path}/tools/linux64    
 tools.maple_upload.upload.params.verbose=-d
 tools.maple_upload.upload.params.quiet=
-tools.maple_upload.upload.pattern="{path}/{cmd}" {serial.port.file} {upload.altID} {upload.usbID} "{build.path}/{build.project_name}.bin" "{runtime.ide.path}"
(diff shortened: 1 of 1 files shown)


stevstrong:
It is not non-sense, it was specially added to solve issue #813 .
Do you have any problem with that?

vinzenz:
OK I see, so the problem is that the fifth parameter in the macos tool is used for dfuse-address => https://github.com/dadul96/Arduino_STM32/blob/c9ab50d6b7c60c75f6ca5948122e1e76cb625b01/tools/macosx/maple_upload#L37

Which caused me headaches uploading

vinzenz:
Same in Linux64:
https://github.com/dadul96/Arduino_STM32/blob/c9ab50d6b7c60c75f6ca5948122e1e76cb625b01/tools/linux64/maple_upload#L13

and Linux:
https://github.com/dadul96/Arduino_STM32/blob/c9ab50d6b7c60c75f6ca5948122e1e76cb625b01/tools/linux/maple_upload#L13

So the solution to this would be to change it to parameter 6 and have 5 empty, i guess

stevstrong:
I think the number in "if" condition should be changed to 6:
```
if [ $# -eq 6 ]; then
    dfuse_addr="--dfuse-address $6"

```

stevstrong:
@vinzenz this PR in this state does not solve anything.
Do you want to make some updates?

vinzenz:
Hi, TBH I don't even remember what this was about exactly and have moved on since.
So I just closed it

PR #850, opened 2021-03-10
STM32F4 backup domain registers and backup domain sram

ag88:
STM32F4: update bkp.h, bkp.c add support for backup registers and backup sram.
added example in STM32F4/libraries/RTClock/examples/BkpTest

### DIFF
--- a/STM32F4/cores/maple/libmaple/bkp.c
+++ b/STM32F4/cores/maple/libmaple/bkp.c
@@ -29,14 +29,13 @@
  * @brief Backup register support.
  */
 
-#include "bkp.h"
-#include "bitband.h"
-
+#include <libmaple/bkp.h>
+#include <libmaple/bitband.h>
+#include <string.h>
 
 /*
- * Data register memory layout is not contiguous. It's split up from
- * 1--NR_LOW_DRS, beginning at BKP->DR1, through to
- * (NR_LOW_DRS+1)--BKP_NR_DATA_REGS, beginning at BKP->DR11.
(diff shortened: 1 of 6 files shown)


ag88:
the comment about this PR :
https://github.com/stevstrong/Arduino_STM32/issues/44#issuecomment-795857671

stevstrong:
What is actually the status of this PR?
Is it safe to merge?

ag88:
this is done back then when I'm trying to use the backup registers on stm32f401.
The readme.md in this commit is here
https://github.com/rogerclarkmelbourne/Arduino_STM32/pull/850/files#diff-5b489eef3ad090c0a5bed41a1f684527533ee35d6300a3ec9e008013bf280689
and the test is here
https://github.com/rogerclarkmelbourne/Arduino_STM32/pull/850/files#diff-afb267cf20737b1168f92a9d093766aab94b0aa65704e08d1a6b2633f796ddac

stm32f401 don't have backup sram but that they do have backup registers and this commit is part of works i'm using in some of my sketches to keep some data in the backup registers as these can be safe kept by VBAT (e.g. a coincell for the RTC) even if the main VDD 3v3 is powered off.
I think this shouldn't break anything

stevstrong:
OK, then I will commit this.
If somebody will encounter a problem related to this, then I will revert it.
Thank you for contributing!

ag88:
thanks ! I'd monitor issues here and in www.stm32duino.com and would respond just in case someone is after all using it

Issue #869, opened 2022-02-07
Malloc hangs at blackpill F401CC - incorrect HEAP settings?


board707:
I am using a STM32 F401CC blackpill.
If I am trying to allocate memory using malloc, my program hangs/crashes on the malloc command. But for bluepill STM32F103 malloc works fine.
My research into the problem led me to the memory heap settings in the file STM32F4/variants/blackpill_f401/ld/common.inc
Lines 152-153 of the file defines borders of the heap. Traditionally the heap is placed between msp and bss sections, but in this file it placed between __msp_init and __data_end__. But __data_end__ is situated in FLASH, far from RAM area and not suitable for HEAP. It looks like a bug.
To fix this, I replaced in the line 152 __data_end__ to __bss_end__  :
`_lm_heap_start = DEFINED(_lm_heap_start) ? _lm_heap_start : __bss_end__;`
and as result malloc on F401 blackpill now works fine.

PR #874, opened 2022-03-21
Add STM32F411CE support

board707:
- Add blackpill_f411 variant
-  fix incorrect HEAP settings on STM32F401CC as issue #869


stevstrong:
I cannot test 411 variant, but if anyone has a problem with this can port an issue.
So I will merge this as I see low damage risk.

PR #878, opened 2022-04-21
Added build options to select HSI clock source - Generic F103C8

purewack:
I have tested this with a generic F103C8 without a crystal and it works fine along with USB, SPI, UART etc, I don't see any issues apart from maybe the obvious clock inaccuracy of the internal RC, but other than that this feature was pretty handy for me and I feel like it should've been there already.


stevstrong:
I will merge this as I see low risk.

PR #889, opened 2022-09-17
fixes #883

### LINKED ISSUE #883: random() without parameter like in standard Arduino API is missing.
Standard Arduino programs do not compile if they use` random();`

E.g.
```
/home/runner/Arduino/libraries/NeoPatterns/src/NeoPatterns.hpp:1474:31: error: no matching function for call to 'random()'
         tRandom.Long = random();
                               ^
/home/runner/Arduino/libraries/NeoPatterns/src/NeoPatterns.hpp:1474:31: note: candidates are:
...
/home/runner/.arduino15/packages/stm32duino/hardware/STM32F1/2021.5.31/cores/maple/wirish_math.h:50:6: note: long int random(long int)
 long random(long max);
```

if I replace the line 50 in wirish_math.h
`long random(long max);`
with
`long random(long max = __LONG_MAX__);`
everything works fine.

Do you require a PR?

ArminJo:
random() without parameter like in standard Arduino API is missing.

### DIFF
--- a/STM32F1/cores/maple/wirish_math.h
+++ b/STM32F1/cores/maple/wirish_math.h
@@ -47,7 +47,7 @@ void randomSeed(unsigned int seed);
  * @return A pseudo-random number in the range [0,max).
  * @see randomSeed()
  */
-long random(long max);
+long random(long max = __LONG_MAX__);
 
 /**
  * @brief Generate a pseudo-random number with lower and upper bounds.


rogerclarkmelbourne:
Thanks

I will need to test compile this to confirm it at least compiles, but I can't do that until later today at the earliest

rogerclarkmelbourne:
Thanks

My 'test' sketch still compiled OK, so I've merged the PR

PR #896, opened 2023-02-13
Support Arduino IDE 2.0

21km43:
Some paths have changed in Arduino IDE 2.0. This time, platform.txt has been changed to work with version 2.0.


rogerclarkmelbourne:
Umm.

Is there a way to detect Arduino 2, because this change would mean the repo is not compatible with Arduino 1 which some people may still be using

board707:
> this change would mean the repo is not compatible with Arduino 1 which some people may still be using.

I would say "the most people still using"

rogerclarkmelbourne:
@board707 

Yes. Arduino 2 is relatively new, and I'm not sure how many people would have updated.

I haven't updated yet, as there was no need to do so.

21km43:
Sorry, I noticed that the existing project supports Arduino IDE 2.0. I believe the reason it was not available before was that I had copied it to the Documents folder instead of the AppData Sorry,

PR #924, opened 2024-04-24
Fix "call to overloaded transfer() is ambiguous" issue

board707:
Two overloaded `SPI.transfer()`methods in file `STM32F1/libraries/SPI/src/SPI.h`:
```
void transfer(const void * buffer, size_t length);

void transfer(uint8_t * trx_buf, uint32 len);
```
produces a "call is ambiguous" error if called as:
`spi.transfer(uint8*, size_t);`

Making the type of second parameter the same in both methods eliminates the error.


stevstrong:
Thank you.

PR #925, opened 2024-04-24
Add return values to STM32F4 Wire.write() methods to be compatible with recent Adafruit libraries

board707:
Methods `write()` of the WireBase class for STM32F4 (`STM32F4/libraries/Wire/WireBase.h`) are erroneously defined as void, which makes them incompatible with recent versions of libraries such as Adafruit GFx. The same methods in the branch for STM32F1 returns int and doesn't produce incompatibility errors.
The fix adds a return value to methods, fixing compatibility issues.


stevstrong:
Thank you!

PR #936, opened 2024-08-30
STM32F4: Fix support of second I2C device (I2C2) and add a third (I2C3) 

board707:
In the current STM32F4 code, support for the second I2C channel is not fully written and works only on some controllers. Despite the fact that the variant files contains macros for setting I2C2 pins, it values are not used in the code and the  I2C2 device is always created on the hard-coded pins PB10 PB11. Because of this, I2C2 does not work on boards where the I2C2 interface uses another pins.
This PR fixes these errors, and also adds support for the third (I2C3) channel.
Tested I2C1, I2C2 and I2C3 on the blackpil_F401 board.


stevstrong:
Thank you!

PR #941, opened 2025-03-17
Fix "Generic STM32F407V series" variant: Remove CCMRAM 

board707:
I had to remove a CCMRAM memory region from the ld-files of "Generic STM32F407V series" variant to make a DMA work with [STM32F407VE board](https://stm32-base.org/boards/STM32F407VET6-STM32-F4VE-V2.0.html).
I copied `ld/common.inc` and `ld/jtag.ld` files from the`blackpill_F411` variant and then the board works fine.

### DIFF
--- a/STM32F4/boards.txt
+++ b/STM32F4/boards.txt
@@ -236,7 +236,7 @@ generic_f407v.upload.maximum_data_size=131072
 
 generic_f407v.build.core=maple
 generic_f407v.build.f_cpu=168000000
-generic_f407v.build.extra_flags= -DCRYSTAL_FREQ=8 -DLED_BUILTIN=PA6
+generic_f407v.build.extra_flags= -DCRYSTAL_FREQ=8 -DLED_BUILTIN=PA6 -DNO_CCMRAM
 generic_f407v.build.variant=generic_f407v
 generic_f407v.build.variant_system_lib=lib_f407.a
 generic_f407v.build.error_led_pin=PA7
--- a/STM32F4/boards.txt
+++ b/STM32F4/boards.txt
@@ -236,7 +236,7 @@ generic_f407v.upload.maximum_data_size=131072
 
 generic_f407v.build.core=maple
 generic_f407v.build.f_cpu=168000000
-generic_f407v.build.extra_flags= -DCRYSTAL_FREQ=8 -DLED_BUILTIN=PA6 -DNO_CCMRAM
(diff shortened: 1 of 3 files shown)


stevstrong:
I don't think removing CCRAM is a good idea.
I think maybe the linker file has a bug, but to find out where exactly, I need a simple sketch which demonstrates your problem.

rogerclarkmelbourne:
IIRC the internal connection to the CCRAM is not the same as the normal RAM, i.e I don't know if CCRAM works with DMA or with all periphericals e.g. FSMC

I don't have any experience of linker files and whether the linker files concatenate the CCRAM as part of the main RAM.
If so the linker files should probably be changed so that the CCRAM is in a separate section

board707:
> @stevstrong : I think maybe the linker file has a bug, but to find out where exactly, I need a simple sketch which demonstrates your problem.

I encountered this effect in a very large project that is unlikely to be used for testing, but I will try to create simple code to demonstrate the problem.

@rogerclarkmelbourne :  the CCRAM should NOT be completely remove from the linker file

However, after deleting this section, everything starts working.

stevstrong:
I think the main problem is that DMA cannot access CCRAM, so any DMA function using the stack (and there are some) will graciously fail.
So stack must placed into SRAM, as you also did. This change solves your problem, but is not optimal, because totally disregards CCRAM.
I have to think about a good solution for the linker file., which includes CCRAM, too.
Or if you have any suggestion, just let me know.

board707:
I think we could leave CCMRAM in the linker memory table, but do not address a msp_init to it.
It is as it is done in ld-files, autogenerated by STM32Cube.

board707:
Hi
I have done some tests and indeed, there is no need to disable CCMRAM completely. It is enough to remove the stack and heap pointer references from this area.

Accordingly, you only need to change the `common.inc` file, replacing in line 43:
```
PROVIDE(__msp_init = ORIGIN(ccmram) + LENGTH(ccmram));
```
`ccmram` with `ram`:
```
PROVIDE(__msp_init = ORIGIN(ram) + LENGTH(ram));
```

and  in line 44 :
```
_lm_heap_start = DEFINED(_lm_heap_start) ? _lm_heap_start : __ccmdata_end__;
```
replace `__ccmdata_end__` with `__bss_end__`

stevstrong:
Can you please update your PR accordingly?

board707:
Done, PR updated.

PR #944, opened 2025-07-01
Change order of F103C8 clock options to ensure 72 MHz defaults.

board707:
After merging a PR #878 the option "HSI 64 MHz" was added to the first place in the list of clock frequency options for the F103C8 bluepill and therefore became the default setting after changing the board.
This leads to confusion, since for most users, the standard clock option of the bluepill board is "72MHz (Normal)". Also many libraries and existing projects are designed for using with 72 MHz clock option.

This PR simply swaps the F103C8 clock options to restore 72 MHz clock as the default choice.


stevstrong:
Thank you!

PR #945, opened 2026-05-28
Remove Java dependencies from Windows toolchain

nebulas-star:
Transplantation Windows version upload_reset tool from [stm32duino/Arduino_Tools](https://github.com/stm32duino/Arduino_Tools/tree/main/src/upload_reset/win) to remove the reliance on the Java environment

Also removed the unnecessary Java source code, and old version dfu-util.exe


board707:
Change 161 files, really?
What is the benefit from it ? 
The package works fine for me without your changes.

rogerclarkmelbourne:
I'm also confused about why you are deleting files, many of them being source code.

If the change is to use a different method to reset and / or upload, then only the file which handles that would need to change, and preferably also you should add an new upload option rather than unilaterally changing something other people are using without any problems

